PR #61, opened 2015-01-22
Update RemoteControl.cs

medxo:
when you make the botowners you remove the spaces so i think you have to remove spaces from here too, then name like "Paladin: medxo" will work too


ORelio:
Very logical suggestion, although I think I was wrong from the beginning when replacing all whitespaces when parsing botowners. I should have first splitted with "," then trimmed each name ie removing only heading and trainling whitespaces. Also, as MCC checks that no parse error occured by checking that the username retrieved from chat message only contains A-Z a-z 0-9 and _ characters, I'm not sure that a title containing a colon and a space would be considered as a valid username. Merging anyway since your suggestion is indeed appropriate :)

PR #65, opened 2015-02-21
Add Herochat/Essentials PM Check

JamieSinn:


ORelio:
Nice, always useful to have more chat formats to be supported :)

PR #66, opened 2015-02-21
Add in game chat to be controlled via remote cntrl

JamieSinn:


ORelio:
You can already use /tell <mybot> send <thetext> to make the bot send a chat message or command.
This makes more sense that using /tell <mybot> <thetext> eg /tell <mybot> <help> would send "help".

PR #68, opened 2015-03-02
Add Essentials pay message check

JamieSinn:
${amount} has been received from {user}

the message is the amount sent. eg.
$1200 has been received from JamieSinn.
message = $1200
sendder = JamieSinn

### DIFF
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -223,6 +223,15 @@ namespace MinecraftClient
                     message = text.Substring(name_end + 2);
                     return isValidName(sender);
                 }
+                //Detect Essentials Pay message
+                //${amount} has been received from {user}.
+                else if (text.StartsWith("$"))
+                {
+                    sender = tmp[4].ToString().Remove(tmp[4].ToString().Length - 1);
+                    message = tmp[0].ToString();
+                    return isValidName(sender);
+                }
+
 
                 else return false;
             }


ORelio:
While I'm not against adding such a feature, the app does not currently have a ChatBot mechanism which would take advantage of that. And why treating this as a private message? I'm not trying to belittle your contribution, but just to understand the reasoning behind it: what would be the purpose?

JamieSinn:
I considered it a private message due to how it is a private message. I have this on my current fork, as the bot I am working on logs lottery messages to be added to a leaderboard for EcoCityCraft. 
The bot also logs who sent it money, into a separate file and then posts that to the server with CURL.

This is just a fun contribution that I thought may overlap, it may not be a global thing, but eh, why not try.

ORelio:
What an interesting project. Nice use of MCC's capabilities! But i'm not sure treating money sendings as private messages in a general-purpose API would be a good idea eg for other people making use of it in forks, or eg for remote control bot (sending money to your bot would cause weird behaviors eg "Unknown command ${amount}" being replied by the bot).

So there are two possibilities:
1) In your bot you check that it's not a public message nor a private message and then look for a money sending message using your piece of code (instead of putting it in the main API)
2) Add a separate money handling method in the bot API, since it's not public nor a whisper, why not, but like private or public messages, there are various formats to detect depending on the plugins, and the original client (non-forked) will not make use of it.

... Also making a plugin system (like the one in minecraft servers) would be a nice thing to avoid porting changes each time a new version of MCC is out, thinking about it...

JamieSinn:
I was planning on using MCProtolcollib, but it was too much of a pain with maven, so I decided to modify this to my needs as I have quite a large amount of C# experience.

I'll close this, as it's not needed.

Issue #71, opened 2015-03-25
No support for changing use of ANSI colors.


drXor:
This is a nice client, but it suffers from an annoying issue: there is an ANSI background I can't remove, which looks terrible on my terminal. I have attempted all sorts of witchery to remove it, to no avail, while keeping the rest of the colors intact. This just looks really bad on my terminal because ANSI "black" is actually white in my emulator theme.

A configuration option for at least disabling background colors would be appreciated. I would fork and fix it myself, but I don't into C#.

ORelio:
No problem for adding a setting, but stay aware that Minecraft chat is mostly white text and would be non-readable on a white background ^^

ORelio:
After a quick look, background color can be saved and restored, and useless changes were performed.
This is now fixed, thanks for your report. Checked on a linux terminal before commiting.

PR #73, opened 2015-04-16
Add option to disable background recolors.

drXor:
The console puts out this annoying black background, which screws with my terminal (I'm on OSX via mono). This PR makes this background opt-out. This PR has not been tested with .NET tools, but mono should be a sufficient facsimile.

(I accidentally made this based of master, I'll open the PR now and maybe fix it later.)

### DIFF
--- a/MinecraftClient/ChatBots/Alerts.cs
+++ b/MinecraftClient/ChatBots/Alerts.cs
@@ -88,8 +88,14 @@ namespace MinecraftClient.ChatBots
                                 ConsoleIO.Write(splitted[i]);
                             }
                         }
-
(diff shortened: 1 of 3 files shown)


ORelio:
As you can see on issue #71 that you opened, this is already fixed in commit 82c95be611a68d1e14813651140a01d32220c6d4 ;)

drXor:
Wat. Uh, ok I guess. It didn't work on osx terminal when I tried it the other day... I'm probably just dazed and confused.

ORelio:
This is not fixed yet in binary release, only in development source code.

PR #76, opened 2015-05-20
Added functionality to Auto Respond Bot

Bancey:
Added the ability to ignore player names, enabled by default.
Stopped the bot from responding to itself, causing some issues where the
bot would infinitely loop because it was detecting something in its
response

### DIFF
--- a/MinecraftClient/ChatBots/Auto Respond.cs
+++ b/MinecraftClient/ChatBots/Auto Respond.cs
@@ -55,7 +55,6 @@ namespace MinecraftClient.ChatBots
                 if (Settings.Respond_IgnoreUserName)
                 {
                     text = regex.Replace(text, "");
-
                 }
                 //Check text to see if bot should respond
                 foreach (string alert in respondon.Where(alert => text.Contains(alert)))
--- a/MinecraftClient/ChatBots/Auto Respond.cs
+++ b/MinecraftClient/ChatBots/Auto Respond.cs
@@ -54,7 +54,6 @@ namespace MinecraftClient.ChatBots
                 Regex regex = new Regex(@"\<[^\)]+\>");
                 if (Settings.Respond_IgnoreUserName)
                 {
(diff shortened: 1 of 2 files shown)


Bancey:
Removed unnecessary console output.

ORelio:
Seems like you are trying to determinate if this is a chat message or not:
```csharp
if (text.Contains("<" + Settings.Username.ToLower() + ">"))
````
```csharp
Regex regex = new Regex(@"\<[^\)]+\>");
````
If so, please use `isPrivateMessage`, `isChatMessage`, `isValidName` and so on
These methods from the `ChatBot` API are here to make your life easier :)

Bancey:
I'm not telling if it is a chat message or not, I'm stripping the <Player_Name> from the front of the string :)

ORelio:
Stripping username is easier if you know that it's a chat message: `isChatMessage` also gives you the message without the username, and makes easier for the bot to ignore itself:
````csharp
//string text = the raw message

string sender = "";
string message = "";

if (isChatMessage(text, ref message, ref sender))
{
    if (sender == Settings.Username)
    {
        return; //this is a message from the bot, feel free to ignore it
    }
    else text = message; //keep only the message, removing username
}

//rest of processing
````
This way you can remove username from any supported chat message format,
eg `<**faction ~nicknamed>: message`, not only from regular `<username> message` syntax.

Plus, less code is easier to read and maintain, and copy-and-paste is not recommended. I saw a copy of `FromFile` earlier, if you think it would be useful for your bot, moving it from `Alerts.cs` to `ChatBot.cs` would make it accessible for any chatbot without needing to copy-and-paste. This way, if a modification to `FromFile` is required, there is only one copy to update :)

Apart from that, nice improvements!
Seeing the bot responding to itself would be quite hilarious ;D

ORelio:
Manually added your changes in 00295611359e1f2773817d61770d76e380181842, thanks for your contribution!

PR #83, opened 2015-07-05
Added Dynamic Host Name support (DNS) and localhost default server

beppe9000:
PS. Can you explain me what the "." means in the line that I commented with the question?


ORelio:
Hello,

Thanks for your pull request, however, you are currently submitting a whole VB project along with your pull request. Can't merge all of this, you'd need to send a pull request with only the modified/added files that are really necessary, I don't really need you test projects ^^

Regarding the only modified file, Settings.cs, did you know that the `TCPClient` constructor, later used in Minecraft Console Client when connectig to the server in `McTcpClient`, will automatically take care of DNS resolutions? So you should not need to add an automatic DNS resolution... Also, `localhost` is *already* used as default server eg when you do not type anything when prompted and just press enter, `localhost` will be used as server IP.

The "." test used in `SetServerIP` is to differenciate server aliases and direct server IP/addresses since aliases won't contain a dot and actual addresses will eg `myserver.com` or `192.168.1.10`. The only direct address without a dot is `localhost`.

Note: Using regexes in simple cases is usually overkill eg splitting with `.` would be faster here.

beppe9000:
Whooooooops it is my fault : i was thinking my other subsequent commits to my own branch did not make it to the prior pull request! What a fool i am... still a lot to learn.

Regarding DNS I got errors when i tred the gui version (stable) so i assumed no DNS resolution was made.

However regarding the ip check with dot : it took about an hour to figure it out: maybe it deserves a comment...

PS.  Do you know how can i merge only selected commits in pull requests? This is embarassing :\

ORelio:
Will add a comment about the dot in server IP. The GUI version of the program is quite old -2013 - and was merely an example for developers. I don't really have the time for working on the GUI. Submitting only selected commits can be achieved by making a new branch with only this commit on it, I guess. I know Git is not-so-easy to use especially for beginners so no worries, I'm not offended.

beppe9000:
okay :)

PR #90, opened 2015-08-17
Fixed issues with passwords containing unicode special characters, such as "Ð".

Pokechu22:
This fixes [the issue that JarJpg was having](http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-tools/1263927-win-mac-linux-minecraft-console-client-1-8-3?comment=1409) with logging in due to their password containing "Ð".

The main fix is the change to ProtocolHandler's jsonEncode method.  Previously,
it used ['char.IsLetterOrDigit'](https://msdn.microsoft.com/en-us/library/cay4xx2f(v=vs.110).aspx) to see if it needed to be escaped, but "Ð" is a
letter and still needs to be escaped.  The fix is to simply check if it's in
the right range.

Previously, the password `TestÐ123` would be sent as `TestÐ123`, which Mojang's servers would reject.  Now, it is sent as `Test\u00D0123`, which is the correct thing.  (Tested with an alt account).  

There's also some changes to those methods for performance and clarity reasons.
Most of this is using a [`StringBuilder`](https://msdn.microsoft.com/en-us/library/system.text.stringbuilder(v=vs.110).aspx) rather than appending to the string.  Not
too important, but it makes things clearer.

You mentioned [here](http://www.minecraftforum.net/forums/mapping-and-modding/minecraft-tools/1263927-win-mac-linux-minecraft-console-client-1-8-3?comment=1415) that the keys were named OemSomething.  That's true, but it seems like the [`KeyChar`](https://msdn.microsoft.com/en-us/library/system.ConsoleKeyInfo.keychar(v=vs.110).aspx) property gets the right symbol automatically, so we fortunately don't have to deal with that.

### DIFF
--- a/MinecraftClient/ConsoleIO.cs
+++ b/MinecraftClient/ConsoleIO.cs
@@ -45,22 +45,18 @@ namespace MinecraftClient
 
         public static string ReadPassword()
         {
-            string password = "";
-            ConsoleKeyInfo k = new ConsoleKeyInfo();
-            while (k.Key != ConsoleKey.Enter)
+            StringBuilder password = new StringBuilder();
+
+            ConsoleKeyInfo k;
+            while ((k = Console.ReadKey(true)).Key != ConsoleKey.Enter)
             {
-                k = Console.ReadKey(true);
                 switch (k.Key)
                 {
-                    case ConsoleKey.Enter:
-                        Console.Write('\n');
-                        return password;
-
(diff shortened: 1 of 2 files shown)


ORelio:
Well, you're right, nice catch with IsLetterOrDigit matching `Ð` and thanks for the optimizations!

PR #92, opened 2015-08-22
Send MC|Brand information upon joining the game

Pokechu22:
[`MC|Brand`](http://wiki.vg/Plugin_channel#MC.7CBrand) is sent by the vanilla client on joining the game to state the version (which is usually "vanilla").  I don't know of any servers that use it, but it *is* used in some minecraft crash reports, so it probably should be sent.  The version info "Minecraft Console Client v1.8.2" is used here.

### DIFF
--- a/MinecraftClient/Protocol/Handlers/Protocol18.cs
+++ b/MinecraftClient/Protocol/Handlers/Protocol18.cs
@@ -139,6 +139,9 @@ namespace MinecraftClient.Protocol.Handlers
                     case 0x00: //Keep-Alive
                         SendPacket(0x00, packetData);
                         break;
+                    case 0x01: //Join game
+                        SendBrandInfo();
+                        break;
                     case 0x02: //Chat message
                         handler.OnTextReceived(ChatParser.ParseText(readNextString(ref packetData)));
                         break;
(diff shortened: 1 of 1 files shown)


ORelio:
Stupid question: Why should we inform the server that we are joining with MCC?

Pokechu22:
There's no reason why we *need* to, but the purpose of the `MC|Brand` system is to state what mods you have (if any), and most other custom clients and servers send it (eg [steven](https://github.com/thinkofdeath/steven/blob/121b0505aa7844b88b74283d05561bf19aa4f3a9/handler.go)).  So it's more of a "Why not" thing.

Pokechu22:
I suppose one case where it would be useful is that plugin developers could serve a menu that's chat compatible rather than defaulting to a GUI if they detect it (though, of course, probably most plugins won't).

ORelio:
Why not, but usually plugin developers don't bother handling third-party clients - I don't even know how you are supposed to retrieve client brand information eg from a Bukkit plugin - and server owners tend to block them instead if there is an easy way of detecting them: [eg1](https://www.spigotmc.org/threads/minechat.3969/), [eg2](https://bukkit.org/threads/blocking-minechat-similars.170000/), [eg3](http://www.curse.com/bukkit-plugins/minecraft/blockminechat), giving brand information makes this easier. Anyway, merging for now, let's see if something positive/negative happens.

PR #93, opened 2015-09-03
Indev

ZizzyDizzyMC:
New message filter addition and the addition of a new (Old) feature of the hero chat messages.
With a config option addition to settings.cs that allows by default [channel][rank] user: message to be disabled.

Other notes: First Github pull request. This is / was a terrible learning experience that cost me 8 hours of my life.


ORelio:
Hello,
And thanks for your pull request, that's really good if it's your first contribution to a project :)
- Seems like you submitted your personnal .suo file containing visual studio settings
- Is your HeroChat parsing utility incompatible with something?
(I mean, is there an advantage disabling it?)

ZizzyDizzyMC:
Yes the herochat detects normal server messages as "[rank] username: message" as private messages.  So I added what's needed to have it disabled by default.  It can be enabled if the server has hero chat because their herochat format is different than *most* common large servers.  It's rather unobtrusive and if I can find the time i'll be adding enable options for each chat format so to reduce the risk of exploitations in the remote control.  I had a server mod once nickname himself my name to abuse a bot.

How the herochat got in there after you removed it from Indev: The "releases" tab of the Indev branch has outdated source code and I had originally modified that outdated code. After realizing this I decided to add the hero chat ability to the updated (latest) source just in case someone down the road wanted it and was too shy to ask for it.

ORelio:
In fact, adding more settings is a pretty good idea, wich could be expanded into a whole INI section :
````ini
[MessageDetection]
# A custom rexeg could be used for detecting your own unusual message format
public=^<!u> !m$ # !u would expand eg to [a-zA-Z0-9_]+ and !m to .* for example
private=^[!u -> me] !m$ # Not sure about intuitive !u or !m tag names
# Or builtin parsers could be used when the above regex does not match
detectvanilla=true
detectfactions=true
detectherochat=false
# And customized eg for your bot abuse issue
allownicknames=false
````
(Just asking for comments before working on it, not asking to do that yourself)

... Which HeroChat feature am I supposed to have removed? I don't recall removing any HeroChat detection feature, and newer chat format were added _below_ older so are used only if older formats doesn't match the provided line.

Yes, the 'releases' section on this GitHub repository only contains latest stable release and latest stable source code, not any Indev release, these are available on the build server.

You can remove a file from your Git repository with:
````
git rm <filename>
git commit -m "removed <filename>"
````
Or alternatively just delete the file and go to the GitHub client to create a commit.
The identity of the commit author is configured in Git or in your GitHub client:
````
git config user.name "My Username"
git config --global user.email "your_email@example.com"
````
Note that GitHub provided an email you may want to use for creating commits in Git, see the email section of your GitHub account. Alternatively you may add an email in the email section of your GitHub account so that GitHub will consider the commits as yours.

ZizzyDizzyMC:
I've added + cleaned my additions to the ini settings and the chatbot.cs so people can disable chat detection for certain chat formats on servers.  Made a new \[Block\] to put the new settings in.

ORelio:
Well, everything seems good now, thanks for your contribution!

Issue #89, opened 2015-08-04
Proxy Issues


DigitalSniperz:
I have my own setup proxy, it allows ALL ports and HTTPS/SSL. When I connect the client to it, after the connecting to minecraft.net it shows this error, any clue what this could mean?

Unhandled Exception:
System.NullReferenceException: Object reference not set to an instance of an object
  at MinecraftClient.Protocol.ProtocolHandler+<>c__DisplayClass4.<doHTTPSPost>b__3 () [0x00000] in <filename unknown>:0
  at System.Threading.Thread.StartInternal () [0x00000] in <filename unknown>:0
[ERROR] FATAL UNHANDLED EXCEPTION: System.NullReferenceException: Object reference not set to an instance of an object
  at MinecraftClient.Protocol.ProtocolHandler+<>c__DisplayClass4.<doHTTPSPost>b__3 () [0x00000] in <filename unknown>:0
  at System.Threading.Thread.StartInternal () [0x00000] in <filename unknown>:0

DigitalSniperz:
I am using the latest beta version, I have tried other version but they consist of the same error. 

Version:
https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/34

DigitalSniperz:
Proof that the proxies can connect to the authserver:

[01:32] Starting: Test 2: Connection through the Proxy Server
[01:32] Connection to https://authserver.mojang.com:443 established through the proxy server.
[01:34] A default web page was successfuly loaded.
[01:34] Test passed.

ORelio:
Hello,
Are you sure that the issue happens while logging in?
Not all proxies are supported, and I did not write the proxy handling part.
See issue #80 for full explanation.

DigitalSniperz:
The error occurs after the "connected to <proxyip>", the proxy is hosted on a VPS and it has all ports unblocked. I've tried it with a proxy checker to check if https authserver/login server works on the proxy and both do. When I use the proxy on minecraft it lets me log in. I don't understand why its not working on your minecraft linux client.

ORelio:
Proxy support is not made by me. It's a free and open source C# library called [Biko Library](https://github.com/bentonstark/starksoft-aspen). This is the only one I could find that is free and open source, and if some proxies are not supported, I can't really know why since I'm not the author of this part of the code... What proxy software are you using?

ZizzyDizzyMC:
[...]

Issue #80, opened 2015-06-23
Proxy


Mouro001:
Doesnt work with proxy
http://i.imgur.com/VcWDPUg.png

Mouro001:
i added my username and password to the proxy and now it connects, but i recieve another error.
"
Proxy destination 172.17.1.3 on port 8080 responded with a 502 code - Bad Gateway.  If you are connecting to a Microsoft ISA destination please refer to knowledge based article Q283284 for more information.  Server response: Proxy Error ( The specified Secure Sockets Layer (SSL) port is not allowed. ISA Server is not configured to allow SSL requests from this port. Most Web browsers use port 443 for SSL requests.  )
Proxy destination 172.17.1.3 on port 8080 responded with a 502 code - Bad Gateway.  If you are connecting to a Microsoft ISA destination please refer to knowledge based article Q283284 for more information.  Server response: Proxy Error ( The specified Secure Sockets Layer (SSL) port is not allowed. ISA Server is not configured to allow SSL requests from this port. Most Web browsers use port 443 for SSL requests.  )
An error occured while attempting to connect to this IP.
Failed to ping this IP.
"

ORelio:
Hello,
Seems like the proxy you are using works with Minecraft Console Client, but the proxy refuses to forward your connection because you are trying to connect to a non standard distant port ie 25565 (Minecraft Server) instead or 80 (HTTP) or 443 (HTTPS). Not every proxy will work with MCC, either because it's not supported by the proxy library used in MCC, or because it has some limitations like port restrictions (like here) or some timeouts for making request (eg auto close connection after X seconds): proxies are usually made and optimized for web browsing.

ORelio:
Noting new since then, closing, feel free to reopen if needed ;)

PR #99, opened 2015-10-16
Proxy work-around / Fix.

ZizzyDizzyMC:
Added 'loginonlyproxy' option with true / false boolean options.
on 'true' only the minecraft login is redirected to the proxy.
Otherwise both the login and the server connection are routed though the
chosen proxy.

This was mentioned by me on Minecraftforums so I eventually pulled it off myself.  Not sure if it works 100% properly BUT it does work on my test box.  True works and false does nothing new.
Provides a semi-workaround to issues #89 and #80 on ORelio/Indev

### DIFF
--- a/MinecraftClient/McTcpClient.cs
+++ b/MinecraftClient/McTcpClient.cs
@@ -111,7 +111,14 @@ namespace MinecraftClient
 
             try
             {
-                client = ProxyHandler.newTcpClient(host, port);
+                if (Settings.LoginOnlyProxy = true)
+                {
+                    client = new TcpClient(host, port);
+                }
+                else
+                {
+                    client = ProxyHandler.newTcpClient(host, port);
+                }
                 client.ReceiveBufferSize = 1024 * 1024;
                 handler = Protocol.ProtocolHandler.getProtocolHandler(client, protocolversion, this);
                 Console.WriteLine("Version is supported.\nLogging in...");
--- a/MinecraftClient/McTcpClient.cs
+++ b/MinecraftClient/McTcpClient.cs
@@ -111,7 +111,7 @@ namespace MinecraftClient
 
             try
             {
-                if (Settings.LoginOnlyProxy = true)
+                if (Settings.OnlyForLogin)
                 {
                     client = new TcpClient(host, port);
                 }
--- a/MinecraftClient/Settings.cs
+++ b/MinecraftClient/Settings.cs
@@ -32,6 +32,7 @@ namespace MinecraftClient
         public static Proxy.ProxyHandler.Type proxyType = Proxy.ProxyHandler.Type.HTTP;
         public static string ProxyUsername = "";
         public static string ProxyPassword = "";
+        public static bool LoginOnlyProxy = false;
 
         //Other Settings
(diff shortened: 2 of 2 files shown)


ZizzyDizzyMC:
This is here because sometimes we just use proxies to login to minecraft but don't intend to (ab)use the proxies to play minecraft.  Most proxy owners don't like this and definitely block your average minecraft port.  

Keep in mind if you can't connect to port 25565 on an outside network this doesn't help you at all.

This is an intended fix for those trying to log in a LOT of accounts using one of my scripts.  Mostly because spamming mojang servers with logins WILL result in an IP blacklist / tempban from their services.  Spreading out logins between proxies to reduce the likelihood of that happening.

NOTE: Use public proxies at your own risk. Mojang may lock your account if you try to login too many times on a proxy others have abused first.

ORelio:
Hello,

Nice addition! The way of doing this is perfectly valid, however there are currently some issues with your pull request that needs your attention before merging:
- What is the purpose of removing `case "herochatmessagesenabled"` from `Settings.cs`?
- Did you know that `=` stands for assignation? This means that `if (Settings.LoginOnlyProxy = true)` will first assign `true` to `Settings.LoginOnlyProxy` and then check that the assignation statement's value is `true`. You should use `==` for testing a value. But as `if` expects a boolean, and `Settings.LoginOnlyProxy` _is already a boolean_, you can use: `if (Settings.LoginOnlyProxy)`

To update your pull request, just create a new commit and push it to your repository on the same branch as used for making the pull request: `Indev`

On a side note: How about using `onlyforlogin` as setting name? Since this setting can be found in `Proxy` section of the configuration file, the end user already knows this is a proxy setting, so no need to specify it again, and I think this would be more explicit, removing the need of adding a comment for explaining what is the purpose of the setting :)

ZizzyDizzyMC:
OH CRAP OOPS LOL.  The single '=' was a mistake should be '=='.  The case "herochatmessages" was removed was a double in the config file. (How I did not catch this earlier is beyond me.) 

Making changes and re-posting soon.

ZizzyDizzyMC:
I'm sad now and give up. The files changed themselves overnight and I clicked reload and now my build is failing. What changes were made to them?

ORelio:
Oh, yes, you are right, that `herochatmessagesenabled` line had nothing to do there. Well, did not change anything, since all your commits were pushed you could clone again your repository and restart working from there. If you can't, I can alternatively merge the pull requests and apply the fixes myself right after merging.

ZizzyDizzyMC:
I'm fixing it. Somehow my github client got set to where YOUR repository was MY repository and downloaded yours and overwrote mine. No reason given.

ZizzyDizzyMC:
Done.  After some minor fiddling with github client I got I switched back to my repository only after somehow committing changes directly to your project (That's impossible right?) after the changes were committed then reverted I switched back over to my repository where I sync'd my changes back to myself and edited those working files.

ZizzyDizzyMC:
It's ready to go. I've tested it though the weekend.

ORelio:
Great! Merged :)

PR #106, opened 2016-01-30
Add an option to add a delay between bot messages

Pokechu22:
Because bots can send several messages quickly, this adds an option to slow down the rate at which messages are produced (to avoid issues with antispam plugins).  This should help solve part of the troubles @mobdon was having in #105.

Right now the default time is 2 seconds per message.  However, messages are sent imediately if the bot doesn't need to delay (so if it's a bot that only outputs one or two messages, those will still happen imediately).  Also, note that it's limited per-bot right now.

I also added an optional parameter to the SendText method so that bots can avoid this behavior if they need to.  In some cases, they'll want to send multiple messages.


ORelio:
Pretty neat. Thanks :)

PR #107, opened 2016-01-31
Update to 1.8's language files.

Pokechu22:
I ocasionally see untranslated messages; this should resolve all of them.

The changes to Settings.cs, the core changes, just update the index file used and the normal hash.  The changes in ChatParser.cs primarilly fix the name: 1.7.4 used 'lang/en-GB.lang' but 1.7.10 and 1.8 use 'minecraft/lang/en-GB.lang' and 'realms/lang/en-GB.lang' (and realms comes first), meaning that the wrong language file is selected.  The name is updated to make sure the right file is used.  I also corrected the indentation in that block.

### DIFF
--- a/MinecraftClient/Protocol/Handlers/ChatParser.cs
+++ b/MinecraftClient/Protocol/Handlers/ChatParser.cs
@@ -85,12 +85,13 @@ namespace MinecraftClient.Protocol.Handlers
                 ConsoleIO.WriteLine("Downloading '" + Settings.Language + ".lang' from Mojang servers...");
                 try
                 {
-                        string assets_index = downloadString(Settings.TranslationsFile_Website_Index);
-                        string[] tmp = assets_index.Split(new string[] { "lang/" + Settings.Language + ".lang" }, StringSplitOptions.None);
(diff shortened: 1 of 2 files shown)


ORelio:
Does `ChatParser.cs` needs to be modified?

Pokechu22:
Yes - as I mentioned, the filename changed from `lang/en_GB.lang` to `minecraft/lang/en_GB.lang`.  Without making the change in `ChatParser.cs`, it actually grabs the realms translation file (`realms/lang/en_GB.lang`), which is useless for this purpose.

ORelio:
Ah. That was harder to see due to the change in indentation, and the debug printing ``Console.WriteLine(hash);`. Merging anyway.

Pokechu22:
... Oops, I thought I removed the `Console.WriteLine(hash);`.  That probably shouldn't be there, sorry.

ORelio:
No worries, just removed it ;)

Pokechu22:
One last thing to note: For it to actually update the language file, you need to delete the old one in the `lang` folder.

ORelio:
As usual :)

PR #109, opened 2016-02-07
Give bots access to plugin channels

Pokechu22:
This pull request adds some methods so that ChatBots can communicate with the server using [Plugin Channels](http://wiki.vg/Plugin_channel).  It also adds a method that is called when the client has logged in as a convenient time to register the channels.


ORelio:
Neat and fully flexible. Great work! (as usual!? ;D)

PR #116, opened 2016-02-23
Fixed IndexOutOfRangeException when bot is unloaded on update

initsuj:
For loop was iterating on a collection that was changed when bot[i].Update() resulted in a bot unloading, leading to index going out of range. Changed the loop collection to be immutable.


ORelio:
Indeed, nice catch and nice fix, thanks!

ORelio:
(in fact that was working previously but now that [i] index is accessed after updating with `ProcessQueuedText()` the issue appeared)

PR #118, opened 2016-02-27
Fix several bugs

Pokechu22:
This should fix several bugs that occur in 1.7.10, one of which involves an  ArgumentOutOfRangeException and another with plugin channels.  It also fixes the new line issue in the default config.


ORelio:
Oh, so the list wasn't cleared.... My bad.

PR #124, opened 2016-03-03
Indev

initsuj:
Account caching is implemented and tested with multiple clients using the same cache file. Still has remote possibility of file access collisions that I will make more graceful in the future

Adds single ini setting: accountcache=none #use 'none', 'memory' or 'disk'

I did not use the Refresh protocol  method because I could not successfully test a .condition where it was beneficial. If the token does not validate, a full login is called..


ORelio:
Seems all right after reading how you implemented it. Good job!

PR #125, opened 2016-03-03
Indev

Bc1151:
I updated the packet id's and copied them into a new class. I would do more however I am mobile and using the github interface.

The data for the protocol was taken from http://wiki.vg/Pre-release_protocol on the commit date.

I really enjoy this project and have been very fond of using it.

I hope this helps for what little it counts for :)


ORelio:
Oh, well, is 1.9 so different that a whole new class is needed? Haven't had the time to take a look at it yet but `Protocol18` can actually handle both `1.7` and `1.8`, and `Protocol16` can handle `1.4`, `1.5` and `1.6` :)

Pokechu22:
1.9 reshuffled a lot of the packets, so it does require a bunch of changes (just look at the table on the start).

ORelio:
Okay, if that's only the IDs I guess it can be implemented in the ID switch depending on the version. I was waiting for the `Protocol` page to be finalized so that I could compare the packets.

Pokechu22:
Pre-release protocol actually basically is a comparison, so it might be good for getting a look.

Also, there's a few other pretty big changes than just IDs - chunk data got changed a lot.  They've got the new version documented [here](http://wiki.vg/SMP_Map_Format); for comparison the old one is [this](http://wiki.vg/index.php?title=SMP_Map_Format&oldid=7164).

I don't think it makes sense to try and switch between packets for each version - I feel like that might end up pretty confusing.  However, it might make sense to use OOP beneficially and have Protocol19 inherit from Protocol18, overriding methods that need significant changes but otherwise reusing things for basic packet structure and such.

ORelio:
Seems good, and we definitely need an abstract class for the basic read and write operations.

ORelio:
Ok, wrote support for 1.9 into Protocol18, so currently no need for copying and pasting the whole class. Of course better refactoring is welcomed.

PR #160, opened 2016-06-11
Implement terrain and movement for 1.9

Pokechu22:
This implements the new chunk format used by 1.9, allowing terrain and movement to also be there.

However, I should note: it's really slow for some reason.  I'm not entirely sure what makes it so slow, but logging in takes about 30 seconds on my computer (during which no keep alives are received...).  It may need some revision to solve this slowness, though there is no obvious performance bottleneck.

Changes:

* Implement the teleport confirm packet, needed to allow moving in 1.9
* Store the current dimension, needed to determine whether skylight is sent
* Handle the clientbound respawn packet, because that's used when dimensions change (eg `/world`).
* Most importantly, deserialize the new [SMP map format](http://wiki.vg/SMP_Map_Format).

### DIFF
--- a/MinecraftClient/Protocol/Handlers/Protocol18.cs
+++ b/MinecraftClient/Protocol/Handlers/Protocol18.cs
@@ -20,6 +20,7 @@ namespace MinecraftClient.Protocol.Handlers
     {
         private const int MC18Version = 47;
         private const int MC19Version = 107;
+        private const int MC191Version = 108;
         private const int MC110Version = 210;
 
         private int compression_treshold = 0;
@@ -37,6 +38,8 @@ namespace MinecraftClient.Protocol.Handlers
         Thread netRead;
         IAesStream s;
         TcpClient c;
+
+        int currentDimension;
 
         public Protocol18Handler(TcpClient Client, int ProtocolVersion, IMinecraftComHandler Handler, ForgeInfo ForgeInfo)
         {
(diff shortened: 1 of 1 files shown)


ORelio:
Awesome! And well implemented. Will take a look at the timings. In 1.8 that was caused by the byte array used for storing unread packet data, maybe there is something like that in 1.9+?

ORelio:
Okay, using the following code:

````C#
case PacketIncomingType.ChunkData:
    if (Settings.TerrainAndMovements)
    {
        ConsoleIO.WriteLineFormatted(DateTime.Now.ToString("§8mm-ss:fff IN"));
        //[...]
        ProcessChunkColumnData(chunkX, chunkZ, chunkMask, false, chunksContinuous, packetData);
        ConsoleIO.WriteLineFormatted(DateTime.Now.ToString("§8mm-ss:fff OUT"));
    }
    break;
````
It appears that chunk loading is actually very fast:
````
//[...]
13-43:850 IN
13-43:854 OUT
13-43:855 IN
13-43:858 OUT
13-43:859 IN
13-43:863 OUT
13-43:863 IN
13-43:868 OUT
13-43:869 IN
13-43:873 OUT
13-43:873 IN
13-43:877 OUT
13-43:877 IN
13-43:881 OUT
13-43:882 IN
13-43:886 OUT
Welcome, TestBot!
````
Each chunk is processed in a couple of milliseconds (4ms on my side).
This have been tested in spigot 1.9.4 overworld and vanilla 1.10 overworld and nether.
So, well, if you're experiencing slowdowns I cannot reproduce them on my side :/

PR #169, opened 2016-06-26
Extend aliases for use in MinecraftClient.ini file

initsuj:
Closed previous pull request because it needed to be cleaned up. Login and ServerIP will be checked against alias files when they are loaded.


ORelio:
Was going to tell you alias were loaded after server IP but you already fixed that in the new pull request. Everything seems fine, thanks for your contribution :+1:

PR #218, opened 2016-11-14
Add 1.11 support

Pokechu22:
1.11 made no protocol changes that should affect this project ([here's the pre-release protocol page](http://wiki.vg/index.php?title=Pre-release_protocol&oldid=8249)).

Well, there is actually one change that _may_ affect it - the [Title packet](http://wiki.vg/index.php?title=Pre-release_protocol&oldid=8249#Title) now supports writing to the action bar just as the chat message packet does.  I didn't implement that, since the title packet is somewhat complicated, but it shouldn't cause any issues (I'm 99% certain that the chat message variant is still the primary variant used in-game, but I haven't fully checked).

### DIFF
--- a/MinecraftClient/Program.cs
+++ b/MinecraftClient/Program.cs
@@ -20,9 +20,9 @@ namespace MinecraftClient
         private static McTcpClient Client;
         public static string[] startupargs;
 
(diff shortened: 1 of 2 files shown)


ORelio:
Thanks for your pull request :+1:

PR #225, opened 2016-12-21
Add 1.11.1 support

Pokechu22:
Again, next to no changes in the protocol, so all that's needed is adding 1.11.1 to the supported version list.


ORelio:
Thanks, merged :)

PR #228, opened 2016-12-25
Add 1.11.2 support

NurdTurd:


ORelio:
Hello and thanks for your contribution.
However that won't work because these constants in Program.cs are only for display purposes, there is another file to edit, see https://github.com/ORelio/Minecraft-Console-Client/commit/bc0a11d55b2f9f176e5f656996936f51dc08fa97 . Don't currently have time for doing that myself, sorry.

ORelio:
Merged as it is complementary to #245

PR #234, opened 2017-01-05
Refreshing Matches File

ReOp14:
Whenever you make a change to the Matches file for the auto responding bot, you must restart MCC. Like scripts, you just have to save and run the scripts. But scripts take around half a second to load, just as the Matches file are instant responding.
What I'm asking for is refreshing the Matches file and lowering the delay of time for scripts to load.


ORelio:
Okay but why is this a pull request? :sweat_smile:
Scripts takes time to compile, it depends on your system speed.
There is already a caching mechanism to avoid compiling the same script twice ;)
Indeed, AutoRespond, and probably other bots, could be improved to reload config but that is currently not the case. Nice idea.

ReOp14:
Oh, well you said this in a issue.
```
You would need to do this every time MCC gets updated.
The best way is submitting a pull request so that it gets added to MCC :)
Just added it for you, just pull the changes and/or grab the latest development build
```
I thought it was an idea submission place. You could add like /reload to reload things without exiting the mc server. My laptop is really old so I understand if its slow.

ORelio:
Okay so I should have explained that better. Pull requests are for submitting your own code so that I'll review your changes and, if everything is good, merge the pull request so that your code gets added to MCC to be part of all subsequent development builds and releases. Simply put, it's a way to submit your new features to the project. Issues are indeed more suited for ideas.

Yes, sorry for the slowness, compiling a script is done by the .NET framework directly and I can't optimize this that much. This is usually fast, less than half a second, but can get pretty slow on old computers. The cache has been added when I noticed compiling a script was taking around 10 (!!) seconds on a Raspberry Pi (1st gen). So doing `/script <script>` annoyingly slow. The cache allow re-launching the script immediately without recompiling after it has been loaded once.

PR #236, opened 2017-01-06
Add reload config option

Shiraz2014:
I'd like there to be a command inside MCC that reloads MinecraftClient.ini Kickmessages.txt basically all the files into the config so I do not have to reboot MCC every little config change. Thanks!


ORelio:
Please open pull requests only if you have some code to submit to the repository in order to get your changes added ;)

ORelio:
Regarding your suggestion you can add it to the [issues](https://github.com/ORelio/Minecraft-Console-Client/issues) section :wink:

PR #245, opened 2017-02-06
add 1.11.2 to supported protocol versions

nokimaro:
protocol version of 1.11.2 is the same as 1.11.1 = 316
http://wiki.vg/Protocol_version_numbers


ORelio:
Thanks, added :+1:

PR #284, opened 2017-06-10
Update to 1.12

Pokechu22:
This adds support for Minecraft 1.12's protocol, and also abstracts out outgoing packet IDs.

I include packet IDs for some 1.12 snapshots, even though snapshot versions aren't supported, because I already had written the code (though in a far more messy variant) for my [MC-106551](https://bugs.mojang.com/browse/MC-106551 '[MC-106551] Concurrency Issue in ThreadedAnvilChunkStorage / crash when using "/save-all flush"') test script; while they won't be used right now, they may be useful if snapshots are ever formally supported in the future.

I also did some minor cleanup of the README.

### DIFF
--- a/MinecraftClient/Program.cs
+++ b/MinecraftClient/Program.cs
@@ -20,9 +20,9 @@ namespace MinecraftClient
         private static McTcpClient Client;
         public static string[] startupargs;
 
-        public const string Version = "1.11.0 DEV";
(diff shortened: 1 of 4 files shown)


ORelio:
Nice & clean, great! Thanks for README fixes :+1:

PR #306, opened 2017-08-03
Implement 1.12.1 protocol

Pokechu22:
Already had this set up for testing [MC-106551](https://bugs.mojang.com/browse/MC-106551 '[MC-106551] Concurrency Issue in ThreadedAnvilChunkStorage / crash when using "/save-all flush"').  A few packet IDs changed in the first snapshot, but that's it.


ORelio:
Thanks for your contribution :+1:

PR #313, opened 2017-09-09
Fix packet compression (+ 1 more)

Pokechu22:
This PR does two things:

* Fixes packet compression when MCC is encoding packets.  Broken packet compression was what was causing MCC to connect to forge servers - a large outbound packet was being sent (the mod list), and that failed to compress correctly.  (I'm not entirely sure why I never had any problems with this before during the first set of forge tests, not to mention that this hasn't ever caused problems elsewhere)

    There were 3 problems:

    1. The compression threshold was compared with `>` when it should have been `>=`.  This didn't cause major problems since it's a rare case.
    2. The compressed packet length was being used when the uncompressed length should have been used.  This broke things severely, but not as much as the next problem:
    3. Only 2 or so bytes of compressed data would ever be sent, due to `ZlibUtils.Compress` being implemented strangely.  That implementation only filled the buffer once or twice (and changing to `ReadByte` still had the same issue).  Writing to a memory stream worked correctly, and now compressed data is actually sent.

*  Stops hardcoding the ID of TeleportConfirm. This was a change that I made earlier but never pushed, and I'm just including it here to get it out.  Always sending the teleport confirm packet even when terrain and movements is disabled is simply needed to ensure proper server behavior (it's minor, and I can't easily give examples of what would happen if not done, but it is still needed).

### DIFF
--- a/MinecraftClient/Protocol/Handlers/Protocol18.cs
+++ b/MinecraftClient/Protocol/Handlers/Protocol18.cs
@@ -1421,12 +1421,10 @@ namespace MinecraftClient.Protocol.Handlers
 
             if (compression_treshold > 0) //Compression enabled?
             {
-                if (the_packet.Length > compression_treshold) //Packet long enough for compressing?
+                if (the_packet.Length >= compression_treshold) //Packet long enough for compressing?
                 {
-                    byte[] uncompressed_length = getVarInt(the_packet.Length);
                     byte[] compressed_packet = ZlibUtils.Compress(the_packet);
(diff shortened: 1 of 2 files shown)


ORelio:
Seems perfect. Thanks and congrats for your investigations :slightly_smiling_face:

Barnehhh:
Very nice, I believe this was the issue I was having weeks back and this may potentially be the fix for it! Still doing some testing to make sure, but thank you. :)

PR #314, opened 2017-09-11
Switch console window encoding to Unicode

Barnehhh:
Unicode characters are becoming quite common within minecraft servers (especially the large ones) nowadays. As of now MCC creates a System beep every single time one of these characters are processed by ConsoleIO. 

I'm no expert in C#, so I'm unsure if ANSI is required for anything, but I've been perfectly fine with UTF-8 so far.

(I'm still not very familiar with github so apologies if theres a duplicate of this request)

### DIFF
--- a/MinecraftClient/Program.cs
+++ b/MinecraftClient/Program.cs
@@ -45,9 +45,9 @@ namespace MinecraftClient
             if (args.Length >= 1 && args[args.Length - 1] == "BasicIO")
             {
                 ConsoleIO.basicIO = true;
-                Console.OutputEncoding = Console.InputEncoding = Encoding.UTF8;
                 args = args.Where(o => !Object.ReferenceEquals(o, args[args.Length - 1])).ToArray();
             }
+            Console.OutputEncoding = Console.InputEncoding = Encoding.UTF8;
 
             //Process ini configuration file
             if (args.Length >= 1 && System.IO.File.Exists(args[0]) && System.IO.Path.GetExtension(args[0]).ToLower() == ".ini")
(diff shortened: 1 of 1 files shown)


Pokechu22:
Tried this out, and while switching to UTF-8 seems to be a good idea, this PR only changes it when basic IO is enabled - the issue still is present when it's not enabled (which is the default).  (n.b. I assumed that I was missing a font which is why I didn't notice any encoding issues).  Always changing it would be better, instead of just doing it when basic IO is enabled:

```
            if (args.Length >= 1 && args[args.Length - 1] == "BasicIO")
            {
                ConsoleIO.basicIO = true;
                args = args.Where(o => !Object.ReferenceEquals(o, args[args.Length - 1])).ToArray();
            }
            Console.OutputEncoding = Console.InputEncoding = Encoding.UTF8;
```

I also am not sure why ANSI was being used (@orelio would know), but I think UTF-8 seems reasonable.

Barnehhh:
Ah you're right, thank you. Late here so didn't notice.  Should be updated now I assume.

Pokechu22:
Thanks, everything looks good to me (just need orelio to merge).  Thank you for your contribution!

ORelio:
That ANSI encoding line was to make sure both the GUI and MCC would use the same encoding, I don't remember the exact issue ID, though, and UTF-8 seems also a safe bet (the GUI should also contain a similar line that would need to be changed).

Regarding encoding outside of BasicIO well, I didn't know you could switch the old cmd.exe to UTF-8 :joy:
Thanks for your contribution !

PR #321, opened 2017-09-18
Add support for 1.12.2

Pokechu22:
While there are protocol changes to the keep alive packet, the way MCC implements it (sending the same bytes to the server) should mean that the changes do not affect us.


ORelio:
Thanks :)

PR #325, opened 2017-10-01
Fix README.md formatting

UltraX3:


ORelio:
Thanks for your contribution, this should already be fixed in the `Indev` branch, which has not made its way to the `master` branch yet :wink:

PR #467, opened 2018-05-25
Fix gravity handling.

CarbonNeuron:
This was probably just a typo, or just forgotten about. But this makes clients respect gravity again.


CarbonNeuron:
This worked for me for 5 mins. Then stopped working.

PR #468, opened 2018-05-25
GetPing AutoTimeout timeout delay change.

Barnehhh:
Increase the time of the fast srvrecord setting from 5 to 10 seconds. As of now the method gets timed out before actually completing/attempting the Protocol18 DoPing causing auto to fail on most servers.


ORelio:
Sorry for the delay, thanks for your contribution. The delay was set because the SRV lookup query often failed and timed out after 30 seconds, hopefully that should be fixed now.

PR #530, opened 2018-07-26
Chinese Transaltion for README

WongKaYong:
Do you mind if I help translating some parts that need translation? because I see that you don't have any words translate to Chinese language, maybe some users might prefer them? Thanks.
@ORelio


ORelio:
Thanks for your contribution :+1:

Issue #525, opened 2018-07-22
Roadmap for 1.13 support (v393 protocol)


Dids:
Any word on when (if ever) we could expect to see initial support for 1.13?

I primarly use an alt account as a chunk loader, so as long as the account is logged in and stays that way, that's all the functionality I'd need.

ORelio:
Unfortunately I have less and less time available to upgrade MCC for each Minecraft version. Any contribution is welcome!

Thankos for your offer. Usually, upgrading MCC for a new Minecraft version involves the following step:
 - Figuring out what packets changed by performing a diff on http://wiki.vg/Protocol
 - Implementing the changes in [Protocol18.cs](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Protocol/Handlers/Protocol18.cs) while staying backward-compatible
 - Marking the MC version as supported in [ProtocolHandler.cs](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Protocol/ProtocolHandler.cs)

I can provide more in-depth explanations if someone is willing to contribute :wink:

gandiber:
@ORelio I have 6 years of experience with C# and actually have been working on a private version of your program for my own needs. I replaced your old proxy library with a solution that seems to work better. I'd definitely be willing to help, but I need more information. Do you have a better way of contact?

Thanks!

ORelio:
Thanks for your offer. I emailed some contact information to your Git email. Let me know if you didnt receive.

TheSnoozer:
I have a local version prepared that almost is finished for 1.13.

However I ran currently into some weird issues when I have ``terrainandmovements=true``. Haven't had the time yet to fully troubleshoot it yet.

iSolarNZ:
> I have a local version prepared that almost is finished for 1.13.
> 
> However I ran currently into some weird issues when I have `terrainandmovements=true`. Haven't had the time yet to fully troubleshoot it yet.

Awesome! Would it be possible for you to share it with me once you've fully completed the update?

TheSnoozer:
I will create a merge request once I have finished it and then hopefully will end up in this repository (but that would be out of my control).

iSolarNZ:
> I will create a merge request once I have finished it and then hopefully will end up in this repository (but that would be out of my control).

Alright, cool. Thank you 👍

Pokechu22:
[...]

PR #536, opened 2018-08-04
Update ProtocolHandler.cs

NurdTurd:


ORelio:
Is there really no change to do other than marking 1.13 as supported? No packet changes to implement?

ORelio:
(See #525)

PR #563, opened 2018-09-25
525: general 1.13 support (v393 protocol)

TheSnoozer:
This MR adds a general 1.13 support. Using this version doesn't crash the client, but in particular for `terrainandmovements=true` there might be some major outstanding work.

See the relevant comment here:
https://github.com/ORelio/Minecraft-Console-Client/issues/525#issuecomment-421668400

Just FYI:
I replaced some methods like this
```c#
byte[] message_val = Encoding.UTF8.GetBytes(message);
byte[] message_len = getVarInt(message_val.Length);
byte[] data = concatBytes(message_len, message_val);
```

with a simpler
```
byte[] data = getString(message);
```
if I'm not mistaken should do the same.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /MinecraftClientGUI.suo
 /MinecraftClient.userprefs
 /.vs/
+SessionCache.ini
--- a/MinecraftClient/Program.cs
+++ b/MinecraftClient/Program.cs
@@ -23,7 +23,7 @@ namespace MinecraftClient
 
         public const string Version = MCHighestVersion;
         public const string MCLowestVersion = "1.4.6";
-        public const string MCHighestVersion = "1.13";
+        public const string MCHighestVersion = "1.13.1";
         public static readonly string BuildInfo = null;
 
         private static Thread offlinePrompt = null;
--- a/MinecraftClient/Program.cs
+++ b/MinecraftClient/Program.cs
@@ -23,7 +23,7 @@ namespace MinecraftClient
 
         public const string Version = MCHighestVersion;
         public const string MCLowestVersion = "1.4.6";
-        public const string MCHighestVersion = "1.12.2";
+        public const string MCHighestVersion = "1.13";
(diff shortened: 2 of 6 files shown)


BlinkSun:
Great work @TheSnoozer  !!! I'll try to follow your "change" to try to include 1.13.1 ! The protocol for 1.13.1 is 401 if i believe what i saw... !!! Thanks again for your nice work !

TheSnoozer:
Unfortunately due to the lack of time I haven't looked into any specifics what changed during the protocols.
To get yourself an quite decent overview I'd recommend https://joodicator.github.io/mc-dev-data/

It highlights when packets have changed and saves a ton of time going through the wiki manually.

Pokechu22:
Check [protocol history](https://wiki.vg/Protocol_History#1.13.1) and [burger](https://pokechu22.github.io/Burger/diff_1.13_1.13.1.html#packets).  The only changes (apart from the protocol version) are to packets that MCC doesn't use; supporting 1.13.1 should be as simple as adding 1.13.1 to the version list.

TheSnoozer:
There doesn't seem to be any packet changes this client should be concerned about.

I added the relevant changes (haven't tested them with an actual server).

ORelio:
Thanks for your contribution!

Yes, `getString` should do the same thing, it was added later so not used everywhere.
FYI `concatBytes` can take more than 2 arrays at the same time so you can merge multiple calls into one.
Also, you do not need to handle specific protocol implementation for snapshots since only the final release is handled by the client. You can merge everything into one same version and use `if (protocol > MC113Version)` directly.

Apart from that everything sounds good, merging :+1:

ORelio:
(And that was the 100th pull request. Congrats!)

TheSnoozer:
Thanks for the review and getting this integrated :-)

gandiber:
Yeah terraincontrol unfortunately no longer works in 1.13 at all.

PR #577, opened 2018-10-16
Better Chinese Translation

l89669:


MinecraftClient/Protocol/ProtocolHandler.cs
```diff
@@ -297,7 +299,7 @@ public static LoginResult GetTokenValidation(SessionToken session)
             {
                 string result = "";
                 string json_request = "{\"accessToken\": \"" + JsonEncode(session.ID) + "\", \"clientToken\": \"" + JsonEncode(session.ClientID) + "\" }";
-                int code = DoHTTPSPost("authserver.mojang.com", "/validate", json_request, ref result);
+                int code = DoHTTPSPost("summermc.cc", "/api/yggdrasil/validate", json_request, ref result);
```
TheSnoozer: What does this have to do with a "better translation"?
Why is it needed anyways?

l89669: I'm sorry, only the first commit is for this pull request, I forgot to switch branch so there's other weired commits.

PR #689, opened 2019-03-30
WIP

TheSnoozer:
Marked as WIP for now, still want to change some major things around.
See individual commits for details.

### DIFF
--- a/MinecraftClient/CSharpRunner.cs
+++ b/MinecraftClient/CSharpRunner.cs
@@ -274,6 +274,18 @@ namespace MinecraftClient
             return base.GetOnlinePlayers();
         }
 
+        /// <summary>
+        /// Get a dictionary of online player names and their corresponding UUID
+        /// </summary>
+        /// <returns>
+        ///     dictionary of online player whereby
+        ///     UUID represents the key
+        ///     playername represents the value</returns>
+        new public Dictionary<string, string> GetOnlinePlayersWithUUID()
+        {
+            return base.GetOnlinePlayersWithUUID();
+        }
+
         /* == Additional Methods useful for Script API == */
 
         /// <summary>
--- a/MinecraftClient/Protocol/Handlers/Protocol18.cs
+++ b/MinecraftClient/Protocol/Handlers/Protocol18.cs
@@ -159,12 +159,13 @@ namespace MinecraftClient.Protocol.Handlers
                 return PacketIncomingType.KeepAlive;
             else if (packetID == JoinGame.getPacketID(protocol))
                 return PacketIncomingType.JoinGame;
+            else if (packetID == ChatMessage.getPacketID(protocol))
+                return PacketIncomingType.ChatMessage;
 
             if (protocol < PacketUtils.MC19Version)
             {
                 switch (packetID)
                 {
-                    case 0x02: return PacketIncomingType.ChatMessage;
                     case 0x07: return PacketIncomingType.Respawn;
                     case 0x08: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x21: return PacketIncomingType.ChunkData;
@@ -185,7 +186,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x0F: return PacketIncomingType.ChatMessage;
                     case 0x33: return PacketIncomingType.Respawn;
                     case 0x2E: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x20: return PacketIncomingType.ChunkData;
@@ -206,7 +206,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x10: return PacketIncomingType.ChatMessage;
                     case 0x35: return PacketIncomingType.Respawn;
                     case 0x2F: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x21: return PacketIncomingType.ChunkData;
@@ -227,7 +226,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x0F: return PacketIncomingType.ChatMessage;
                     case 0x34: return PacketIncomingType.Respawn;
                     case 0x2E: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x20: return PacketIncomingType.ChunkData;
@@ -248,7 +246,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x0F: return PacketIncomingType.ChatMessage;
                     case 0x35: return PacketIncomingType.Respawn;
                     case 0x2F: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x20: return PacketIncomingType.ChunkData;
@@ -269,7 +266,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x0E: return PacketIncomingType.ChatMessage;
                     case 0x35: return PacketIncomingType.Respawn;
                     case 0x2F: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x21: return PacketIncomingType.ChunkData;
@@ -290,7 +286,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x0E: return PacketIncomingType.ChatMessage;
                     case 0x36: return PacketIncomingType.Respawn;
                     case 0x30: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x21: return PacketIncomingType.ChunkData;
@@ -311,7 +306,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x0E: return PacketIncomingType.ChatMessage;
                     case 0x37: return PacketIncomingType.Respawn;
                     case 0x31: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x21: return PacketIncomingType.ChunkData;
@@ -332,7 +326,6 @@ namespace MinecraftClient.Protocol.Handlers
             {
                 switch (packetID)
                 {
-                    case 0x0E: return PacketIncomingType.ChatMessage;
                     case 0x38: return PacketIncomingType.Respawn;
                     case 0x32: return PacketIncomingType.PlayerPositionAndLook;
                     case 0x22: return PacketIncomingType.ChunkData;
--- a/MinecraftClient/Protocol/Handlers/Protocol18.cs
+++ b/MinecraftClient/Protocol/Handlers/Protocol18.cs
@@ -225,12 +225,13 @@ namespace MinecraftClient.Protocol.Handlers
                 return ServerChatMessage.getPacketID(protocol);
             else if (packet == PacketOutgoingType.ClientStatus)
                 return ServerClientStatus.getPacketID(protocol);
+            else if (packet == PacketOutgoingType.ClientSettings)
+                return ServerClientSettings.getPacketID(protocol);
 
             if (protocol < PacketUtils.MC19Version)
             {
                 switch (packet)
                 {
-                    case PacketOutgoingType.ClientSettings: return 0x15;
                     case PacketOutgoingType.PluginMessage: return 0x17;
                     case PacketOutgoingType.TabComplete: return 0x14;
                     case PacketOutgoingType.PlayerPosition: return 0x04;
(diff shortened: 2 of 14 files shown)


MinecraftClient/CSharpRunner.cs
```diff
@@ -274,6 +274,18 @@ public bool SendText(object text)
...
         }
 
+        /// <summary>
+        /// Get a dictionary of online player names and their corresponding UUID
+        /// </summary>
+        /// <returns>
+        ///     dictionary of online player whereby
+        ///     UUID represents the key
+        ///     playername represents the value</returns>
+        new public Dictionary<string, string> GetOnlinePlayersWithUUID()
```
TheSnoozer: I have tested this new method with the following script
```
//MCCScript 1.0

/* This is a sample script that will load a ChatBot into Minecraft Console Client
 * Simply execute the script once with /script or the script scheduler to load the bot */

MCC.LoadBot(new ExampleBot());

//MCCScript Extensions

/* The ChatBot class must be defined as an extension of the script in the Extensions section
 * The class can override common methods from ChatBot.cs, take a look at MCC's source code */

public class ExampleBot : ChatBot
{
    public override void Initialize()
    {
        LogToConsole("Sucessfully Initialized!");
    }

    public override void GetText(string text)
    {
        LogToConsole("Message: " + GetVerbatim(text));
        Dictionary<string, string> uuid2Player = GetOnlinePlayersWithUUID();
        foreach (string uuid in uuid2Player.Keys)
        {
            LogToConsole("UUID: " + uuid + " / username: " + uuid2Player[uuid]);
        }
    }
}
```

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -1367,7 +1367,39 @@ private static ushort[] readNextUShortsLittleEndian(int amount, List<byte> cache
...
+                guid = new Guid(javaUUID);
+            }
+            return guid;
+        }
+
+        /// <summary>
+        /// convert a Java big-endian Guid to a .NET little-endian Guid.
+        /// </summary>
+        /// <returns>GUID in little endian.</returns>
+        private static Guid ToLittleEndian(byte[] java)
```
TheSnoozer: I'm confident there are better and smarter versions on how do to that conversion in C#. In all honesty I just took the first [example](https://stackoverflow.com/a/10149822/7890667) that worked :P

TheSnoozer:
@ORelio I know this is not done yet, but I'd like to ask you to take a look at the general format/structure I was thinking for a while now. In general I was thinking to cleanup the huge switch-case statements in `Protocol18.cs` for determining the packet ID and handling the packet. In theory each packet is represented in it's own class and thus getting the ID and handling the packet is implemented in each class and not just represented as switch-case inside `Protocol18.cs`.

WDYT?

ORelio:
Hi @TheSnoozer,

Since this PR is starting to get huge and cannot be merged anymore, I'd suggest to resubmit UUID-related changes separately as they seems relevant and well implemented, so that we can merge them first.

Regarding protocol refactoring, this seems a good idea to move field encoding/decoding stuff from `Protocol18` in a separate class such as `PacketUtils`. Additionally, in Vanilla Minecraft, each packet has its own class in a separate file with id and encoding/decoding utilities so that it can decode/encode itself. At first, this approach was overkill since MCC was chat-only and only processed a few packets in the switch case (see how small is the `Protocol16` switch), however this seems more relevant now that MCC also handles terrain, Forge and other stuff, and the `Protocol18` handler is getting huge.

When logging in, a dictionary with packet ID (`int`) as key and `PacketHandler` as value could be created with correct mappings for that version, so that upon receiving a packet, the correct packet handler is called, or the packet is ignored entierely if absent from the dictionary (unimplemented packet). This way, each packet would have its own class in its own file and would process its own events and data. When the packet is too different, maybe a new packet handler could be done the same way as `Protocol16` and `Protocol18` so that the correct one will be called while maintaining clear code for each one (minimizing the numerous `if (protocol > version)` statements that would be kept for small differences only).

(Also, I think you do not need to implement this for `Protocol16`, which has not really changed since 1.7 and will likely not benefit from the refactoring)

vkorn:
Hi all, sorry for highjacking the thread, but I think this PR is the best place to discuss this. I've started to look into Console Client today and it has a great idea behind. However, adding a new version support is not quite a straightforward process right now. Even tho this PR brings more clean approach, still there's a roof for an error. Having a lot of conditions in packets handling is not "beginner friendly". 

For example, with recent changes JoinGame looks like this

```
             case PacketIncomingType.JoinGame:
                    handler.OnGameJoined();
                    PacketUtils.readNextInt(packetData); // entity ID
                    PacketUtils.readNextByte(packetData); // Game mode
                    if (protocolversion >= PacketUtils.MC191Version)
                        currentDimension = PacketUtils.readNextInt(packetData); //dimension
                    else
                        currentDimension = (sbyte)PacketUtils.readNextByte(packetData);
                    
                    if (protocolversion < PacketUtils.MC114pre5Version)
                        PacketUtils.readNextByte(packetData); // Difficulty 
                    PacketUtils.readNextByte(packetData); // Max players
                    PacketUtils.readNextString(packetData); // Level type
                    if (protocolversion >= PacketUtils.MC114pre5Version)
                        PacketUtils.readNextVarInt(packetData); // View distance
                    
                    if (protocolversion >= PacketUtils.MC18Version)
                        PacketUtils.readNextBool(packetData);  // Reduced debug info - 1.8 and above
```

Maybe it makes sense to extract packets handlers for specific version? So instead of generic `Protocol18Handler` you can have a factory which will initialize proper version handler? 

Something like this 
```
internal struct JoinGameResult
    {
        public int Dimension { get; internal set; }
        public int ViewDistance { get; internal set; }
    }
    
    internal abstract class JoinGamePacket
    {
        public virtual JoinGameResult Handle(IMinecraftComHandler handler, List<byte> packetData)
        {
            var res = new JoinGameResult();
            handler.OnGameJoined();
            ReadEntityID(packetData);
            ReadGameMode(packetData);
            res.Dimension = ReadDimension(packetData);
            ReadDifficulty(packetData);
            ReadMaxPlayers(packetData);
            ReadLevelType(packetData);
            ReadReducedDebugInfo(packetData);
            return res;
        }

        protected virtual int ReadEntityID(List<byte> packetData)
        {
            return PacketUtils.readNextInt(packetData);
        }

        protected virtual byte ReadGameMode(List<byte> packetData)
        {
            return PacketUtils.readNextByte(packetData);
        }

        protected virtual byte ReadDifficulty(List<byte> packetData)
        {
            return PacketUtils.readNextByte(packetData);
        }

        protected int ReadDimension(List<byte> packetData)
        {
            return PacketUtils.readNextInt(packetData);
        }

        protected virtual byte ReadMaxPlayers(List<byte> packetData)
        {
            return PacketUtils.readNextByte(packetData);
        }

        protected virtual string ReadLevelType(List<byte> packetData)
        {
            return PacketUtils.readNextString(packetData);
        }

        protected virtual bool ReadReducedDebugInfo(List<byte> packetData)
        {
            return PacketUtils.readNextBool(packetData);
        } 
    }
```

So now with the changes for a new version I have couple options: 
1. Override Handle method in my version's protocol or 
1. Add a new property method which will do nothing for older versions and implement it in current version

E.g. adding
```
          ReadLevelType(packetData);
            ReadViewDistance(packetData)
            ReadReducedDebugInfo(packetData);
```

where 

```
 protected virtual int ReadViewDistance(List<byte> packetData)
        {
            return 0;
        }

```

Then for my version I could define such handler 
```
internal class JoinGamePacket114 : JoinGamePacket
    {
        protected override byte ReadDifficulty(List<byte> packetData)
        {
            return 0;
        }

        protected override int ReadViewDistance(List<byte> packetData)
        {
            return PacketUtils.readNextVarInt(packetData);
        }
    }
```

Instantiate it through factory

```
    internal static class JoinGamePacketFactory
    {
        public static JoinGamePacket Instantiate(int protocolVersion)
        {
            if (protocolVersion < PacketUtils.MC114pre5Version)
            {
                return new JoinGamePacketAnyOld();
            } 
            
            return new JoinGamePacket114();
        }
    }
```

And have a very short and clean handler

```
case PacketIncomingType.JoinGame:
                    var h = JoinGamePacketFactory.Instantiate(protocolversion);
                    var res = h.Handle(handler, packetData);
                    currentDimension = res.Dimension;
                    break;
```

Of course it makes sense to init once and just reuse, this is just an example. 


P.S. Sorry for a long post and thanks again for your awesome work!

vkorn:
To elaborate further, you can "chain" versions with such approach. Defining something like this
```
    internal interface IJoinGamePacket
    {
        JoinGameResult Handle(IMinecraftComHandler handler, List<byte> packetData);
        int MinVersion();
    }
```

and implementing 3 version for this particular case

```
 internal class JoinGamePacket18 : JoinGamePacket
    {
        public override int MinVersion()
        {
            return PacketUtils.MC18Version;
        }

        protected override bool ReadReducedDebugInfo(List<byte> packetData)
        {
            return PacketUtils.readNextBool(packetData);
        }
    }

    internal class JoinGamePacket191 : JoinGamePacket18
    {
        public override int MinVersion()
        {
            return PacketUtils.MC191Version;
        }

        protected override int ReadDimension(List<byte> packetData)
        {
            return PacketUtils.readNextInt(packetData);
        }
    }


    internal class JoinGamePacket114 : JoinGamePacket191
    {
        public override int MinVersion()
        {
            return PacketUtils.MC114pre5Version;
        }

        protected override byte ReadDifficulty(List<byte> packetData)
        {
            return 0;
        }

        protected override int ReadViewDistance(List<byte> packetData)
        {
            return PacketUtils.readNextVarInt(packetData);
        }
    }
```

You can use generic factory after 

```
 internal static class JoinGamePacketFactory
    {
        public static IJoinGamePacket Instantiate(int protocolVersion)
        {
            var versions = new SortedDictionary<int, IJoinGamePacket>();

            foreach (var t in Assembly.GetExecutingAssembly().GetTypes()
                .Where(x => x.GetInterfaces().Contains(typeof(IJoinGamePacket))))
            {
                var i = (IJoinGamePacket) Activator.CreateInstance(t);
                versions.Add(i.MinVersion(), i);
            }

            foreach (var pair in versions.OrderByDescending(x => x.Key))
            {
                if (protocolVersion >= pair.Key)
                {
                    return pair.Value;
                }
            }

            throw new NotSupportedException();
        }
    }
```

TheSnoozer:
> a dictionary with packet ID (`int`) as key and `PacketHandler` as value could be created with correct mappings for that version

Exactly this was my idea :-)

However I can remember that at some point I got stuck since I needed to feed the `LookupDictionary` with an instance of the actual packet. However for the client side packets we essentially just reading the packet (somewhat the reading is a factory method that creates the actual packet class that matches the packet id). In a nutshell I wanted to get something like this https://stackoverflow.com/a/47280772/7890667 or https://stackoverflow.com/a/25021319/7890667. Maybe I missed something along the lines...



In regards to @vkorn comment: I can't agree more. It currently requires a certain amount of will to step through the complexity of the code. In some cases I would certainly agree that a separate method would make sense depending on the protocol version. However I think in most cases you could rewrite the code in such a way that the code speaks to you without too much overhead.

Illustrative example:

```
class ClientJoinGame {
      public ClientJoinGame readPacket() {
          // handler.OnGameJoined();
          this.EntityId = PacketUtils.readNextInt(packetData);
          this.GameMode = PacketUtils.readNextByte(packetData);
          if (protocolversion >= PacketUtils.MC191Version)
              this.Dimension = PacketUtils.readNextInt(packetData);
          else
              this.Dimension = (sbyte)PacketUtils.readNextByte(packetData);
          
          if (protocolversion < PacketUtils.MC114pre5Version)
              this.Difficulty = PacketUtils.readNextByte(packetData);
          this.MaxPlayers = PacketUtils.readNextByte(packetData);
          this.LevelType = PacketUtils.readNextString(packetData);
          if (protocolversion >= PacketUtils.MC114pre5Version)
              this.ViewDistance = PacketUtils.readNextVarInt(packetData);
          
          if (protocolversion >= PacketUtils.MC18Version)
              this.IdkWhatThisIs = PacketUtils.readNextBool(packetData);
      }
}
```

Pokechu22:
I want to say I tried doing something like this back when I was a more active contributor, but I'm not sure.

In Java, I'd do something like this nowadays:

```java
Map<Integer, Function<byte[], Packet>> PACKETS_BY_ID = new HashMap<>();
PACKETS_BY_ID.put(1, SomePacket::new);
```

where `SomePacket::new` references the constructor - like `new SomePacket(byte[] bytes)` or so.  I don't know if C#'s lambdas have the same kind of expressiveness for this, or if you're stuck doing factories manually.  Though it does get complicated either way...

vkorn:
@TheSnoozer chained conditions is always hard to read :) 
@Pokechu22 you can use reflection, which is fully implemented by core/mono nowadays. For example https://github.com/vkorn/Minecraft-Console-Client/blob/various-features/MinecraftClient/Protocol/Packets/PacketFactory.cs. I did a quick refactoring just to check an idea which I've described above, you can check 1.14 commit in a fork.

TheSnoozer:
I'm currently not sure what would be the best strategy to proceed. It certainly would make sense to create smaller chunks of code changes that are more digestible. I certainly will have a look at the code from @vkorn and I think at this stage I just create a new MR.

PR #697, opened 2019-04-10
/look command.

repository:
My first working with a C language _**ever**_, with the exception of some Arduino stuff, so apologies for any sloppy code.

### DIFF
--- a/MinecraftClient/Commands/Look.cs
+++ b/MinecraftClient/Commands/Look.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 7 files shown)


Mhowser:
Will it be possible to make the MCC player look at a specific entity/player constantly, like those player NPC plugins?

repository:
Probably not, it requires knowing where other entities are and I don't think the client does this currently, and can't be bothered to figure out how to do it.

ORelio:
Thank you for your contribution! For a first C# contribution it's way cleaner that you could think 😉 
Only one thing: C#6 Interpolated Strings are not C#4 compliant so I'll fix that for compatibility.

Issue #625, opened 2019-01-22
Getting a Player's UUID


FoamyTheFag:
I would like to create a ChatBot where it gets a user's UUID and stores it. I have already figured out everything else, besides how to actually get a player's UUID from their in game name.

I tried making my own string in McTcpClient with a parameter for ign, however it always returns null and I can't think of another way to get a player's UUID.

TheSnoozer:
Mhh interesting.
Inside a MCC-Script you currently could use the [MCC.GetOnlinePlayers()](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/ChatBot.cs#L655) method to get all the player names, but it does not return the UUIDs.
We currently maintain a dict of UUID <-> playerName relation inside the ``onlinePlayers`` variable, so I guess there is currently no such thing where you get the data from.

However it should be an easy extension to get the information you need.

ORelio:
Now implemented thanks to @TheSnoozer contribution. Check it out :wink:

Issue #493, opened 2018-06-13
UUID not properly storing due to GUID conversion.


Barnehhh:
https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Protocol/Handlers/Protocol18.cs#L1185

https://i.imgur.com/kOkYfJp.png

UUID on top is what MCC by default currently gets for each online player, bottom is what that player's actual UUID is.

GUID's byte order for the first 3 fields differs from UUID causing it to scramble.

TheSnoozer:
Sorry for the delay, you are right the conversion is broken.

This is due Java's UUID being represented as big-endian Guid, while C# tries to interpret it as little-endian.

For a fix I used [the following conversion](https://stackoverflow.com/a/10149822/7890667)

There are various different options on how the conversion could happen (e.g. https://gist.github.com/MrZoidberg/9bac07cf3f5aa5896f75 or something along the lines of https://stackoverflow.com/a/18023109/7890667)

Example on how this is being broken:
```
before: 
[MCC] [ExampleBot] UUID: b69849b7-196e-3648-b23f-7b38464cf5ec / username: d4gobert 

after (correct UUID):
[MCC] [ExampleBot] UUID: b74998b6-6e19-4836-b23f-7b38464cf5ec / username: d4gobert 
```

ORelio:
@TheSnoozer Reviewed your coomit, seems all good! Feel free to add it to the main codebase using a pull request 😉

ORelio:
Fix is now included in latest development build thanks to @TheSnoozer :wink:
Feel free to reopen if it still does not work as expected.

PR #702, opened 2019-04-22
Fix wrong guid conversion

TheSnoozer:
As suggested, this is the MR to fix the UUID-related changes separately.

See individual commits for references.


ORelio:
Thanks for your contribution :wink:

Solves #493 and #625

Issue #599, opened 2018-11-21
Implement terrain handling for Minecraft 1.13


Mhowser:
https://github.com/cabaletta/baritone

Would it be possible to integrate this into MCC? Might solve /move not working on 1.13.

Edit: I was under the impression it was on 1.13 already, it's 1.12.2, but could be a stepping stone towards 1.13 movement support.

Mhowser:
I forgot that mcc was made in C# 😅 /move does "work" in 1.12.x, it could
certainly use some enhancing though.

ORelio:
Hi,
Currently, MCC implements pathfinding for moving in a world of blocks. It does not necessarily know about some special blocks such as slabs, but the pathfinding mechanism is still pretty generic.
However in 1.13 terrain format changed. MCC does not know how to process data from the server to build its internal representation of the world. As such, it cannot determine a path.
So to fix pathfinding we do not need to reimplement it, we rather need to implement terrain data decoding for Minecraft 1.13.

vkorn:
@ORelio I was able to fix terrain for 1.14, however, I'm using completely re-written approach, which we've discussed in https://github.com/ORelio/Minecraft-Console-Client/pull/689 
To avoid parsing palette files in memory (and carrying extra 2mb in the binary), I made a simple pre-generator (https://github.com/vkorn/Minecraft-Console-Client/tree/various-features/MaterialGenerator) for simplified list of mats https://github.com/vkorn/Minecraft-Console-Client/blob/various-features/MinecraftClient/Protocol/WorldProcessors/BlockProcessors/114Pre5/Material.cs 

Pretty much you can copy and paste this part into current implementation to fix terrain.

ORelio:
Hi @vkorn thanks for your work on this.

I've read how the 1.13+ palette thing works, and I think it's clever since it greatly reduces chunk data size, although not practical for our implementation which tries to be light and needs to have more and more static Mojang assets to work with newer MC versions (this reminds me of translation files for chat in MC 1.6).

[...]

PR #704, opened 2019-04-23
Extract gathering of packet-ids into individual classes

TheSnoozer:
After some thinking and further research on @vkorn's code I came to the conclusion that it's time to entirely ditch my previous approach and boil his changes down to what is needed. To emphasis - this is not my code, this is essentially the diff with limited noise (e.g. commented out code is removed instantly).

Note that this introduces a new project 'MaterialGenerator' that requires some external dependencies.
For me (linux user) this now boils down to:
```
rm -rf ./MinecraftClient/bin/
rm -rf ./MinecraftClient/obj/
rm -rf packages

DATE=`date +%Y-%m-%d`
COMMIT=`git rev-parse --short HEAD`
git checkout HEAD ./MinecraftClient/Properties/AssemblyInfo.cs
echo -e "\n[assembly: AssemblyConfiguration(\"Built on $DATE from commit $COMMIT\")]" >> ./MinecraftClient/Properties/AssemblyInfo.cs


# download required packages into packages:
# wget https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
mono nuget.exe restore MaterialGenerator/MaterialGenerator.csproj -PackagesDirectory packages

msbuild MinecraftClient.sln /t:build /p:Configuration=Debug /p:Platform="x86" /p:TargetFrameworkVersion="v4.0" /verbosity:minimal
msbuild MinecraftClient.sln /t:build /p:Configuration=Release /p:Platform="x86" /p:TargetFrameworkVersion="v4.0" /verbosity:minimal


git checkout HEAD ./MinecraftClient/Properties/AssemblyInfo.cs

echo "Launching Debug version"

mono ./MinecraftClient/bin/Debug/MinecraftClient.exe
```

Also the submodule needs to be initialized properly.


vkorn:
I would strongly suggest to avoid pushing this to master, at least because Forge handshake was commented out (I simply don't know what it is and how to test), plus still a few packets are not yet implemented -- I did only essentials so far:
```
        KeepAlive = 1,
        JoinGame = 2,
        ChatMessage = 3,
        Respawn = 4,
        PlayerPositionAndLook = 5,
        ChunkData = 6,
        MultiBlockChange = 7,
        BlockChange = 8,
        KickPacket = 14,
```

In addition, as @ORelio mentioned in https://github.com/ORelio/Minecraft-Console-Client/issues/599 this will break bots, if they relate on a Material. 

I can help with packets -- they're pretty fast to convert. For Materials, I think IBlock should expose string Name() which will return old material for pre-14 (or pre-13 if it's backported) versions and full minecraft:name for a new ones. This change is fairly easy as well. 

Submodule is, unfortunately, unavoidable, since it's the only NBT fork with LongArray type implementation and this PR was not merged into fNBT yet (thus, not pushed to nuget).

Pokechu22:
> at least because Forge handshake was commented out (I simply don't know what it is and how to test)

See https://wiki.vg/Minecraft_Forge_Handshake, though that page probably hasn't been maintained too much.  I don't know what the handshake looks like in 1.13 (it's probably different with 1.13 and the login plugin message system); the information on there probably only applies to 1.12 and earlier.   It's implemented using plugin channels, and it needs to go through for login to complete on a forge server with content mods.

> Submodule is, unfortunately, unavoidable, since it's the only NBT fork with LongArray type implementation and this PR was not merged into fNBT yet (thus, not pushed to nuget).

There is NBTExplorer which IIRC is based on the substrate library, but I'm not quite sure what its state is.  I do know NBTExplorer supports LongArray though.

vkorn:
@Pokechu22 thx, I'll check Forge out. 

Regarding Substrate, the problem is that in one of the new 1.14 packets, specifically ChunkData, height maps are coming "in the middle" of packet and we need something, that can read partial stream data as an NBT entity. I didn't such API in Substrate.

Pokechu22:
Ah, yeah, I forgot about that change and it does seem like it would be a bit complicated.  You could also just implement the needed code itself - I don't think MCC needs to mess with the heightmaps, so it should be sufficient to just identify the length and skip over the relevant number of longs.  Probably something like this (and this totally depends on how NBT is actually structured and whether the root tag has a name, which I haven't checked at all here):

```c#
void SkipHeightmaps(List<byte> data) {
    byte type = data.Pop();
    Assert(type == 10); // TAG_Compound
    short length = ReadShort(data);
    data.Pop(length);
    byte type2;
    while ((type2 = data.Pop()) != 0) { // TAG_End
         Assert(type2 == 12); // TAG_Long_Array 
         int length2 = ReadInteger(data);
         for (int i = 0; i < length2; i++) {
             ReadLong(data);
         }
    }
}
```

It's probably better to do a simple incomplete implementation of the relevant subset of NBT for this instead of bringing in a dependency - better to keep things lightweight here.

ORelio:
OK so, good news and bad news.

Good news is that 1.13 terrain handling has been implemented (#599), without adding too many changes in `Protocol18.cs`. Additionally, I noticed that various cases were added for snapshot-specific packet versions, and since we don't need that I squashed them into one case per major release so less code for each packet.

Well... bad news is that your pull request cannot be merged in its current form. Sorry 😕

TheSnoozer:
I see, not a problem :-)

I guess maintaining only major release helps in reducing all the weird special cases.

Do you think it would still make sense to split up the packet handling into separate classes?

Merci :-)

Pokechu22:
As a note, back when I was more actively contributing I tended to include snapshots simply because I was updating my code as snapshots released so it was natural to do it that way.  The [protocol history](https://wiki.vg/Protocol_History) page does help with doing it.  But when doing everything all at once, it's probably not worth implementing snapshots.

TheSnoozer:
Well yeah going through the protocol history manually is a pain. 

I'd recommend https://joodicator.github.io/mc-dev-data/ that was written by a maintainer of a minecraft python client that I used for quite some time....essentially the page is also build by python that downloads and parses the protocol history from the wiki page(s) automatically :-)

ORelio:
Hi,

When adding support for a new version, the easiest for me was to locate stage page for previous version of [Protocol}(https://wiki.vg/Protocol) and performing a diff with current version using the built-in wiki feature. Then, look at each packet implemented by MCC and add relevant changes.

`Protocol18.cs` is still very large, so maybe it can be separated, not necessarily all at once but step by step.

ORelio:
Just splitted `Protocol18.cs` into several files in order to reduce its size: https://github.com/ORelio/Minecraft-Console-Client/commit/d2cbc9f1c38bd90c3ae9e8ebfed8ec1bb28a1b79. This does not change much its internal behavior, but at least each file address a particular part of the protocol handling:
 - Protocol18: Handles Vanilla packets
 - Protocol18Forge: Handles Forge packets
 - DataTypes: Handles packet field encoding/decoding
 - SocketWrapper: Handles encryption and basic R/W operations

PR #728, opened 2019-05-14
Add 1.14.1 support

xdavidwu:
It seems like there is no big changes in protocol 480 (1.14.1).
Chat support tested and worked just fine.


ORelio:
Thanks! Merged 😉

Issue #674, opened 2019-03-17
System.OverflowException: VarInt too big


SweepingLoutus:
When my console starts and connects to the server and on join does the command /server skyblock  as it does so it gives out this error that I don't know how to fix. https://gyazo.com/dd3b72210e339255dcdb1ad8f0a1d720

TheSnoozer:
hi thanks for reporting your issue.
There are some issues that are related to that (e.g. https://github.com/ORelio/Minecraft-Console-Client/issues/379).

Sadly so far I never was able to reproduce this specific problem.
To what server are you connecting?

SweepingLoutus:
The server was skycade. I have two different accounts which arent related but have the exact same files and it was happening on one of them for sometime. But it just happened to both of them. One of them has fixed its self. Any questions please ask.

SweepingLoutus:
Here is a pastebin of all the error and starting of it with debug messages on https://pastebin.com/Hag68LkB

TheSnoozer:
Please change the password for your accounts immediately

SweepingLoutus:
Erm why

TheSnoozer:
Nevermind should be good, the log contained session information which looked like your credentials ;-)

BTW, so far i was not able to reproduce the issue. What OS are you using? Windows 10?

SweepingLoutus:
Yep, windows 10

SweepingLoutus:
ok so i have just tried to log an account on a vps and got this https://gyazo.com/61cedf397eaab09a19d2d254834868d2
https://pastebin.com/LHF3f9aT
This leads me to believe that it is a problem with my set up of the client.

TheSnoozer:
Can you perhaps share your MinecraftClient.ini? Make sure to remove all credentials before :-)

Insyder28:
Hi, I'm not really good with code and stuff, but your issue seems similar to mine, and I fixed it by logging in using the actual minecraft client and moving the character to a different position. I don't know why it worked, but it did, and maybe it will for you.

SweepingLoutus:
Sorry for not responding for a while my internet has been playing up over the last couple of days there is my MinecraftClient.ini file https://pastebin.com/hcSFh0VB

TNTUP:
In some areas of my server I'm getting VarInt errors https://i.qcfb.ca/20190621154838.png @ORelio

TerrainMovement is disabled and if I specify manual version, I don't know if it would solve it

[...]

PR #758, opened 2019-06-21
#674: fix one of the System.OverflowException: VarInt too big

TheSnoozer:
See https://github.com/ORelio/Minecraft-Console-Client/issues/674#issuecomment-504552416


ORelio:
The "Teleport ID" field is always present since "Terrain and Movements" is a feature of MCC. The official client cannot disable terrain and movements so it always process the field. https://wiki.vg/Protocol#Player_Position_And_Look_.28clientbound.29

PR #772, opened 2019-07-15
fixed match parameters not passing into script files. (issue: 770)

yunusemregul:
https://github.com/ORelio/Minecraft-Console-Client/issues/770

### DIFF
--- a/MinecraftClient/ChatBots/AutoRespond.cs
+++ b/MinecraftClient/ChatBots/AutoRespond.cs
@@ -99,9 +99,27 @@ namespace MinecraftClient.ChatBots
                     if (regex.IsMatch(message))
                     {
                         Match regexMatch = regex.Match(message);
-                        for (int i = regexMatch.Groups.Count - 1; i >= 1; i--)
-                            toSend = toSend.Replace("$" + i, regexMatch.Groups[i].Value);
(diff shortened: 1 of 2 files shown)


ORelio:
Thanks for contributing! It seems interesting to add $u in scripts, although this may not work with C# scripts. I'm also unsure if arguments are safely escaped while being passed to the Script command. I'll need to further review your pull request before deciding if it's OK to merge ;)

yunusemregul:
yea I guess it wont work with c# scripts. I hope I could help.

ORelio:
Ok so I took (way too much) time to think out this feature and came up with something that is 100% sure not to lead to any unexpected behaviour if a player plays with `"` inside a match and works with both regular and C# scripts.

Scripts and commands can now run with local variables, that are defined by the caller, available as usual through `%var%` for regular scripts, and `MCC.GetVar()` for C# scripts.

The following local variables are defined by AutoRespond:
 - `match_u` which corresponds to `$u` with player username
 - `match_1` which corresponds to `$1` regex capture
 - `match_2` which corresponds to `$2` regex capture
 - ... and so on for further regex captures

This way, you can effectively access match information from within a script. Thank you for sending a pull request, this helped me to come up with this feature even though I ended up writing it separately 👍

PR #773, opened 2019-07-19
Add 1.14.4 support

0Jumpero:
I really like MCC, but I don't know too much about it's code. Looking at the previous pull request from @NurdTurd which added 1.14.3 support I tried to add 1.14.4 support. I hope that's enough to make it work.
Thanks again for this awesome tool.


ORelio:
It should work for basic chat support as long as the protocol did not change. Terrain requires additional changes to at least mark the version as supported, see comments in source code.

Issue #729, opened 2019-05-15
Replicating /home with an MCC bot


EchoesNetwork:
Hello!

For a while now I've used MCC to replicate /home and /sethome commands in vanilla Minecraft. This makes use of 1.13.2's /forceload command, which can cause some server lag, so i'm looking to find a work around for that, but I've gotten stuck on figuring out how to get a player's coordinates with a chatbot and append those to an existing /teleport command in a new file. Here's what I'm trying to do, step by step:

The process starts in matches.ini, with the player executing the command `!sethome`:
```
[Match]
regex=^.*\G!sethome\z.*$
action=send /execute at $u run summon area_effect_cloud ~ ~1 ~ {CustomNameVisible:1b,Team:"homeCloud",Particle:"dust",Duration:36288000,Tags:["newHome"],CustomName:"{\"text\":\"$u\",\"color\":\"blue\"}"}
ownersonly=false

[Match]
regex=^.*\G!sethome\z.*$
action=send /kill @e[type=area_effect_cloud,tag=finishedHome,name=$u]
ownersonly=false

[Match]
regex=^.*\G!sethome\z.*$
action=send /execute at $u run tp @s $u
ownersonly=false

[Match]
regex=^.*\G!sethome\z.*$
action=move get
ownersonly=false

[Match]
regex=^.*\G!sethome\z.*$
action=send /tag $u add addHome
ownersonly=false

[Match]
regex=^.*\G!sethome\z.*$
action=send /execute at EchoesRadio run tp @s EchoesRadio
ownersonly=false
```
This does the following:

- An area effect cloud named after the user is summoned at the user's location to provide a visual marker of the home. It's tagged as newHome, which lets a series of command blocks recognize it and run /forceload at its location, thereby allowing my MCC bot to teleport the user to his/her home cloud from any location on the map.

- Any pre-existing home (tagged "finishedHome") is deleted.

- The bot teleports itself to the player

- The bot gets its position in the world and prints those coordinates into the console.

- The user is tagged with addHome, which allows some command blocks to inform him his new home has been set.

- The bot teleports back to another bot called EchoesRadio.

[...]

PR #805, opened 2019-09-20
#729: allow to use GetServerPort/GetServerHost/GetUsername/GetUserUUID in MCC scripts

TheSnoozer:
Related to https://github.com/ORelio/Minecraft-Console-Client/issues/729

This MR allow to use GetServerPort/GetServerHost/GetUsername/GetUserUUID in MCC scripts

Note I left out  `GetSessionID` since I feel this should not be exposed to a MCC script


ORelio:
Looks good! Thanks for contributing 😉

PR #837, opened 2019-12-22
Added 1.15.1 protocol

Mhowser:


AggroMC:
Thanks! Please Merge!

ORelio:
Hello @Mhowser, thanks for contributing!
Indeed there are no changes in 1.15.1 we can just mark it as supported.
There were some additional changes to perform to change the startup message and mark terrain as handled, see aba2ec79d1666e2d777811691d00dc843d97b562
Merry Christmas!

Mhowser:
Ah, I will remember that for next time, thanks for merging!

PR #855, opened 2020-02-07
Added the 1.15.2 protocol

rrvk:
This will add the 1.15.2 protocol to the client


KevinDenys:
Do you have a build with this, pr is taking ages..

rrvk:
@KevinDenys You can easily checkout my fork and then build it your self. Because i dont realy like to publish the build here because it is not yet the official build. And if you dont realy know how to check it out and build it your self you can email me at rvankammen@gmail.com

PR #856, opened 2020-02-09
Update CSharpRunner.cs

milutinke:
Added System.Net to allow communication with the web.


ORelio:
Seems good. For your information you can also use fully qualified class names e.g. `System.Net.WebClient` instead of `WebClient` for namespaces missing in CSharpRunner.

PR #857, opened 2020-02-09
Update CSharpRunner.cs

milutinke:
Added support for including libraries via comments
Example: //using System.Net;

Build passing: http://i.pics.rs/Q8ntM.png
Tested.


ORelio:
Why not! Would be nice to also document this feature in a sample script, if possible.

milutinke:
Done

ORelio:
Merged #859. Thanks!

PR #859, opened 2020-02-10
Update and rename README.txt to README.md

milutinke:
Added detailed explanation for C# scripts.
Added new better styling which is easier to read


ORelio:
Perfect, thanks. The README was previously in TXT for including with stable releases, but they are not provided anymore due to limited time, so converting to Markdown was a good idea.

Issue #808, opened 2019-09-21
Adding right click, left clicking interacting protocols


jakeau:
Was wondering if there was a place I could contact you about MC protocols. Cheers.

Mhowser:
Wiki.vg has all the documentation for mc protocols

ORelio:
Related: #502 #205 and #183
Documentation: https://wiki.vg/Protocol
If you have any question on how MCC handles packet or how a packet works, feel free to ask 😉

jakeau:
Cheers for the help.

I am attempting to implement this IOT click a button. I am testing it by attacking the sender.
https://wiki.vg/Protocol#Use_Entity

`        public void SendUseEntity(int target, int type = 0, float x = 0, float y = 0, float z = 0, int hand = 0)
        {
            handler.SendUseEntity(target, type, x, y, z, hand);
        }

        public bool SendUseEntity(int target, int type = 0, float x = 0, float y = 0, float z = 0, int hand = 0)
        {
            byte[] yawpitch = new byte[0];
            PacketOutgoingType packetType = PacketOutgoingType.UseEntity;


            try
            {
                SendPacket(packetType, dataTypes.ConcatBytes(
                    dataTypes.GetVarInt(target),
                    dataTypes.GetVarInt(type),
                    dataTypes.GetFloat(x),
                    dataTypes.GetFloat(y),
                    dataTypes.GetFloat(z),
                    dataTypes.GetVarInt(hand)));
                return true;
            }
            catch (SocketException) { return false; }
        }

`

I am being disconnected from the server. I am using varInt Target as the playerID of the player it wishes to attack. I believe this is wrong, what are the target value and the best way to obtain it. 

Cheers.

ORelio:
If not already the case, I'd suggest to run a vanilla server on your own computer since an anti-cheat plugin may interfer with your work (you'll make it right later, get it working first on vanilla even if it's not fully correct) and the console may show useful error messages.

Regarding the packet itself: make sure the ID of your target is right, visible, and you are looking at it. You may need to enable terrain and movements. You may also want to examine packet content sent by a vanilla client. To do this, I just added for you a [tool](https://github.com/ORelio/Minecraft-Console-Client/tree/master/DebugTools) previously developed for issues such as #195. It's a man-in-the-middle proxy that can intercept and display packets from a Vanilla client as long as encryption is disabled (offline mode servers only). Just need to add the relevant parsing code.

PR #891, opened 2020-03-21
Auto attack mobs around client player 1.15+

ReinforceZwei:
Auto attack hostile mobs within player range.
Made for AFK Exp farm that require player to hit the mobs.

Currently only implemented for MC 1.15+ (protocol version 573+)

related issues
#808


ReinforceZwei:
I found that my version of code is not the newest one.
Maybe I should pull the newest version first then make another PR
Also I want to implement for lower protocol version too (as I have free time now)
So I just close this PR first

Issue #502, opened 2018-06-18
Fishing? (protocol question regarding start and stop of item usage)


Simran-B:
Hi everyone,

there's a popular [AFK fish & XP farm design](https://youtu.be/Y31TEbKjSpg?t=1m40s) which exploits the game mechanics pretty clever to fully automate fishing.

In short: it uses some redstone, an iron door and a noteblock so that all you need is a mouse down event for the RMB (use item) while aiming at the right spot. Due to the clever design, the durability of the fishing rod is not unnecessarily diminished, but only reels in whatever you caught in the moment you catch it. The catch is collected by a hopper and stuffed into a chest.

This works very nicely in the normal game client with an enchanted fishing rod (especially Mending is required to make it last infinitely). I wrote a little script in AutoIt to send the required mouse down event, because taping down the physical mouse button seemed pretty stupid to me.

Now, I really wonder how this could be ported to MC console client. I went through the protocol description and found [**Use Entity**](http://wiki.vg/Protocol#Use_Entity) which seems to be relevant here. There are **no begin and end flags** however. For entering a mine cart etc., it is a single action, but bows and fishing rods kinda have multiple states.

For instance, the bow charges up in power in three intervals. Because there doesn't seem to be anything in the protocol for that, does it mean this is handled on the client side completely? Wouldn't it allow cheaters to always shoot with maximum power without properly charging the bow (effectively firing more rapidly while dealing maximal damage)?

Coming back to the fishing scenario, I'd like to know if the fish farm design would be completely irrelevant for a console client. Does it only trick the regular game client to automatically catch fish? Or is there a "bite" event coming from the server and another packet that tells you what you caught? I can't imagine that this happens on client side, as it would be really easy to manipulate to catch whatever you want, like a `/give` command.

So, if there's no way to directly translate the mouse down event to network packets, how else could this be implemented? Is there an easy way to intercept MC traffic to analyze fish farming with the game client? (Couldn't get Wireshark with Npcap and its Loopback Adapter to work for a local MC serve so far)

ReinforceZwei:
Hello,

Just want to announce that I have figured out how to fish with MCC (including use an item, detect if a fish is caught)
Here is a little prove screenshot:
![image](https://user-images.githubusercontent.com/39955851/77247017-0127b700-6c68-11ea-9946-7436c02b64d0.png)

But there are a lots of problems still need to be solved.
[...]

PR #892, opened 2020-03-22
Auto attack mobs around client player (updated version)

ReinforceZwei:
Auto attack hostile mobs within player range.
Made for AFK Exp farm that require player to hit the mobs.

Currently only implemented for MC 1.8-1.15

related issues
#808 
#502


ORelio:
Looks really cool! I'll review your PR and get back to you ASAP 😉

MinecraftClient/McTcpClient.cs
```diff
@@ -1018,5 +1081,298 @@ public void OnPluginChannelMessage(string channel, byte[] data)
...
+
+        /// <summary>
+        /// Called when an entity moved over 8 block.
+        /// </summary>
+        /// <param name="EntityID"></param>
+        /// <param name="X"></param>
+        /// <param name="Y"></param>
+        /// <param name="Z"></param>
+        /// <param name="onGround"></param>
+        public void OnEntityTeleport(int EntityID, Double X, Double Y, Double Z, bool onGround)
```
ORelio: (Same as other event, would map to OnEntityMove() )

MinecraftClient/Protocol/Handlers/PacketIncomingType.cs
```diff
@@ -30,6 +33,14 @@ enum PacketIncomingType
         OpenWindow,
         WindowItems,
         SetSlot,
-        UnknownPacket
```
ORelio: This one should stay at the bottom of the list 😉

MinecraftClient/Protocol/Handlers/Protocol16.cs
```diff
@@ -733,5 +733,23 @@ public static bool doPing(string host, int port, ref int protocolversion)
             }
             catch { return false; }
         }
+
+        // reinforce
+        public bool SendInteractEntityPacket(int EntityID, int type)
+        {
+            return true;
```
ORelio: Non-implemented methods should return false, also these addition could be moved to line 643 below SendClientSettings() and SendLocationUpdate() which are also not implemented.

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -242,7 +245,9 @@ internal bool HandlePacket(int packetID, List<byte> packetData)
                         float pitch = dataTypes.ReadNextFloat(packetData);
                         byte locMask = dataTypes.ReadNextByte(packetData);
 
-                        if (handler.GetTerrainEnabled())
+                        // player pos is needed for calculate entity distance
+                        // modified by reinforce
+                        if (handler.GetTerrainEnabled() || true)
```
ORelio: When terrain is disabled, entity handling should be disabled as well, thus not requiring this condition bypass. Or if entity handling could work without terrain handling (unsure if it would), add a setting for enabling entity handling and use `|| handler.GetEntitiesEnabled()` instead of `|| true` 😉

MinecraftClient/McTcpClient.cs
```diff
@@ -1018,5 +1081,298 @@ public void OnPluginChannelMessage(string channel, byte[] data)
...
+        {
+            if (!Settings.AutoAttackMobs) return;
+            // calculate server tps for adjusting attack cooldown
+            if (lastAge != 0)
+            {
+                DateTime currentTime = DateTime.Now;
+                Double tps = (WorldAge - lastAge) / (currentTime - lastTime).TotalSeconds;
+                lastAge = WorldAge;
+                lastTime = currentTime;
+                if (tps <= 20 || tps >= 0)
```
ORelio: Seems like this condition will always be true, did you mean `&&` instead of `||`?

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -535,6 +540,130 @@ internal bool HandlePacket(int packetID, List<byte> packetData)
                         SendPacket(PacketOutgoingType.ResourcePackStatus, dataTypes.ConcatBytes(responseHeader, dataTypes.GetVarInt(3))); //Accepted pack
                         SendPacket(PacketOutgoingType.ResourcePackStatus, dataTypes.ConcatBytes(responseHeader, dataTypes.GetVarInt(0))); //Successfully loaded
                         break;
+                    case PacketIncomingType.SpawnEntity:
+                        int EntityID = dataTypes.ReadNextVarInt(packetData);
```
ORelio: Handle entity-related packets only if terrain is enabled (or entity handling is enabled if such setting is added)

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -1023,5 +1152,50 @@ public static bool doPing(string host, int port, ref int protocolversion, ref Fo
...
+                return true;
+            }
+            catch (SocketException) { return false; }
+            catch (System.IO.IOException) { return false; }
+            catch (ObjectDisposedException) { return false; }
+        }
+        // TODO: Interact at block location
+        public bool SendInteractEntityPacket(int EntityID, int type, float X, float Y, float Z, int hand)
+        {
+            return true;
```
ORelio: return false when unimplemented (or remove this API if unneeded)

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -1023,5 +1152,50 @@ public static bool doPing(string host, int port, ref int protocolversion, ref Fo
...
+        }
+        public bool SendInteractEntityPacket(int EntityID, int type, float X, float Y, float Z)
+        {
+            return true;
```
ORelio: return false; (or remove this API if unneeded)

MinecraftClient/Protocol/Handlers/Protocol18PacketTypes.cs
```diff
@@ -42,6 +42,17 @@ public static PacketIncomingType GetPacketIncomingType(int packetID, int protoco
                     case 0x2D: return PacketIncomingType.OpenWindow;
                     case 0x2E: return PacketIncomingType.CloseWindow;
                     case 0x2F: return PacketIncomingType.SetSlot;
+
```
ORelio: Good job at implementing packet support all the way back to MC 1.7 👏

MinecraftClient/Settings.cs
```diff
@@ -230,6 +231,7 @@ public static void LoadSettings(string settingsfile)
                                                 case "privatemsgscmdname": PrivateMsgsCmdName = argValue.ToLower().Trim(); break;
                                                 case "botmessagedelay": botMessageDelay = TimeSpan.FromSeconds(str2int(argValue)); break;
                                                 case "debugmessages": DebugMessages = str2bool(argValue); break;
+                                                case "autoattackmobs": AutoAttackMobs = str2bool(argValue); break;
```
ORelio: If implemented as an internal ChatBot, this setting would be moved to its own INI file section (same with autofish 😉)

MinecraftClient/Settings.cs
```diff
@@ -558,6 +560,7 @@ public static void WriteDefaultSettings(string settingsfile)
                 + "debugmessages=false                # Please enable this before submitting bug reports. Thanks!\r\n"
                 + "scriptcache=true                   # Cache compiled scripts for faster load on low-end devices\r\n"
                 + "timestamps=false                   # Prepend timestamps to chat messages\r\n"
+                + "autoattackmobs=false               # Auto attack mobs around client player\r\n"
```
ORelio: (Same comment as above)

MinecraftClient/Protocol/Handlers/PacketOutgoingType.cs
```diff
@@ -19,6 +19,10 @@ enum PacketOutgoingType
         TabComplete,
         PlayerPosition,
         PlayerPositionAndLook,
-        TeleportConfirm
+        TeleportConfirm,
+        //modified by reinforce
```
ORelio: There's information about who added what in the git log and git blame sections, adding too many of these would make the code less readable in the long run. You can use `<remarks>added by reinforce</remarks>` instead in method descriptions, if needed.

MinecraftClient/McTcpClient.cs
ReinforceZwei: Actually the tps calculation is not very accurate (I don’t know why) and sometimes exceeds the range of 0-20

ReinforceZwei:
Thank you for your comments!!
Actually I’m not very familiarize with writing C# and using git and github. This is my first time writing and contributing such a huge project.
I will make changes base on those comments tomorrow.

BTW I am implementing inventory support for auto selecting tools for attacking and fishing and I need some help with the `Inventory` Class.
I need some explanation on each field, what data should be stored in those fields.
Or maybe I should write a new class for inventory?

ReinforceZwei:
Changes about auto attacking and auto fishing were done.
Please make another reviews, thanks.

ORelio:
Hi,

For a first time contribution, its already very good since you were able to understand the existing code and add changes to the appropriate places. Only had a few remarks for 900+ lines of changes 👍

Regarding the `Inventory` class, that is work in progress by @xPenguinx from commit 5895a4161fab5fe77d42726250f4988cf6f77de6 but was not completed into full inventory handling. As fas as I understand it, the fields are the following:

* `id`: inventory indentifier for use in the protocol (similar usage as entity id)
* `InventoryType`: type of inventory (chest, furnace, dispenser...)
* `title`: custom title shown in Minecraft GUI when opening the inventory
* `slots`: amount of item slots inside the inventory (not the items themselves)
* `public Dictionary<Item, int> items` should contain the items but not used yet in the code

You can use the existing class or replace it with your own implementation, i.e. re-implementing the existing packet handling from 5895a4161fab5fe77d42726250f4988cf6f77de6 with your own Inventory class and expand it to full inventory handling.

MinecraftClient/ChatBot.cs
```diff
@@ -137,6 +137,17 @@ internal void ProcessQueuedText()
...
 
+        public virtual void OnPlayerProperty(Dictionary<string, Double> prop) { }
+
+        public virtual void OnServerTpsUpdate(Double tps) { }
+
+        public virtual void OnEntityMove(Mapping.Entity entity) { }
+
+        public virtual void OnEntitySpawn(Mapping.Entity entity) { }
+
+        public virtual void OnEntityDespawn(int EntityID) { }
```
ORelio: Maybe pass the Entity object here instead of only passing the ID? Maybe the ChatBot will want to do something based on other properties of the Entity being deleted, e.g. watch for entity death without watching for spawn or move.

MinecraftClient/Commands/AutoAttack.cs
```diff
@@ -8,39 +8,35 @@ namespace MinecraftClient.Commands
...
                 {
-                    if (!handler.AutoAttack)
+                    if (handler.GetEntityHandlingEnabled())
                     {
-                        handler.AutoAttack = true;
-                        return "Auto attack turned on.";
+                        // TODO: check if the bot already loaded
+                        // I don't know how :C
+                        handler.BotLoad(new ChatBots.AutoAttack());
+                        return "Auto Attack now enabled.";
```
ORelio: That kind of interaction is indeed not implemented yet, existing bots have an associated INI File setting like this:
```ini
[AutoRelog]
enabled=true
```
Then a single instance of the bot is loaded on launch if the setting is enabled, and that's it. You could maybe remove both commands and stick back to a static setting like other bots for now, then add support for dynamically enabling/disabling the built-in bots at a later time.

A way for that could be a a `/chatbot` command which could show current status of bots and allow the user to load/unload builtin bots. Of course McTcpClient would need to expose additional methods for allowing that.

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -510,10 +509,10 @@ internal bool HandlePacket(int packetID, List<byte> packetData)
                         }
                         break;
                     case PacketIncomingType.WindowItems:
-                        if (handler.GetInventoryEnabled()||true)
+                        if (handler.GetInventoryEnabled()||true) // set true in .ini but GetInventoryEnabled seems like return false
```
ORelio: You need to adjust the check at line 70, otherwise inventory handling may auto-disable on server join:
```csharp
if (handler.GetInventoryEnabled() && protocolversion > MC114Version)
{
    ConsoleIO.WriteLineFormatted("§8Inventories are currently not handled for that MC version.");
    handler.SetInventoryEnabled(false);
}
```
This check sets the maximum version where inventory handling has been implemented and tested. On new Minecraft version, basic support is implemented first, and terrain may come at a later time. It might also be the case for inventories.

MinecraftClient/McTcpClient.cs
ORelio: Seems like you resolved it anyway in newer commits 😉

ReinforceZwei:
I am implementing the inventory support too and it is almost done. Just need to add backward support for 1.13.2 or older.
Do I commit those changes here or make another PR?

ORelio:
Thank you for your work, now AutoAttack and AutoFish is available for everyone :)
You can continue working on inventory support in a separate pull request 👍

ReinforceZwei:
AutoFish is not well implemented and the way to use it is a bit tricky.
I should include a instruction of usage in somewhere.
Maybe in the config file?

ORelio:
Yes, you can add instructions inside https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/config/README.md 🙂

Issue #675, opened 2019-03-17
Allow to enter username:password for login


ARandomSqueaker:
Make it so when you're entering the login info, you can put email:password instead of putting email then password. Its annoying copying the info then copying the ip.

If there is a suggestion area, how do i find it.

CarbonNeuron:
Just a recommendation, edit the username/password in the settings.ini file, and the IP address. That way you don't have to write them in Everytime

PR #901, opened 2020-03-29
Allowing to enter username:password for login

HRAshton:
Allows to enter username:password pair in console.

Issue #675

### DIFF
--- a/MinecraftClient/Program.cs
+++ b/MinecraftClient/Program.cs
@@ -122,18 +122,35 @@ namespace MinecraftClient
 
             //Asking the user to type in missing data such as Username and Password
 
-            if (Settings.Login == "")
-            {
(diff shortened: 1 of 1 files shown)


ORelio:
Hi,
Thanks for your contribution.
Could you detail a use case for this? Because when using the "Login" field, the password is not hidden.

Mhowser:
Many account lists use the format username:password, this PR will allow
easier automation for logging in accounts en masse.

ORelio:
In that case, you'll want to use it in command-line, right?
So instead of using:
```
MinecraftClient.exe login password serverip
```
You'll want:
```
MinecraftClient.exe login:password serverip
```
Instead of typing the credentials in the console like in this PR.
I can accommodate such changes if there is demand.

Issue #912, opened 2020-03-31
Console Client crashes/disconnect when using the /inventory click command


ReinforceZwei:
I am still looking at the code and try to fix it.
related:
#903 
```
>/inventory 0 click 23

未處理的例外狀況: System.Collections.Generic.KeyNotFoundException: 指定的索引鍵
不在字典中。
   於 System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   於 MinecraftClient.Protocol.Handlers.Protocol18Handler.SendClickWindow(Int32
windowId, Int32 slotId, Item item) 於 C:\Users\reinforce\source\repos\Minecraft- Console-Client-master\MinecraftClient\Protocol\Handlers\Protocol18.cs: 行 1312
   於 MinecraftClient.McTcpClient.ClickWindowSlot(Int32 windowId, Int32 slotId)
於 C:\Users\reinforce\source\repos\Minecraft-Console-Client-master\MinecraftClie nt\McTcpClient.cs: 行 1326
   於 MinecraftClient.Commands.Inventory.Run(McTcpClient handler, String command , Dictionary`2 localVars) 於 C:\Users\reinforce\source\repos\Minecraft-Console-C lient-master\MinecraftClient\Commands\Inventory.cs: 行 46
   於 MinecraftClient.McTcpClient.PerformInternalCommand(String command, String&  response_msg, Dictionary`2 localVars) 於 C:\Users\reinforce\source\repos\Minecr aft-Console-Client-master\MinecraftClient\McTcpClient.cs: 行 372
   於 MinecraftClient.McTcpClient.CommandPrompt() 於 C:\Users\reinforce\source\r
epos\Minecraft-Console-Client-master\MinecraftClient\McTcpClient.cs: 行 275
   於 System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   於 System.Threading.ExecutionContext.RunInternal(ExecutionContext executionCo ntext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   於 System.Threading.ExecutionContext.Run(ExecutionContext executionContext, C ontextCallback callback, Object state, Boolean preserveSyncCtx)
   於 System.Threading.ExecutionContext.Run(ExecutionContext executionContext, C ontextCallback callback, Object state)
   於 System.Threading.ThreadHelper.ThreadStart()
```

ReinforceZwei:
Another issues.
Server kicked me when I try to click on an exist item.
```
[...]

Issue #911, opened 2020-03-31
Inventory handling disconnects from server


lolitsmebarry:
Whenever i try to click an item in inventory with "/inventory 1 click 15" it disconnects me from the hub.

![bug](https://user-images.githubusercontent.com/22051319/77984658-7e12fa80-7312-11ea-9151-d8b5515891d5.PNG)

Edit:
im trying to join a server from the hub with compass inventory.

ORelio:
Please retry with the newest [development build](https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/artifacts)?

ORelio:
Should be fixed as per #913, feel free to reopen if it isnt.

Issue #903, opened 2020-03-30
Throws server at /inventory 0 click itemslot


Nekiplay:
enableentityhandling=true         # Toggle entities handling
terrainandmovements=true          # Uses more ram, cpu, bandwidth
inventoryhandling=true            # Toggle inventory handling (experimental, enable only for development)

![image](https://user-images.githubusercontent.com/35975332/77865215-acc39f00-7246-11ea-8832-1661cceaa166.png)

Mhowser:
Can't read Cryllic, what's wrong?

ORelio:
He is disconnected upon sending a Click Window packet using the new Inventory feature, which means the packet is either malformed or some plugin was not happy with it. Protocol 578 is 1.15.2 and I mainly tested this feature using that Minecraft version, but not with inventories going as far as slot 80 (is it even valid?)

ORelio:
Unfortunately we could not reproduce this issue.

PR #913, opened 2020-03-31
Fix Console Client crashes when using the /inventory click command

ReinforceZwei:
issues #912
related:
#911 
#903 

Little explanation:
Player inventory is sent in `WindowItems` packet but didn't add the `window_actions[windowID] = 0;` when handling that packet.


ReinforceZwei:
@Nekiplay If you found any problem, please create a new issues.

ORelio:
You are right, click actions are never initialized for inventory 0 since there is no "Open Window" for it.

Issue #914, opened 2020-03-31
 Inventory Handling Error


Nekiplay:
mcversion=1.15.2 # Use 'auto' or '1.X.X' values
brandinfo=mcc # Use 'mcc','vanilla', or 'none'
enableentityhandling=true # Toggle entities handling
terrainandmovements=true # Uses more ram, cpu, bandwidth
inventoryhandling=true # Toggle inventory handling (experimental, enable only for development)
![image](https://user-images.githubusercontent.com/35975332/77993359-70289e00-7341-11ea-9544-20b06a006982.png)
![image](https://user-images.githubusercontent.com/35975332/77993450-a2d29680-7341-11ea-904b-483555177b3f.png)

mcversion=1.8.9 # Use 'auto' or '1.X.X' values
![image](https://user-images.githubusercontent.com/35975332/77993326-5c7d3780-7341-11ea-9c5d-09ee6fa381f2.png)

ORelio:
What build are you using? The `/inventory` command is translated in Russian, which is not standard.

Nekiplay:
> What build are you using? The `/inventory` command is translated in Russian, which is not standard.

Last https://github.com/ORelio/Minecraft-Console-Client/commit/5d0172a1f9e778f6ef306f9fd55ab68a90728c7a

Nekiplay:
mcversion=1.15.2                     # Use 'auto' or '1.X.X' values
enableentityhandling=true         # Toggle entities handling
[AutoAttack]
# Entity Handling NEED to be enabled first
enabled=true
![image](https://user-images.githubusercontent.com/35975332/78017236-62850f80-7365-11ea-8ed9-89ab451e860c.png)

ORelio:
Seems like a calculation issue in [AutoAttack](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/ChatBots/AutoAttack.cs#L88) @ReinforceZwei

ReinforceZwei:
I couldn't reproduce this bug and I can't read the exception message because I don't understand Russian. I cannot debug :cry:

Nekiplay:
https://youtu.be/Y6ljwJYYeQc

ReinforceZwei:
Entity handling don't work on Minecraft version 1.12.2 or below.
Console will automatically disable entity handling when version unsupported.
Are you using the latest version of build?

Nekiplay:
[...]

PR #915, opened 2020-03-31
Fix AutoAttack internal tps not getting updated

ReinforceZwei:
What a embarrassing bug
Not sure if this fixes the issue #914


ReinforceZwei:
Also fixed part of the issue #914 Console crashes when Entity handling enabled on a unsupported MC version.

ORelio:
Thanks!

Issue #549, opened 2018-08-19
unable to auto respawn when dead


Blaze420z:
trying to auto respawn but it does not allow the commands to go thru when im dead
how do i get console client to auto respawn like a hacked client does please helpD;

Blaze420z:
id i type the command in the console client it works but i would like it to auto respawn but i cant get it to cause commands wont go thru when you die inless you type it in manually

ORelio:
Try using `AutoReponse` to detect the "you are dead" message, to run the `respawn` MCC command.

Yugnat2412:
Hello, I'm struggling with the same issue here. Since there is a chat message config, I think the bot doesn't read the death message. I match 'Botname' with respawn, and when I say it in game with another account it works, but when it says 'Botname died' in chat, nothing happens. It probably has to do with this config for the bot to know what is a chat message and what isn't

Yugnat2412:
I fixed it ! I didn't know actionother was for this kind of messages, it's actually not too hard. One thing is the bot sends a death message only when someone else kills it, I can't code a /killbot command it seems

PR #918, opened 2020-04-01
Add auto respawn after player dead

ReinforceZwei:
I was finding how to know if player got killed and I found it.
Hope this feature is useful.

related issue:
#549


ORelio:
Nice idea! Thanks. FYI `LogToConsole` in chatbots and `ConsoleIO.WriteLogLine()` in other parts should be preferred over `ConsoleIO.WriteLine` for message printed by MCC while logged in to a server. It automatically changes the color of the text, prepend [MCC] and the bot name, if applicable. I'll change it for this PR 😉

ReinforceZwei:
I see, I will use those logging methods in later development 😉

ORelio:
Done! Also changed to not use a thread because Protocol18 might not be fully thread safe (unsure).

PR #926, opened 2020-04-03
Added command to check player health and saturation

ReinforceZwei:
Hope it is useful.


MinecraftClient/Protocol/IMinecraftComHandler.cs
```diff
@@ -212,7 +212,7 @@ public interface IMinecraftComHandler
         /// <param name="item">Item (may be null for empty slot)</param>
         void OnSetSlot(byte inventoryID, short slotID, Item item);
 
-        void OnUpdateHealth(float health);
+        void OnUpdateHealth(float health, int Food);
```
ORelio, line 215: (same here: `int food`)

PR #930, opened 2020-04-07
Added AutoEat and Fixed command Health command name not changed

ReinforceZwei:
My bad :/

### DIFF
--- a/MinecraftClient/Inventory/Item.cs
+++ b/MinecraftClient/Inventory/Item.cs
@@ -70,24 +70,5 @@ namespace MinecraftClient.Inventory
                 return null;
             }
         }
-
-        /// <summary>
-        /// Check item is a food
-        /// </summary>
-        /// <returns>True if is a food</returns>
-        public bool IsFood()
-        {
-            // non-poison and stackable food
-            // remarks: auto eat may works with non-stackable food <- not tested
-            int[] foods = { 524, 765, 821, 823, 562, 763, 680, 629, 801, 585, 788, 630, 670, 674, 588, 587, 768, 673, 764, 777, 677, 679, 625, 800, 584, 787, 626, 678, 876, 627 };
-            if (foods.Contains((int)Type))
-            {
-                return true;
-            }
-            else
-            {
-                return false;
-            }
-        }
     }
 }
--- a/MinecraftClient/Inventory/Item.cs
+++ b/MinecraftClient/Inventory/Item.cs
@@ -70,5 +70,24 @@ namespace MinecraftClient.Inventory
                 return null;
             }
         }
+
+        /// <summary>
+        /// Check item is a food
+        /// </summary>
+        /// <returns>True if is a food</returns>
+        public bool IsFood()
+        {
+            // non-poison and stackable food
+            // remarks: auto eat may works with non-stackable food <- not tested
+            int[] foods = { 524, 765, 821, 823, 562, 763, 680, 629, 801, 585, 788, 630, 670, 674, 588, 587, 768, 673, 764, 777, 677, 679, 625, 800, 584, 787, 626, 678, 876, 627 };
+            if (foods.Contains((int)Type))
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
+        }
     }
 }
--- a/MinecraftClient/Inventory/ItemTypeExtensions.cs
+++ b/MinecraftClient/Inventory/ItemTypeExtensions.cs
@@ -7,12 +7,7 @@ namespace MinecraftClient.Inventory
 {
     public static class ItemTypeExtensions
     {
-        /**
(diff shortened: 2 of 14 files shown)


ReinforceZwei:
Also implemented AutoEat

MinecraftClient/McTcpClient.cs
```diff
@@ -1558,6 +1566,61 @@ public void OnUpdateHealth(float health, int food)
                     ConsoleIO.WriteLogLine("You are dead. Type /respawn to respawn.");
                 }
             }
+            if (Settings.AutoEat)
```
ORelio, line 1569: Same as with AutoAttack, `McTcpClient.cs` is quite large, so any automation should be moved to a ChatBot 🙂

MinecraftClient/McTcpClient.cs
```diff
@@ -1558,6 +1566,61 @@ public void OnUpdateHealth(float health, int food)
...
+        public void OnHeldItemChange(byte slot)
+        {
+            CurrentSlot = slot;
+        }
+
+        /// <summary>
+        /// Try to find food in the hotbar and eat it
+        /// </summary>
+        /// <returns>True if found</returns>
+        public bool FindFoodAndEat()
```
ORelio, line 1598: Since inventories are reachable from ChatBots, I  _think_ this also could be moved to the AutoEat bot 🤔

ReinforceZwei:
The Auto Attack bot needs to know if it is eating or not so that auto attack could pause while player is eating. If auto eat was moved into a chatbot, we might need to worry about how to make chatbot can communicate with each other without depending on the McTcpClient. 
As the case, should auto eat to be moved into a chatbot?

ORelio:
If I understand correctly, the way it is implemented, MCC does not know when the player has finished eating, and needs to wait for health update before putting back `Eating` to false. So this variable is specific to `AutoEating`? If so, even if not ideal, you can use a static flag on the `AutoEating` class:

````csharp
namespace MinecraftClient.ChatBots
{
    class AutoEating : ChatBot
    {
        private static bool Eating = false;
        public static bool IsEating { get { return Eating; } }

        //...
    }
}
````

Then other ChatBots can access the flag as `AutoEating.IsEating`.

File: MinecraftClient/Inventory/ItemTypeExtensions.cs [hunk 3]
```diff
@@ -0,0 +1,52 @@
...
+using System.Linq;
+using System.Text;
+
+namespace MinecraftClient.Inventory
+{
+    public static class ItemTypeExtensions
+    {
+        /**
+         * I see "this" in the MaterialExtensions class method, why need that?
+         * public static bool IsSolid(this Material m)
```
ORelio, line 12: That's an [extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods), which is an add-on for the Material enum. In that specific case where the subject of the method is not the current class, the first argument specifies what type of object is being manipulated, and explicitly reference it as a variable. so if you define `public static bool IsFood(this ItemType m)` in a different class than `ItemType`, you can still call it as a method from ItemType:
```C#
ItemType myItemType = ItemType.Apple;
if (myItemType.IsFood())
{
    //...
}
```

If you omit `this`, the method is not considered as an extension method, and you need to call it explicitly:

```C#
ItemType myItemType = ItemType.Apple;
if (ItemTypeExtensions.IsFood(myItemType))
{
    //...
}
```

It's just for convenience when you can't add a method to the class directly. (which is the case on enums)

MinecraftClient/Inventory/Item.cs
```diff
@@ -70,5 +70,24 @@ public string DisplayName
...
+        /// <summary>
+        /// Check item is a food
+        /// </summary>
+        /// <returns>True if is a food</returns>
+        public bool IsFood()
+        {
+            // non-poison and stackable food
+            // remarks: auto eat may works with non-stackable food <- not tested
+            int[] foods = { 524, 765, 821, 823, 562, 763, 680, 629, 801, 585, 788, 630, 670, 674, 588, 587, 768, 673, 764, 777, 677, 679, 625, 800, 584, 787, 626, 678, 876, 627 };
+            if (foods.Contains((int)Type))
```
ORelio, line 83: Just use `if (Type.IsFood())` after defining the extension method (see below). Then you can remove the `foods` array completely.

ReinforceZwei: Seems like I totally forgot about the array here :P

MinecraftClient/Inventory/ItemTypeExtensions.cs
ReinforceZwei: Thanks for your detail explanation!

ReinforceZwei:
Also fixed #933

ORelio:
Thanks for your contribution. Merged 😄

Issue #934, opened 2020-04-10
AutoAttack stop working with AutoEat enabled and no food on hotbar


ReinforceZwei:
The reason is `Eating` was set to `true` even no food on the hotbar and never set back to false, so that it didn't pass the `if (!AutoEat.Eating)` condition.
I made some fix attempt but still not working
```cs
public override void OnHealthUpdate(float health, int food)
{
    if (food <= HungerThreshold || (food < 20 && health < 20))
    {
        if (FindFoodAndEat())
        {
            Eating = true;
        }
        else
        {
            Eating = false;
        }
    }
    // keep eating until full
    if (food < 20 && Eating)
    {
        if (!FindFoodAndEat())
        {
            // ran out of food :(
            Eating = false;
        }
    }
    if (food >= 20 && Eating)
    {
        Eating = false;
        ChangeSlot(LastSlot);
    }
}
```
I have no idea how to fix it :cry:

ORelio:
Seems like in `OnHealthUpdate()`, `FindFoodAndEat()` is called twice. Is it necessary? If `FindFoodAndEat()` returns `true` in the first `if`, then you are already eating, no need to execute the "keep eating" part, right?

ReinforceZwei:
Actually it is necessary. After eating the first food, hunger may exceed the threshold but not at the full hunger. The second call handled that and make player keep eating more food until full.

ReinforceZwei:
I made some improvement for AutoEat in [51f5497](https://github.com/ReinforceZwei/Minecraft-Console-Client/commit/51f54979daf9c71ff073034e16de2684fd9e92f1) but still cannot solve the issue.
I think the cause of the problem is inventory update delay. If the player ate the last food in the hotbar and the hunger bar is not filled up, the next `FindFoodAndEat()` call will return true even no food in the hobar. But if I "make" a HealthUpdate by punching the player later, the `FindFoodAndEat()` work properly and AutoAttack resumed.
So should I add a delay before calling the `FindFoodAndEat()` method?

ReinforceZwei:
It works after adding a bit delay before calling the `FindFoodAndEat()` method.

PR #937, opened 2020-04-11
Fixed AutoAttack stop working with AutoEat enabled and no food on hotbar

ReinforceZwei:
#934

### DIFF
--- a/MinecraftClient/ChatBots/AutoEat.cs
+++ b/MinecraftClient/ChatBots/AutoEat.cs
@@ -3,8 +3,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace MinecraftClient.ChatBots
 {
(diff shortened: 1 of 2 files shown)


MinecraftClient/ChatBots/AutoEat.cs
```diff
@@ -20,15 +21,22 @@ public AutoEat(int Threshold)
...
-                FindFoodAndEat();
+                Eating = FindFoodAndEat();
+                if (!Eating)
+                    ChangeSlot(LastSlot);
             }
             // keep eating until full
             if (food < 20 && Eating)
             {
-                FindFoodAndEat();
+                Task.Factory.StartNew(delegate
```
ORelio, line 33: Unfortunately, MCC is not fully thread safe and might cause an exception when using the ChatBot API from a different thread. As a workaround, you can set a boolean like `EatOnNextUpdate = true` (or use a counter decremented each tick), then do the logic in `OnUpdate()`. Apologies for the low code quality of MCC's core 😅

ReinforceZwei: Don’t mind. I have came up with a better solution that do not require Threading. I will modify and test the code tomorrow.

ReinforceZwei:
My idea is decrease player's inventory food count by one after found a food so that we don't need to wait for the server to *update* our inventory, but `ChatBot.GetPlayerInventory()` return a **copy** of player inventory, hence I cannot modify the player inventory inside a ChatBot. Should I add a new API or make that method return the reference of player inventory?

ORelio:
I think it could pose other issues if the player inventory no longer reflect what's really in the player inventory server-side. Another bot could try to use an "empty" slot that actually contains food. However, the AutoEat bot could _store_ its copy of the inventory, or store slot ID and item count, then modify its own copy.

ReinforceZwei:
It might become quite complicated if AutoEat store its own copy of inventory as it require some work on synchronize the inventory with both McTcpClient and server. I think I will just use the `OnUpdate()` as the solution.

Issue #951, opened 2020-04-16
Failed to check session


Atarka2:
recently i have been getting failed to check session messages even though i am using the latest version, on some occasions immediatly when i launch it i get a failed to check session error message, and on some occasions when it attempts to reconnect i get the message. this is extremely annoying since it renders the client useless

IchUseKomplexe:
Same for me

slutfarm:
Same.

CarbonNeuron:
Same!

CarbonNeuron:
`Performing request to sessionserver.mojang.com
> POST /session/minecraft/join HTTP/1.1
> Host: sessionserver.mojang.com
> User-Agent: MCC/1.12.2 DEV
> Content-Type: application/json
> Content-Length: 433
> Connection: close
> 
> {"accessToken":"REDACTED","selectedProfile":"REDACTED","serverId":"REDACTED"}

< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
< Connection: close
< Date: Sat, 18 Apr 2020 00:56:26 GMT
< x-minecraft-request-id: d13103463d9c9511
< X-Cache: Miss from cloudfront
< Via: 1.1 010086e284fd9714db3b3e3d1d295b60.cloudfront.net (CloudFront)
< X-Amz-Cf-Pop: ORD51-C1
< X-Amz-Cf-Id: 5y7CUO40WJbRR47GIwe9_x-KQQm4P7THoskbpYOQPPwfTloyAVqCjA==
< 
< 0
< 
< 
`
This is with debug messages on. Is the connection supposed to be closed?

ORelio:
Thanks for raising the issue, should be fixed in latest dev build including @CarbonNeuron's fix 👍

PR #952, opened 2020-04-18
Fix "failed to check session cache"

CarbonNeuron:
This fixes this issue #951 randomly the Mojang session server will send back a status code of 204 with no content instead of a status code of 200. Status Code 204 is just saying that it doesnt need to change anything from last time from looking at [the documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204). It still is a success code.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# Default ignored files
+/.idea.MinecraftClient/.idea/workspace.xml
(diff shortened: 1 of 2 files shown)


Legend-of-iPhoenix:
In fact, everything in [2xx indicates success](https://tools.ietf.org/html/rfc7231#section-6.3).

CarbonNeuron:
I guess so

ORelio:
Thanks for the fix! Indeed, any 2XX code should be OK in session check 😉

Issue #964, opened 2020-04-23
AutoAttack not attacking mobs


OverHash:
Running off the latest development build.

I have enabled AutoAttack in config, and even modified `AutoAttack.cs` to have it's initialize function as this:
```cs
public override void Initialize()
{
	LogToConsole("Initializing AutoAttack");
	if (!GetEntityHandlingEnabled())
	{
		LogToConsole("Entity Handling is not enabled in the config file!");
		LogToConsole("This bot will be unloaded.");
		UnloadBot();
	}

	LogToConsole("AutoAttack loaded");
}
```

These both log to the console, 
![image](https://user-images.githubusercontent.com/46231745/80060042-b6100500-8581-11ea-9987-b0127b138245.png)

On the server that I am trying to use this on, entities can stack. This means that if there are 3 skeletons nearby, they will "merge" into one big skeleton. 
![image](https://user-images.githubusercontent.com/46231745/80060150-f5d6ec80-8581-11ea-9a50-4b150b9ae97b.png) _An example of a stack of 20 skeletons_

However, whenever my client is nearby the skeletons, it does absolutely nothing. It acknowledges something is nearby, and even returns `true` saying it interacted with the entity, but it never throws a punch.

I modified `AutoAttack.cs`'s `Update` method to be this:
```cs
public override void Update()
{
    if (!AutoEat.Eating)
    {
        if (attackCooldownCounter == 0)
        {
            attackCooldownCounter = attackCooldown;
            if (entitiesToAttack.Count > 0)
            {
                foreach (KeyValuePair<int, Entity> a in entitiesToAttack)
                {
                    bool interactResult = InteractEntity(a.Key, 1);
					LogToConsole("interaction with " + a.Key + " after hitting was " + interactResult);
                }
            }
        }
        else
        {
            attackCooldownCounter--;
        }
    }
}
```

This gives me output something along the lines of this:
```txt
[...]

PR #967, opened 2020-04-23
Clean AutoAttack

OverHash:
A lot of this code is messy and has unused variables. Additionally, there are memory leaks as dead entities are not removed from memory.

Fixes some issues related to #964

### DIFF
--- a/MinecraftClient/ChatBots/AutoAttack.cs
+++ b/MinecraftClient/ChatBots/AutoAttack.cs
@@ -18,6 +18,7 @@ namespace MinecraftClient.ChatBots
         private Double attackCooldownSecond;
         private int attackRange = 4;
         private Double serverTPS;
+        private float health = 100;
 
         public override void Initialize()
         {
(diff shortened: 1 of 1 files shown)


OverHash:
This PR still needs work, it needs to be able to listen for
- player movement changes (to see what entities went out of its range as it moved)
- player death (to stop attacking entities)

ReinforceZwei:
Player movement was handled properly, also and the calculation of distance between player and entities.
For player death, you can override `ChatBot.OnHealthUpdate()` method to know player's health. If the health is 0, that mean the player was dead.

OverHash:
Player movement isn't handled properly. If the entities all stay in the same spot, and the player moves back 5 units, it will still continue to attack.

What method can I override to add a callback for player movement?

ORelio:
Hi,

Just pushed a few formatting fixes.
The current state of your PR seems good to me, please let me know when you are ready for merging.

Regarding movement: You can get coordinates using the bot API but there is not event currently when the player is forcefully moved by the server. `GetCurrentLocation()` is supposed to be up to date at the time it is called, i.e. if a new location is received from the server, it will be reflected in the ChatBot API.

OverHash:
Thanks! I'm new to C# so not quite sure what the formatting is.

About to push a commit which should be the last of my changes. Please let me know your thoughts on this.

ORelio:
Seems good! Merged your pull request.

PR #980, opened 2020-04-30
Add UUID to entities

CarbonNeuron:
Adds UUID to entities and fixes some of the functions that spawn them. Removes some redundant object creation.


ORelio:
Looks good, merged. Thanks for your contribution 👍

PR #982, opened 2020-05-01
Automatically look at nearest player

CarbonNeuron:
This adds a bot which will look at the nearest player, It also implements a ChatBot method to Look at a Location. Requires Terrain handling, and entity handling to be enabled.


CarbonNeuron:
Sorry about the 2 commits that were already merged. I don't really use git that much, But It looks like there hasn't been any changes so it shouldn't be a problem.

ORelio:
Thanks for your pull request! I added your changes in commit 9dae40153c41ce74f73d99e1af26fd321210b1ea to avoid any conflict with already merged commits. 😉

CarbonNeuron:
Awesome!

Issue #938, opened 2020-04-12
Add Logging for Internal Commands and Stuff


Chtholly:
Please implement Logging to file for all used Commands and output like eg. Inventory
and i need a .cs for logging playerjoins/lefts to file and send me a message but theres no notification in chat

Nekiplay:
![image](https://user-images.githubusercontent.com/35975332/79060729-165d9780-7ca2-11ea-862a-0cb0c5f39da4.png)
i am use my code

ORelio:
Could be interesting indeed. To my understanding, this would be similar to the existing ChatLog bot, but for logging commands entered in MCC and output of internal commands. Adding this to the TODO list.

CarbonNeuron:
Think this can be closed now due to #984 getting merged.

PR #984, opened 2020-05-01
Added Entity Action handling

CarbonNeuron:
Added Entity Action handling.

Sincerely sorry about the previous pull request. Hopefully this one is better.


CarbonNeuron:
Oh, I see part of why I was having problems. The AutoLook chatbot is not in the main Project anymore, But In the Config folder. For obvious reasons, its not needed in the main project. How do you make that work with Settings.cs though? Can it load the bot from the file still?

CarbonNeuron:
This PR now also includes internal command logging. For #938 Because I couldn't figure out how to push to a branch without it adding to this commit.

MinecraftClient/ChatBot.cs
```diff
@@ -662,6 +664,45 @@ public bool GetInventoryEnabled()
...
+        public Dictionary<int, Container> GetInventories()
+        {
+            return Handler.GetInventories();
+        }
+        /// <summary>
+        /// Use item in hand
+        /// </summary>
+        protected void UseItemInHand()
+        {
+            Handler.UseItemOnHand();
```
ORelio, line 676: For consistency, could you also rename `UseItemOnHand` to `UseItemInHand`? Thanks 😉

MinecraftClient/ChatBot.cs
```diff
@@ -662,6 +664,45 @@ public bool GetInventoryEnabled()
...
+        /// start Sneaking
+        /// </summary>
+        protected bool Sneak()
+        {
+            return SendAction(Protocol.ActionType.StartSneaking);
+        }
+        /// <summary>
+        /// Sends UnSneak
+        /// </summary>
+        protected bool UnSneak()
```
ORelio, line 688: Since Sneaking is either ON or OFF, maybe define the following function instead of 2 functions?
```C#
protected bool Sneak(bool on)
```

MinecraftClient/ChatBot.cs
```diff
@@ -662,6 +664,45 @@ public bool GetInventoryEnabled()
...
+        {
+            return SendAction(Protocol.ActionType.StopSneaking);
+        }
+        /// <summary>
+        /// Send Entity Action
+        /// </summary>
+        private bool SendAction(Protocol.ActionType action)
```
ORelio, line 695: I think `SendEntityAction` would be clearer. Since this is the Bot API, it is almost impossible to change function names after creating the methods without breaking user-created bots, so we should choose them wisely. 😄

MinecraftClient/ChatBots/ChatLog.cs
```diff
@@ -12,11 +12,12 @@ namespace MinecraftClient.ChatBots
 
     public class ChatLog : ChatBot
     {
-        public enum MessageFilter { AllText, AllMessages, OnlyChat, OnlyWhispers };
+        public enum MessageFilter { AllText, AllMessages, OnlyChat, OnlyWhispers, OnlyInternal };
```
ORelio, line 15: How about `OnlyInternalCommands` for this field?

MinecraftClient/ChatBots/ChatLog.cs
```diff
@@ -92,6 +100,14 @@ public override void GetText(string text)
             }
         }
 
+        public override void OnInternalCommand(string commandName,string commandParams, string Result)
+        {
+            if (saveInternal)
+            {
+                save($"Internal {commandName}({commandParams}): {Result}");
```
ORelio, line 107: String Interpolation is only available starting with C# 6. MCC is targeting .NET 4 with C# 4.
You can easily circumvent this by using String formatting instead:
```C#
save(String.Format("Internal {0}({1}): {2}", commandName, commandParams, result));
```

MinecraftClient/ChatBots/ChatLog.cs
```diff
@@ -92,6 +100,14 @@ public override void GetText(string text)
             }
         }
 
+        public override void OnInternalCommand(string commandName,string commandParams, string Result)
```
ORelio, line 103: C# [naming conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions) involve using `PascalCase` for method names and class attributes, and `camelCase` for local variables and arguments, so here this argument should be `string result`. You'll encounter some `incorrect_names` inside MCC's code, this is however not compliant and should be improved whenever possible. 😉

MinecraftClient/Commands/Sneak.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MinecraftClient.Commands
+{
+    public class Sneak : Command
+    {
+        private bool sneaking = false;
+        public override string CMDName { get { return "TSneak"; } }
```
ORelio, line 11: Seems like there is an extra T: `return "Sneak"`

MinecraftClient/Protocol/Action.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MinecraftClient.Protocol
+{
+    public enum ActionType
```
ORelio, line 8: Both the file and enum should have the same name.
Maybe use `EntityActionType.cs` and `EntityActionType` for enum?

MinecraftClient/Protocol/Handlers/Protocol16.cs
```diff
@@ -239,6 +239,10 @@ private string readNextString()
             }
             else return "";
         }
+        public bool SendEntityAction(int PlayerEntityID, int ActionID)
+        {
+            return true;
```
ORelio, line 244: `return false;` for unimplemented actions.

MinecraftClient/Settings.cs
```diff
@@ -161,6 +161,9 @@ public static byte MCSettings_Skin_All
         //Auto Eating
         public static bool AutoEat_Enabled = false;
         public static int AutoEat_hungerThreshold = 6;
+        
+        //Auto Looking
+        public static bool AutoLook_Enabled = false;
```
ORelio, line 166: Seems like a leftover from your previous Pull Request. Can be removed from Settings.cs.

MinecraftClient/Commands/Sneak.cs
CarbonNeuron: This was because it does what a mod does called ToggleSneak. I shortened it to TSneak but I will fix it.

PR #995, opened 2020-05-12
Strip Minecraft Color Codes (§) in BasicIO mode

5tanly:
Adds a configuration setting in MinecraftClient.ini to strip all color codes and formatting from output when in BasicIO mode. Makes chat logs much more readable.

I use Docker to run Mono, so I need to use BasicIO mode to be able to run, stripping color and formatting codes makes it easier to interact with players.

Screenshot_2020-05-12 Portainer
Screenshot_2020-05-12 Portainer(1)

### DIFF
--- a/MinecraftClient/ConsoleIO.cs
+++ b/MinecraftClient/ConsoleIO.cs
@@ -345,11 +345,7 @@ namespace MinecraftClient
                 {
                     if (BasicIO_NoColor)
                     {
-                        string colorcodes = "0123456789abcdefklmnor";
-                        foreach (char c in colorcodes)
-                        {
-                          str = str.Replace("§"+c, string.Empty);
-                        }
+                        str = ChatBot.GetVerbatim(str);
                     }
                     Console.WriteLine(str);
                     return;
--- a/MinecraftClient/ConsoleIO.cs
+++ b/MinecraftClient/ConsoleIO.cs
@@ -56,6 +56,13 @@ namespace MinecraftClient
         public static bool BasicIO = false;
 
         /// <summary>
+        /// Determines whether to use interactive IO or basic IO.
+        /// Set to true to disable interactive command prompt and use the default Console.Read|Write() methods.
+        /// Color codes are NOT printed when BasicIO_NoColor is enabled.
+        /// </summary>
+        public static bool BasicIO_NoColor = false;
+
+        /// <summary>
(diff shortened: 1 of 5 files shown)


MinecraftClient/ConsoleIO.cs
```diff
@@ -336,6 +336,14 @@ public static void WriteLineFormatted(string str, bool acceptnewlines = true, bo
                 }
                 if (BasicIO)
                 {
+                    if (Settings.ConsoleIOColor_Enabled == false)
+                    {
+                        string colorcodes = "0123456789abcdefklmnor";
+                        foreach (char c in colorcodes)
+                        {
+                          str = str.Replace("§"+c, string.Empty);
+                        }
```
ORelio, line 345: Instead of rewriting this, you can use the existing method `GetVerbatim` from the ChatBot class [here](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/ChatBot.cs#L243) after switching it to Public visibility.

MinecraftClient/Settings.cs
```diff
@@ -601,6 +603,7 @@ public static void WriteDefaultSettings(string settingsfile)
                 + "scriptcache=true                   # Cache compiled scripts for faster load on low-end devices\r\n"
                 + "timestamps=false                   # Prepend timestamps to chat messages\r\n"
                 + "autorespawn=false                  # Toggle auto respawn if client player was dead (make sure your spawn point is safe)\r\n"
+                + "consoleiocolors=true               # Strips the color codes from output when in BasicIO mode\r\n"
```
ORelio, line 606: Using a setting may pose an issue with programs and scripts calling MCC and expecting color codes to be present. I'd suggest to use a command-line argument (e.g. `BasicIO-NoColors`) instead and set the setting directly inside the ConsoleIO class 🙂

5tanly:
I believe this is all the changes that would be required, BasicIO-NoColor is now a command line option, that will enable BasicIO mode aswell, but remove color code using the GetVerbatim method.

Also I added dotfiles to gitignore. It was annoying needing to deselect dotfiles every time I commit 😄

ORelio:
I don't mind ignoring hidden files, seems OK for me. Once you push new commits to your repository, I get notified and can review them again. Of course, you can also manually request a review to explicitly tell me that your PR is now good to be reviewed again 😉

PR #1009, opened 2020-05-22
Fix #1006

### LINKED ISSUE #1006: Fishing hook disappearing while fishing
Hey, It's me again. Now the question is:
The server I'm playing sometimes removes "trash" altogether with the hook entity and AutoFishing does not notice this, continuing to wait before a fish gets caught. Time before this happens depends on the amount of players and entities. Right now I'm trying just to blindly resume fishing with a `/useitem` task running every couple minutes, which is more of problem-making rather problem-solving because it most of the time stops the process while its still running. Any idea on how to keep AutoFishing running?

ReinforceZwei:
Also make AutoFishing thread-safe.


ORelio:
Thanks! Merged.

Issue #1010, opened 2020-05-22
Implements right click and middle click


Stymix:
Hi, someone can help me by implementing middle click (knurl) and right click when I pressed the E button ?

Because we can only left click for the moment.

Thanks !

ReinforceZwei:
After checking the [protocol](https://wiki.vg/Protocol#Click_Window) document, there is no middle click action for survival mode. Not sure why you need the middle click.

For right click, it is easy to implement. I will do it when free.

Stymix:
I need middle click for use it in a plugin who opened a gui

Stymix:
And thanks for the right click :smile:

ReinforceZwei:
> I need middle click for use it in a plugin who opened a gui

I don't know how vanilla game will send the middle clicking packet when in survival mode so I m not sure how to implement it correctly.

BTW, the implementation of right clicking have been done.

Stymix:
In faction when u want to sell your items in a gui shop, u can use the middle click

Stymix:
and thanks for the right clicking

ReinforceZwei:
Okay, I will try to implement it.
You may need to test it for me when I finished.

ReinforceZwei:
Middle click was done. Please see my [repo](https://github.com/ReinforceZwei/Minecraft-Console-Client/tree/click).
It would be nice if you could test it for me before merging into master branch.

ORelio:
Should be implemented now.

PR #1012, opened 2020-05-23
Add right click and middle click function for inventory

ReinforceZwei:
issues #1010 #988

### DIFF
--- a/MinecraftClient/Commands/Inventory.cs
+++ b/MinecraftClient/Commands/Inventory.cs
@@ -9,7 +9,7 @@ namespace MinecraftClient.Commands
     class Inventory : Command
     {
         public override string CMDName { get { return "inventory"; } }
-        public override string CMDDesc { get { return "inventory <<id>|player|container> <list|close|click <slot> <left|right|middle>>: Interact with inventories"; } }
+        public override string CMDDesc { get { return "inventory <<id>|player|container> <list|close|drop <slot> <all>|click <slot> <left|right|middle>>: Interact with inventories"; } }
 
         public override string Run(McTcpClient handler, string command, Dictionary<string, object> localVars)
         {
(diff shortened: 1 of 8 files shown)


MinecraftClient/Protocol/IMinecraftCom.cs
```diff
@@ -144,9 +144,10 @@ public interface IMinecraftCom : IDisposable, IAutoComplete
         /// </summary>
         /// <param name="windowId">Id of the window being clicked</param>
         /// <param name="slotId">Id of the clicked slot</param>
+        /// <param name="buttom">0 for left click, 1 for right click, 2 for middle click</param>
         /// <param name="item">Item in the clicked slot</param>
         /// <returns>True if packet was successfully sent</returns>
-        bool SendClickWindow(int windowId, int slotId, Item item);
+        bool SendClickWindow(int windowId, int slotId, byte buttom, Item item);
```
ORelio, line 150: I think maybe other actions would be impemented at a later time so how about renaming to `SendWindowAction` and using either an Enum _or_ raw protocol values instead of auto-changing 2 to 3? The parameter name would be `WindowAction action` or `byte action` 😉

MinecraftClient/McTcpClient.cs
```diff
@@ -1511,13 +1511,13 @@ public bool UseItemOnHand()
         /// Click a slot in the specified window
         /// </summary>
         /// <returns>TRUE if the slot was successfully clicked</returns>
-        public bool ClickWindowSlot(int windowId, int slotId)
+        public bool ClickWindowSlot(int windowId, int slotId, byte buttom)
```
ORelio, line 1514: See remark in IMinecraftCom.cs

MinecraftClient/Commands/Inventory.cs
```diff
@@ -9,7 +9,7 @@ namespace MinecraftClient.Commands
     class Inventory : Command
     {
         public override string CMDName { get { return "inventory"; } }
-        public override string CMDDesc { get { return "inventory <<id>|player|container> <list|close|click <slot>>: Interact with inventories"; } }
+        public override string CMDDesc { get { return "inventory <<id>|player|container> <list|close|click <slot> <L|R|M>>: Interact with inventories"; } }
```
ORelio, line 12: Although it requires more characters to type,  `<L|R|M>` may be unclear especially in scripts, and given that this command is already pretty complex. I think it should be documented as `<left|right|middle>`. You can optionally add support for `l` `r` `m` by only checking the first character, e.g. document `<left|right|middle>` but check if `b.ToLower()[0] == "r"` instead of `b.ToLower() == "right"`.

ReinforceZwei:
Sure, I will do it tomorrow.

ReinforceZwei:
Also implemented dropping item from inventory #1016

LidForr:
ReinforceZwei Hi, How do you use a right click ?

ReinforceZwei:
@LidForr `inventory <<id>|player|container> <list|close|drop <slot> <all>|click <slot> <left|right|middle>>`

The right click function is for clicking item in the inventory. For example, `/inventory player click 36 right` will right click on the item at slot 36.

LidForr:
Thank you, and thank you for responding so quickly.

LidForr:
Thank you, and thank you for responding so quickly.
so can he right-click in the direction he's looking ?

ReinforceZwei:
@LidForr In short, no. It is not simple as its look to implement left/right clicking in the direction player is looking at.

LidForr:
Okey,Thanks

PR #1015, opened 2020-05-23
Faster terrain packet handling in memory

eulertour0:
First apologize for poor English and github manner.
I'm newbie at English and Github.

I had problem with using terrain and movements.
cpu usage going to 100%(in one thread) and timeout from server.
Some of Issues with 100% cpu usage are same reason.

The cause of problem is in packet processing.
List.RemoveAt(0); It will take time in proportion to n(=List.Count)
because RemoveAt(0) operation will reallocate size of n area and copying.
And packet processing will take time in proportion to m^2(m=packet size)

You can avoid this problem with using Queue instead List.
It can run packet processing in proportion to m(=packet size) time.

In 1.15 chunk packet size is around 15000.
so m^2 ≈ 225000000 it is so big.

### DIFF
--- a/MinecraftClient/Protocol/Handlers/DataTypes.cs
+++ b/MinecraftClient/Protocol/Handlers/DataTypes.cs
@@ -34,10 +34,11 @@ namespace MinecraftClient.Protocol.Handlers
         /// <param name="offset">Amount of bytes to read</param>
         /// <param name="cache">Cache of bytes to read from</param>
         /// <returns>The data read from the cache as an array</returns>
-        public byte[] ReadData(int offset, List<byte> cache)
+        public byte[] ReadData(int offset, Queue<byte> cache)
         {
             byte[] result = cache.Take(offset).ToArray();
-            cache.RemoveRange(0, offset);
+            for (int i = 0; i < offset; i++)
+                cache.Dequeue();
             return result;
         }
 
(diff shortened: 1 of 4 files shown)


ORelio:
Hi,

Welcome to Github and thanks for your contribution!

Indeed, according to [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.removeat?view=netcore-3.1) for List:
> This method is an O(n) operation, where n is (Count - index).

O(n) is not packet size ^ 2 but just packet size (i.e. process all elements once).

And [this page](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1.dequeue?view=netcore-3.1) for Queue:
> This method is an O(1) operation.

Constant time regardless of packet size. Seems way betted indeed.

I thought List was a little more optimized and did not reallocate every time an element is removed, like a StringBuilder for instance. So replacing it with a queue makes sense, especially since it require much less changes in code than I'd have expected.

Before merging, I left a comment about a slight change that would further improve performance. All you need to do is creating a commit and pushing it to your fork repository. It'll automatically add to the pull request. Thanks! 🙂

ORelio:
Just added the changes for you. Merged! Thanks.
For any further pull requests you'll need to [sync your fork](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork). 😉

PR #1044, opened 2020-06-06
ConsoleIcon support Windows 10 on new Console & OnLatencyUpdate Event

Nekiplay:
### Done

- [x] ConsoleIcon support Windows 10
- [x] OnLatencyUpdate Event

### Example
**OnLatencyUpdate** 
image

### DIFF
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -212,6 +212,14 @@ namespace MinecraftClient
         /// <param name="uuid">Player UUID</param>
         /// <param name="gamemode">New Game Mode (0: Survival, 1: Creative, 2: Adventure, 3: Spectator).</param>
         public virtual void OnGamemodeUpdate(string playername, Guid uuid, int gamemode) { }
+        
+        /// <summary>
+        /// Called when the Latency has been updated for a player
+        /// </summary>
+        /// <param name="playername">Player Name</param>
+        /// <param name="uuid">Player UUID</param>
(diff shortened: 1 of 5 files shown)


5tanly:
Just an idea for future pr's, maybe separate them into 1 feature per pr, would make it easier to review.

Also, wouldn't onlatency spam like crazy because players latency is constantly changing?

Nekiplay:
> Just an idea for future pr's, maybe separate them into 1 feature per pr, would make it easier to review.
> 
> Also, wouldn't onlatency spam like crazy because players latency is constantly changing?

This is just an example.

ORelio:
Hi,

Looks cool!

OnLatency() does nothing by default because it calls all bots with the default empty method that does nothing. It's just an additional API for custom bots that may want to use the Latency value.

Regarding player icon, it's nice that you found a workaround for Windows 10! Ideally I'd prefer that you use the Mojang API for this (https://wiki.vg/Mojang_API#UUID_-.3E_Profile_.2B_Skin.2FCape) but it's more difficult that Minotar which seems reliable and free to use, so OK, let's go for it.

Thanks for sending a much cleaner pull request this time and your efforts in documenting the new methods. This makes my review work much faster! 👍

As noted by @5tanly it's also easier to review with only one feature per PR, but I understand that maintaining different branches for each new feature in order to send multiple PR at the same time is complicated. This amount of changes in a single PR is OK for me so you can continue this way.

Issue #1046, opened 2020-06-07
Is there a way to show login (email address) on the window titlebar?


grantemsley:
I've tried setting `consoletitle=%username% - %login%`, but that doesn't work.
Is there a variable for the login that can be used?

I'm starting several clients with a bat script with login and password specified in the command line, and it would make it easier to identify which is which when there are problems if I could show the email address in the window.

Nekiplay:
#1048

Nekiplay:
You could also specify it manually

Nekiplay:
Modify this request pool to your project yourself

PR #1048, opened 2020-06-07
add %login% variable

Nekiplay:


ORelio:
Hi,
1. This PR contains changes from the previous PR #1044 - Cannot merge
2. Could you explain a use case for login? Should contain email address, so not very useful.

grantemsley:
@ORelio #1046 has the use case. 

I launch all my clients with login and password specified on the command line in a script.  When one is acting up, it's a lot easier to find it by it's login in the script than matching the username to the email first.  Basically would save me from having to keep a list of the username for each login.

ORelio:
@grantemsley Seems legit, why not. Will wait for a proper PR.
Also @Nekiplay please document it in [user manual](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/config/README.md) if you add a new built-in %variable% 😉

PR #1050, opened 2020-06-08
Add %login% variable

### LINKED ISSUE #1046: Is there a way to show login (email address) on the window titlebar?

grantemsley:
Fixes #1046 
Replaces #1048 

Work is Nekiplay's, I just copied it into a separate pull request so it can be merged properly, since they hadn't yet.


ORelio:
Thanks, merged!

PR #1053, opened 2020-06-10
add GetGamemode() on ChatBots

Nekiplay:
### **Done**
1. [x]  **GetGamemode() on ChatBots**
2. [x] **Gamemode check for /inventory 0 creativegive**

### DIFF
--- a/MinecraftClient/McTcpClient.cs
+++ b/MinecraftClient/McTcpClient.cs
@@ -1670,7 +1670,7 @@ namespace MinecraftClient
             string playerName = null;
             if (onlinePlayers.ContainsKey(uuid))
             {
-                playerName = onlinePlayers[uuid];
+                if (onlinePlayers[uuid] == username)
                 this.gamemode = gamemode;
             }
         }
--- a/MinecraftClient/McTcpClient.cs
+++ b/MinecraftClient/McTcpClient.cs
@@ -57,6 +57,7 @@ namespace MinecraftClient
         private DateTime lastKeepAlive;
         private object lastKeepAliveLock = new object();
         private int respawnTicks = 0;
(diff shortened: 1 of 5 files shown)


MinecraftClient/McTcpClient.cs
```diff
@@ -1664,5 +1664,20 @@ public void OnHeldItemChange(byte slot)
                 bot.OnHeldItemChange(slot);
             CurrentSlot = slot;
         }
+        
+        public void ChangeGamemode(int gamemode, Guid uuid)
```
ORelio, line 1668: I think this method should be named `OnGamemodeChange` since it's an event callback.

MinecraftClient/McTcpClient.cs
```diff
@@ -1664,5 +1664,20 @@ public void OnHeldItemChange(byte slot)
...
             CurrentSlot = slot;
         }
+        
+        public void ChangeGamemode(int gamemode, Guid uuid)
+        {
+            string playerName = null;
+            if (onlinePlayers.ContainsKey(uuid))
+            {
+                playerName = onlinePlayers[uuid];
+                this.gamemode = gamemode;
```
ORelio, line 1674: Since this event is changing some player gamemode it's not necessarily the client's gamemode. Maybe a check is missing here?

MinecraftClient/McTcpClient.cs
```diff
@@ -1664,5 +1664,20 @@ public void OnHeldItemChange(byte slot)
...
+            }
+        }
+        
+        public void ChangeGamemode(int gamemode)
```
ORelio, line 1678: Maybe also change this method name to `OnGamemodeChange`?

PR #1101, opened 2020-07-02
Add command register for ChatBot

ReinforceZwei:
ChatBot can register their own command for interacting with user


ReinforceZwei:
I want to make the original commands to be loaded first so that if there is a command name conflict, the original command will have a higher priority.

ORelio:
Thanks, looks cool! FYI method parameters and local variables should begin with a lowercase letter, as well for private fields in classes. Only fields beginning with uppercase letters are class names, method names and public fields in classes. I'll try to ajust that.

PR #1102, opened 2020-07-02
Fix for the mail script.

Daenges:
I changed some lines and added the option to change settings through mc chat.

### DIFF
--- a/MinecraftClient/config/Modules/Mail.cs
+++ b/MinecraftClient/config/Modules/Mail.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Linq;
 using System.Data;
 using System.Runtime.Serialization.Formatters.Binary;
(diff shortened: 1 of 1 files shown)


Daenges:
I messed it up a bit. But only one file should be affected by the change.

Daenges:
Now you just need a way to execute  this file somehow. I tried setting up a script version of it, but it didn't work out.

ORelio:
I'll try to convert it to a script for you.

ORelio:
OK, actually it cannot be converted to a Script, because Scripts are compiled on the fly so your [Serializable] class will have a different ID each time it is compiled, thus preventing successful loading of the serialized file on next script load.

A workaround for this would be to use custom text format like the SessionCache.ini for instance. You'd have to build and read the file yourself instead of using serialization, which requires more work, unfortunately. Will push the script to the repository so you can try out.

Also a small suggestion: You can access `Settings.DebugMessages` from your script to check whether to display debug messages or not. This is the global setting in MCC from the `debugmessages` field in `MinecraftClient.ini` 😉

Daenges:
First, thanks for all your effort! I really appreciate your engagement in this project!
Second, I will change the option settings to the global one in my next commit! Thanks for this hint!
Third, is there any way to build this file in the project, without rewriting the whole script?

Daenges:
Maybe I could delete the whole options class, if the program is able to use the normal config file.

PR #1108, opened 2020-07-05
Updating and including the mail script to the bot list.

Daenges:
As requested, I added the synchronization between the config debugmessage bool and the one in the script. Furthermore, I added a way to send anonymous mails by writing "tellonym". The messages are saved as normal mails, which lets them count to the mail cap, due to the fact that the host knows the sender.

I added the script to the chat bots because of the problems, that the scripts have with serialization. Instead of rewriting the whole serialization part, my idea was to add the script to the other chat bots, to avoid the compiling issues. Then the serialization would work perfectly fine.
After this you could remove the option class at some point and move all the settings to the config file with the addition to activate the whole script.


Daenges:
The problem is, that I don't know neither how to add a section to the config file, nor how to autoload the script like the other bots. :/

Daenges:
You could also just put this script to the other bots and write a description in the config to tell the users how to customize the bot through the minecraft chat.

Daenges:
Okay! I got a working version! I didn't understand some code, so please correct me if I made something wrong!

ORelio:
Well, MCC itself doest not use serialization because it makes thing harder to troubleshoot, users cannot view the content of serialized filed. That is why everything is in the INI (config...) or CSV format (session cache...). To integrate your bot as a built-in MCC ChatBot, I'll need to make sure it does not introduce any issue (including security bugs) and it would be better to use the same file formats as the others. I'll have a look later when I have more time to evaluate your pull request. 😉

Daenges:
Your concerns are justified. Maybe some characters cause problems while beeing serialized.

Daenges:
Also added the option to toggle the autorespawn of this script and wrote comments to all commands and functions.

Daenges:
I created a "Bot Manual" which shows all commands with a small description of what they are doing. But I don't know where to upload it. There is no folder of manuals and directly in the bot folder would be strange too. :/

Daenges:
Basically all file actions are now handeled in one function to guarantee through a interval (timer) that two (or more) bots won't interfere with the files.
I also added some new commands and removed old ones, which got useless due to the changes.
:D

ORelio:
Thanks. Still hadn't time to review & convert serialized files to human-readable ones. Regarding configuration and help, just name your files with the exact same name as the bot, I think I'll place them in the config/ChatBots optional folder.

Daenges:
@ORelio if you have any tips for my code (not exclusively to the serialization part), please let me know!

ORelio:
I still haven't had time to properly review your PR but FYI there might be some issues in the bot owner stuff due to nickname changes since your bot has many admin commands: #1142. To protect against this, a solution would be to move the admin commands to console only instead of PMs e.g. through [`RegisterChatBotCommand`](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/ChatBot.cs#L1179).

Daenges:
@ORelio I'm sorry, but I don't understand, how to add a command through this. Could you give me an example on how to add a command?

Daenges:
So I could technically add all moderation commands as normal commands, but for only one script that would be too many classes.

ORelio:
Here is an example taken from [AutoCraft](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/ChatBots/AutoCraft.cs):
```C#
public override void Initialize()
{
    RegisterChatBotCommand("mycommand", "My Custom Command", CommandHandler);
}

public string CommandHandler(string cmd, string[] args)
{
    if (args.Length > 0)
    {
        switch (args[0])
        {
            case "test":
                return "Command Result!";
            case "help":
                return "Available commands: test.";
        }
    }
    return "Custom command. Try /mycommand test or /mycommand help."
}
````

**Comment by Daenges** [comment]
Ahh. Okay. I try my best to implement it!

**Comment by Daenges** [comment]
Theoretically players could also exploit the "MaxMailsPerPlayer" setting, but therefor the "MaxSafedMails" setting exsists.

**Comment by Daenges** [comment]
Okay, that were many features at once! To summarize: I added a few moderation commands and reworked the command reading.

**Comment by ORelio** [comment]
Hi, just had a quick look at your changes. That's quite a lot of commands! Do you think it would be cleaner to register only one command such as `/mail` and act on the first argument e.g. `/mail addignored`, `/mail changemailpath` and so on?

**Comment by Daenges** [comment]
> 
> 
> Hi, just had a quick look at your changes. That's quite a lot of commands! Do you think it would be cleaner to register only one command such as `/mail` and act on the first argument e.g. `/mail addignored`, `/mail changemailpath` and so on?

It would be a considerable sollution! Never thought about that. I will try it out this way in ne next commit! The only concern I have, is that instead of a 13 line code block I would need several if statements in a seperat fuction to extract the arguments and search the right function for the command. 
Maybe if I add further Dictionaries, for instance to safe all integer values... Then I could directly extract the value from the args.
I will try around a bit. :D

**Comment by ReinforceZwei** [comment]
You can use `switch` statement to replace multiple if for handling arguments.

**Comment by Daenges** [comment]
My idea was to use a dictionary <string, method> so that i can directly convert the given arguments and convert them into an action. Then I only need to parse all arguments to the method, like the register command does. Thereby I would avoid many if/switch statements.

**Comment by Daenges** [comment]
@ReinforceZwei Okay, I looked it up. To execute a method through a dictionary, I need to make it static (to convert it into a Func<> statement), which means, that I need to rework all methods and the option class due to the extensive dependincies between each other. I think it is easier, to just write a big switch tree, which analyzes the syntax instead of changing every method.

**Comment by Daenges** [comment]
Oh.. just realized, that due to the new syntax I need to alter a few numbers.

**Comment by Daenges** [comment]
Okay, so apparently even microsoft is disadvising the use of the binary formatter. On top of that @ORelio mentioned, that the host should be able to read the safed data. This brings me to the point that I want to change the serialization method.
I had the idea of saving the data to json, but every serialization method I tried failed, because the class has start parameters.
Do you have any suggestions on which method I should use and how to implement it?
https://docs.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide

**Comment by ORelio** [comment]
How about some kind of INI or CSV format? Would not be super hard to create or parse.

Example for CSV in SessionCache:
https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Protocol/Session/SessionCache.cs#L201
Only issue here, you need to define a separator character that is not allowed in mails.

Example for INI in AutoRespond:
https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/ChatBots/AutoRespond.cs#L132
Here you can define field=value with values containing the = character but it's a bit more involved.

If you're interested, I can include IniFile.cs to the repository:
https://github.com/ORelio/Sharp-Tools/blob/master/Utility/INIFile.cs
It's for saving and loading dictionaries as INI files. Each section has a name and contains a dictionary of key/values so each mail could be saved as a Mail1, Mail2, Mail3... section in the INI file. You would still need to retrieve fields from each section and populate the class from the dictionary.

**Comment by Daenges** [comment]
@ORelio So your idea would be, to make an own serialization method, based on AutoRespond and SessionCache? Your dictionary serialization is pretty cool, but my mails are safed in an own class, and I don't think it is good to put them into dictionaries. The only way I could use it, would be for the option class, but even there I only use the dictionaries for some integer and bools.
I am not very familiar with serialization methods, do I need to obbey any rules if I create my own format?

**Comment by ORelio** [comment]
MCC uses two different serialization formats that have a very simple structure:

1. INI for most configuration files
```ini
[SectionName]
# Comment
field=value
```
You can have several sections with field=value fields, where value can be anything.

2. CSV for session cache
```csv
# Comment
# sender,recipient,message
ORelio,Daenges,Hello world!
```

In both cases you need to parse it yourself, that is why I gave some examples to illustrate.
INI should not have duplicate section names, although the AutoRespond bot does it.
CSV is easier to create and read, but only has fixed fields, one per column.

INIFile.cs gives an example on how to read and write INI files. You need to build a dictionary, then give it to the class and it will generate an INI file. Conversely, it gives you a dictionary and you need to extract fields to populate your classes. So there is still a bit of work to do.

In MCC, the INI file is read directly and fields are populated to the class directly. See the examples linked above for code examples. Generating an INI file is quite easy as well, you can have a counter and iterate over your mails, then append sections to your INI file:

```ini
[Mail1]
sender=ORelio
recipient=Daenges
message=Hello world!
delivered=false
```

Then `[Mail2]` and so on. However as stated above if you parse it manually as in AutoRespond you can have duplicate section names so everything can be in `[Mail]` sections.

Daenges:
`public void serializeMail(Message[] msg_list)
        {
            Dictionary<string, Dictionary<string, string>> messageContent = new Dictionary<string, Dictionary<string, string>>();
            int mailnum = 0;

            foreach (Message msg in msg_list)
            {
                messageContent.Add("mailnum" + mailnum.ToString(), new Dictionary<string, string>());

                messageContent["mailnum" + mailnum.ToString()].Add("sender", msg.GetSender());
                messageContent["mailnum" + mailnum.ToString()].Add("destination", msg.GetDestination());
                messageContent["mailnum" + mailnum.ToString()].Add("content", msg.GetContent());
                messageContent["mailnum" + mailnum.ToString()].Add("timestamp", msg.GetTimeStamp().ToString());
                messageContent["mailnum" + mailnum.ToString()].Add("anonymous", msg.isAnonymous().ToString());

                // Parse to INI Serializer => messageContent
                mailnum++;
            }
        }`
@ORelio so something like this?

ORelio:
Export messages:
```C#
Dictionary<string, Dictionary<string, string>> iniFileDict = new Dictionary<string, Dictionary<string, string>>();
for (int msgNum = 0; i < msgList.Length; i++)
{
    Dictionary<string, string> iniSection = new Dictionary<string, string>();
    iniSection["sender"] = msg.GetSender();
    iniSection["destination"] = msg.GetDestination();
    iniSection["content"] = msg.GetContent();
    iniSection["timestamp"] = msg.GetTimeStamp().ToString();
    iniSection["anonymous"] = msg.isAnonymous().ToString();
    iniFileDict["mail" + msgNum] = iniSection;
}
INIFile.WriteFile("MailDatabase.ini", iniFileDict, "Mail Database");
```
Import messages:
```C#
Dictionary<string, Dictionary<string, string>> iniFileDict = INIFile.ParseFile("MailDatabase.ini");
List<Message> messages = new List<Message>();
foreach (KeyValuePair<string, Dictionary<string, string> iniSection in myDictionary)
{
    //iniSection.Key is "mailXX" but we don't need it here
    string sender = iniSection.Value["sender"];
    string destination = iniSection.Value["destination"];
    string content = iniSection.Value["content"];
    DateTime timestamp = DateTime.Parse(iniSection.Value["timestamp"]);
    bool anonymous = INIFile.Str2Bool(iniSection.Value["anonymous"]);
    messages.Add(new Message(string sender, string destination, string content, anonymous)); //TODO timestamp
}
```

Daenges:
@ORelio Now you only need to add the INIFile.cs to the MinecraftClient folder. Or should I add it to the request?

ORelio:
you can add it to the pull request. Just change its namespace from `namespace SharpTools` to `namespace MinecraftClient` and put it in the same folder as `Json.cs` 😉

Nekiplay:
> @ORelio Now you only need to add the INIFile.cs to the MinecraftClient folder. Or should I add it to the request?

I am use Class
https://pastebin.com/xysb0fTX

ORelio:
I think `DllImport("kernel32.dll")` will not work on Mac and Linux, only Windows.

Nekiplay:
> I think `DllImport("kernel32.dll")` will not work on Mac and Linux, only Windows.

https://www.nuget.org/packages/ini-parser/
and use //dll & //using

Daenges:
I think it is uneccessary to import external, unknown code, if we have a working sollution, which is simple and made by ORelio. Doing so, means more time to analyze the code for weak points, which could cause bugs or exploits.
ORelio wants to reread the code, to double check its security. If we now add more libraries etc. he will never finish checking the code due to his rare spare time. On top of that I would need to rewrite thw whole serialization part AGAIN, just to achieve a similar result, which may even be bugged.

ORelio:
I'll have more time next week to review your pull request 🙂

Daenges:
No rush! I just wanted to mention, that it would take much more time to analyze code, which is completely strange and might not even be suitable for my needs.
But I am glad to hear about your encouragement in this project! 😄

ORelio:
Good news, I finally had the time to fully review your pull request. In order to better integrate to the existing code base and make the bot easier to maintain, I refactored parts of the code to keep the core functionality while removing some settings that may not be useful outside debugging cases. As a result, the bot should be easier to maintain on the long run.

Quoted from 5c43e1208e8a1cd708306ce08925ecc87e4b25b8:
```
Rename Mail to Mailer
Move options to MinecraftClient.ini
Make the bot much simpler by removing some settings
Create specific MailDatabase and IgnoreList classes
However the core functionality for users is the same

Settings removed:
 - allow_sendmail: Cannot use Mailer if it's disabled
 - allow_receivemail: Cannot use Mailer if it's disabled
 - path_setting: Settings moved to MinecraftClient.ini
 - debug_msg: MCC already has a setting for that with LogDebugToConsole()
 - auto_respawn: MCC already has a built-in auto-respawn feature
 - allow_selfmail: Is it really necessary to block self mails? ;)
 - maxcharsinmsg: Automatically calculated based on max chat message length
 - timeinutc: DateTime is not show to the recipient so I think it's not absolutely necessary
 - interval_sendmail: Set to 10 seconds for now

Internal Commands removed:
 - changemailpath: Now a static setting in config
 - changesettingspath: Now a static setting in config
 - updatemails: Already updated every 10 seconds
 - getsettings: Shown on startup with debugmessages=true
 - resettimer: Seems only useful for debugging
 - setbool: Settings are static in config file
 - setinteger: Settings are static in config file

All user commands are retained:
 - mail
 - tellonym
```

After this review, I think the Mailer bot is a very nice thing, and the idea of passing mails between servers is very clever. I'm happy to tell you that Mailer is now an official Minecraft Console Client bot 👏

Daenges:
You did great work.
Nevertheless, you removed some essential commands.
- allow_selfmail was there, to protect the bot from unneccessary traffic.
- maxcharsinmail was there, to control the length of mails and save disk space. (Save more messages with less content)
- resettimer is VERY iportant if you want to run a cross server system. With this command you can control when the interval starts, so that the bots won't interfere with eachother, while deserializing and serializing the mails. Otherwise you would need to start the .exe files with a certain delay, which is not very precise.

The first two are just some moderation commands, which, in my opinion, would still improve the bot. But the last one is neccessary to guarantee that the bot also works over several instances.

But all in all I'm proud to see my bot in the console client! 😃

Daenges:
@ORelio Okay, I deleted my last comment, because I saw, that you changed very much and my argument didn't make any sence. 😅 I am very proud to see my script in the console client. It is very impressing how much you could compress my code! Thanks for your collaboration, it was awesome! ;D

ORelio:
Hi,

Actually, the current synchronization mechanism is not very good because it does not handle the case where two clients will want to write the file at the same time, or might miss changes introduced between refreshes if several clients take new mails in a short time span, so there is room for improvement. Resettimer is not the correct way of solving that, we need to add retries on write, and file system watcher as in session cache mechanism ([SessionCache.cs](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Protocol/Session/SessionCache.cs)).

Although Selfmail is not very useful, maybe the user will want to do this in order to try out the bot? A Minecraft chat message is capped to 256 characters so maxMessageLength will never be very large. There is not much to worry about about disk space, let's assume a total of 500 characters per message in INI file, that gives us 500 * 2000 = 1MB with default settings where maximum capacity is set to 2000 messages 😄

Daenges:
Oh, thanks for your reply! I think i will try to understand the refactored code through the next days and look for improvements.

ORelio:
Moved the SessionFileMonitor into a generic class and implemented synchronization with FileMonitor inside Mailer 😉

Daenges:
Finally a good and easy way if serialization in the client!

ORelio:
I guess you meant "synchronization"

PR #1120, opened 2020-07-09
Return exit codes for specific failures when in non-interactive mode

### LINKED ISSUE #1116: Return exit codes on login failure, server disconnect
Is there any way I can tell what caused a failure from my batch file?  Many programs set the exit code/errorlevel differently depending on why they exited, but MCC doesn't seem to.

I have exitonfailure turned on, and a batch file that restarts the console client when it exits. Now I'm trying to find a way to differentiate between login failures and server problems, so I can have my console client stop trying on login failures, but get restarted for everything else.  That would help avoid getting your IP blacklisted from trying bad passwords too many times.

Any suggestions on how I could do that?

grantemsley:
Fixes #1116 

When in non-interactive mode (exitonfailure is enabled), returns distinct exit codes for some of the known failures.

This will let users handle them differently in their scripts.  For example I have a script that restart the console client if it's killed or crashes.  Right now I do something like this:

```
:start
MinecraftClient.exe
timeout /t 60
goto start
```

Now I can do this:
```
:start
MinecraftClient.exe
if '%ERRORLEVEL%'=='4' goto loginfailed
timeout /t 60
goto start

:loginfailed
echo 'Login failure. Stopping.'
pause
```

Doing this should avoid hammering the authentication servers with a bad password, and prevent getting IP blacklisted as easily.

### DIFF
--- a/MinecraftClient/Program.cs
+++ b/MinecraftClient/Program.cs
@@ -324,14 +324,14 @@ namespace MinecraftClient
         /// <summary>
         /// Disconnect the current client from the server and exit the app
         /// </summary>
-        public static void Exit()
+        public static void Exit(int exitcode = 0)
         {
             new Thread(new ThreadStart(delegate
             {
                 if (client != null) { client.Disconnect(); ConsoleIO.Reset(); }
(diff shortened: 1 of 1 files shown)


ORelio:
Seems good to me. Thanks for your contribution!

PR #1127, opened 2020-07-12
Entity command

Nekiplay:
### Done

- [x] Entity command

### DIFF
--- a/MinecraftClient/Commands/Entitycmd.cs
+++ b/MinecraftClient/Commands/Entitycmd.cs
@@ -9,7 +9,7 @@ namespace MinecraftClient.Commands
     class Entitycmd : Command
     {
         public override string CMDName { get { return "entity"; } }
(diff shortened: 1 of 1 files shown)


MinecraftClient/Commands/Entitycmd.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Linq;
+using System.Text;
+using MinecraftClient.Inventory;
+
+namespace MinecraftClient.Commands
+{
+    class Entitycmd : Command
+    {
+        public override string CMDName { get { return "entity"; } }
+        public override string CMDDesc { get { return "entity <id> <list|attack|use"; } }
```
ReinforceZwei, line 12: Missing a `>`, `entity <id> <list|attack|use`

Nekiplay: Thx

PR #1145, opened 2020-07-24
Remove local function syntax in AutoCraft

ReinforceZwei:
Seems like AppVeyor do not support this syntax. [MSDocs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions)
image

### DIFF
--- a/MinecraftClient/ChatBots/AutoCraft.cs
+++ b/MinecraftClient/ChatBots/AutoCraft.cs
@@ -332,7 +332,7 @@ namespace MinecraftClient.ChatBots
                 string value = line.Substring(key.Length + 1);
                 switch (section)
                 {
-                    case "recipe": parseRecipe(key, value); break;
+                    case "recipe": parseRecipe(key, value, lastRecipe); break;
                     case "autocraft": parseMain(key, value); break;
                 }
             }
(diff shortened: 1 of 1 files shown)


ORelio:
Ah, yes. You cannot do this in C#4

ReinforceZwei:
@ORelio The build process still fails. The problem line is `Enum.TryParse(value, out ItemType itemType)`

ORelio:
Ah, also it's not possible to declare an `out` variable directly in method call. Fixed.

ReinforceZwei:
Oh I see. My visual studio told me I could do that.

ORelio:
Yes, because on newer C# versions you can do this. The project is still using C#4.0 because it's supported on older systems and Mono versions. We might migrate to .NET 5 once available since it would have a Linux runtime hopefully better than Mono.

Mhowser:
.NET 5 is out https://dotnet.microsoft.com/download/dotnet/5.0

ReinforceZwei:
Isn't it a preview version? At least we should wait for the release version.

Mhowser:
Well, it might be a good idea to start porting over to the preview versions. That way the heavy work will be done by the time .NET 5 is fully released. But I'm not a developer so I don't know.

ORelio:
It's better to wait for it to be stable and generally available to all major operating systems. Once we migrate to it, the project won't compile and run on .NET 4 runtimes anymore so the win / lose tradeoff must be worthwhile.

PR #1152, opened 2020-07-28
Add FindBlock on GetWorld

Nekiplay:
### Done
- [x] List<Block> FindBlock


ORelio:
Good idea. Adding this to the repository.
Remark: Please avoid copy-and-paste when only 1 line is different. Also do not forget the /// documentation! 😉

PR #1165, opened 2020-08-01
Add Auto download dlls

Nekiplay:
### Done
- [x] Pull Request


ORelio:
Seems dangerous because it automatically downloads and install a library without letting the user check its content. It would be safer and easier to provide both the script and the DLL to your users rather than downloading it afterwards.

PR #1176, opened 2020-08-03
Added ToLower to player check

Daenges:
Added a toLower() to the check, if the Recipient is online. Otherwise the mail would not be sent, just because one letter was higher or lower.

### DIFF
--- a/MinecraftClient/ChatBots/Mailer.cs
+++ b/MinecraftClient/ChatBots/Mailer.cs
@@ -279,8 +279,8 @@ namespace MinecraftClient.ChatBots
                 ignoreList = IgnoreList.FromFile(Settings.Mailer_IgnoreListFile);
 
                 // Process at most 3 mails at a time to avoid spamming. Other mails will be processed on next mail send
-                HashSet<string> onlinePlayer = new HashSet<string>(GetOnlinePlayers());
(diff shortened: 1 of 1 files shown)


Daenges:
Oh.. on cracked servers there could be more than one way of spelling the name! Ooops my fault! =P

ORelio:
I think you are right, I did the Lowercase thing when taking the mail, but not when delivering it. Fixing.

Daenges:
A problem is, that I think the char combination of a premium/mojang account is unique, but on non premium servers (without verification that you bought the game) it is possible, that players can have the same name with different capitalization. Lowering the names would lead to an incompatibility to non premium servers.

ReinforceZwei:
Actually the username are case-insensitive, regardless mojang account or offline account.

Daenges:
Strange.. because I was able to put two accounts with the same name, but different capitalization on my local server.

ReinforceZwei:
Maybe it's my misunderstanding of offline account, but for mojang account, I am sure it is case-insensitive.

ORelio:
Plugins are often case insensitive so players having the same name with only capitalization change will have issues in many plugins, not just the mailer bot in minecraft console client. 😄

Issue #1180, opened 2020-08-05
Tab List


5tanly:
**Prerequisites**
 - [x] I have read and understood the [user manual](https://github.com/ORelio/Minecraft-Console-Client/tree/master/MinecraftClient/config)
 - [x] I made sure I am running the latest [development build](https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/artifacts)
 - [x] I tried to [look for similar feature requests](https://github.com/ORelio/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

**Feature description**
Is there currently a way to view the tab list in this program?
I know there is /list as a command that returns a list of all the players but on some servers there are prefixes and suffixes to names. I assume it would have to be this packet to be added, if its not already. https://wiki.vg/Protocol#Player_Info

I have no idea how to post pictures but this is what I mean
https://proxy.mc-market.org/6803f8c7c4b8e6d2e672baa9e0c412a1a8580397?url=https%3A%2F%2Fgyazo.com%2F2a1445720063b7a41339e43e9b724fe8.png

Pokechu22:
IIRC `/list` is implemented using the data from the player info packet (hence it being a command built into MCC instead of using the server's `/list`), but I guess it doesn't include the other metadata such as prefixes and suffixes.

5tanly:
Thanks!
I tried [searching the code](https://github.com/search?l=&q=GetOnlinePlayers+repo%3AORelio%2FMinecraft-Console-Client&type=Code) and only came up with 7 results.
None of them seem to be the code I'm looking for. Maybe I just dont have a clue how github search works

ORelio:
If you need it, you can add OnPlayerJoin/OnPlayerLeave to the ChatBot API, submit that addition in a pull request, and use it from a ChatBot for anything you need.

Nekiplay:
closed?

ORelio:
I think not, because @5tanly wants to access prefix and suffix, which is currently ignored by MCC. So he can't retrieve them yet unless making changes to the protocol handler and ChatBot API.

Nekiplay:
To do this, as I understand it, you need to add Teams
![image](https://user-images.githubusercontent.com/35975332/89716297-3473a400-d9c5-11ea-961f-0c44a9161767.png)

ORelio:
Exactly. Currently, Teams are not processed at all.

PR #1181, opened 2020-08-06
Add player join/leave event to ChatBot API

ReinforceZwei:
#1180


ORelio:
Thanks!

ORelio:
Noticed just after merging that I accidentally introduced a bug. Fixed in 2034729f743445e82ac63f3637ab65aa8045b9f7.

PR #1195, opened 2020-08-11
Add entity handling for 1.10 to 1.12.2

ReinforceZwei:
Tested by using `/entity` command on vanilla server.

Actually EntityPalette113 cannot be used in 1.13 below so I generated a new EntityPalette112 by using custom javascript script (I don't know how to use the C# one)


ORelio:
My bad, I thought that randomization of those IDs only occured in Minecraft 1.14.

For palette generators in MCC you need to generate the data reports as stated in the wiki https://wiki.vg/Data_Generators then add a call to the palette generator in Program.cs, launch the project once and it'll generate the C# files for you.

However, the C# entity palette generator included in MCC works only for 1.14 and greater so the 1.13 palette was created by hand from the Wiki, so you needed to generate the 1.12 palette otherwise anyway.

ReinforceZwei:
Oh I see. Then should I include the script that I used into this repo?

ORelio:
You may, although it might not be of use unless versions lower than 1.10 also have randomized IDs
If the script is simple you might as well put it here as a code snipped 😉

PR #1205, opened 2020-08-14
Implemented entity metadata and entity health

ReinforceZwei:


Nekiplay:
Good job

ORelio:
Seems like Entity Metadata IDs change between Minecraft versions:
https://wiki.vg/Pre-release_protocol#Entity_Metadata
Are you sure it's `8` for all versions between 1.13 and 1.16?

ReinforceZwei:
`1.10+  -> 7`
`1.14+ -> 8`
_Not sure is 1.14+ or 1.14.4+ as the history record is kinda messy and I can only find one record for 1.14.4_

ORelio:
That's right, I got the same by looking to Entity_Metadata's history:
1.13 https://wiki.vg/index.php?title=Entity_metadata&oldid=14277#Living 7
1.14 https://wiki.vg/index.php?title=Entity_metadata&oldid=15239#Living 8
So it's dependent on the Minecraft version

Pokechu22:
It is dependent on Minecraft versions.  If you want to see a specific version, check https://pokechu22.github.io/Burger/1.14.4.html or similar (also available in JSON at https://pokechu22.github.io/Burger/1.14.4.json).  The old documentation on pre-release protocol linked from https://wiki.vg/Protocol_version_numbers also should have information about metadata.

MinecraftClient/McClient.cs
```diff
@@ -59,6 +59,7 @@ public class McClient : IMinecraftComHandler
         private object lastKeepAliveLock = new object();
         private int respawnTicks = 0;
         private int gamemode = 0;
+        private int protocolVersion;
```
ORelio, line 62: `McClient.cs` should not need to use the protocol version. Protocol-dependant stuff should be set in the protocol handler. Maybe add a OnEntityHealth event in `McClient.cs` and call it from the protocol handler?

ReinforceZwei: OK. Give me some time

PR #1208, opened 2020-08-15
Implement inventory mouse item dragging

ReinforceZwei:
Example usage
```cs
RegisterChatBotCommand("drag", "drag", (string cmd, string[] args) => {
    // cmd args: <target slot to use> <slots to drag on...>
    var arg = args.ToList();
    int slot = int.Parse(arg[0]);
    arg.RemoveAt(0);

    // put item to cursor
    WindowAction(0, slot, WindowActionType.LeftClick);
    // start dragging
    WindowAction(0, -999, WindowActionType.StartDragRight);
    // send slots to drag on (make sure slot is empty)
    foreach (var i in arg)
    {
        WindowAction(0, int.Parse(i), WindowActionType.AddDragRight);
    }
    // end dragging
    WindowAction(0, -999, WindowActionType.EndDragRight);
    return "ok";
});
```
_Slots to drag on can be any empty slot, even they aren't next to each other (which you can't achieve in vanilla client :D)_

### DIFF
--- a/MinecraftClient/Inventory/WindowActionType.cs
+++ b/MinecraftClient/Inventory/WindowActionType.cs
@@ -5,22 +5,84 @@ using System.Text;
 
 namespace MinecraftClient.Inventory
 {
+    /// <summary>
+    /// Represents mouse interactions with an inventory window
+    /// </summary>
     public enum WindowActionType
     {
+        /// <summary>
+        /// Left click with mouse on a slot: grab or drop a whole item stack
+        /// </summary>
         LeftClick,
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


ORelio:
Looks good! Would be nice to also have a built-in command for it, but it's fine nonetheless.
Added documentation for window action types because it was starting to get confusing 😉

ReinforceZwei:
@ORelio The description for WindowActionType DropItem/stack is slightly incorrect. Those two actions do not require grabbing the items before dropping them.

ORelio:
@ReinforceZwei Indeed.

Issue #1209, opened 2020-08-16
Entity handling not completely work on 1.16.1


ReinforceZwei:
![image](https://user-images.githubusercontent.com/39955851/90329790-34cdf980-dfda-11ea-9c48-010c2851464a.png)
_Google translation: An item with the same index key has been added_

ReinforceZwei:
![image](https://user-images.githubusercontent.com/39955851/90332514-183cbc00-dff0-11ea-94b7-5b9b92e07457.png)

ReinforceZwei:
EntityEquipment is fixed.

For metadata, after I added a debug line trying to see which key is duplicated, I can't reproduce the bug again.

PR #1211, opened 2020-08-16
Fix 1.16 entity metadata and equipment

ReinforceZwei:
#1209

### DIFF
--- a/MinecraftClient/Protocol/Handlers/DataTypes.cs
+++ b/MinecraftClient/Protocol/Handlers/DataTypes.cs
@@ -624,8 +624,6 @@ namespace MinecraftClient.Protocol.Handlers
                     default:
                         throw new System.IO.InvalidDataException("Unknown Metadata Type ID " + type + ". Is this up to date for new MC Version?");
                 }
-                // issue #1209: An item with the same index key has been added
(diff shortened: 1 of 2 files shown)


MinecraftClient/Protocol/Handlers/DataTypes.cs
```diff
@@ -610,22 +610,27 @@ private object ReadNbtField(Queue<byte> cache, int fieldType)
...
+                        throw new System.IO.InvalidDataException("Unknown Metadata Type ID " + type + ". Is this up to date for new MC Version?");
                 }
-                data.Add(Key, Value);
-                Key = ReadNextByte(cache);
+                // issue #1209: An item with the same index key has been added
+                // That shouldn't happen, but if yes then we over-write the existing one
+                if (data.ContainsKey(key))
+                    data[key] = value;
+                else
+                    data.Add(key, value);
```
ORelio, line 632: You can use `data[key] = value` directly even if the entry is not yet in the dictionary. No need to check before setting the value 😉

ReinforceZwei:
That commit didn't fix the issue so reverted it. Will be fixed in 1.16.2 branch.

PR #1213, opened 2020-08-16
/inventory add show item Damage

Nekiplay:
### Done
- [x] Pull request

### DIFF
--- a/MinecraftClient/Commands/Inventory.cs
+++ b/MinecraftClient/Commands/Inventory.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using MinecraftClient.Inventory;
-
 namespace MinecraftClient.Commands
 {
     class Inventory : Command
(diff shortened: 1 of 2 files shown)


MinecraftClient/Commands/Inventory.cs
```diff
@@ -1,9 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using MinecraftClient.Inventory;
-
```
ORelio, line 6: This empty line should be kept for readability

MinecraftClient/Commands/Inventory.cs
```diff
@@ -86,8 +84,21 @@ public override string Run(McClient handler, string command, Dictionary<string,
                                 {
                                     string displayName = item.Value.DisplayName;
                                     if (String.IsNullOrEmpty(displayName))
-                                        response.Add(String.Format(" #{0}: {1} x{2}", item.Key, item.Value.Type, item.Value.Count));
-                                    else response.Add(String.Format(" #{0}: {1} x{2} - {3}§8", item.Key, item.Value.Type, item.Value.Count, displayName));
+                                    {
+                                        Dictionary<string, Object> itemnbt = item.Value.NBT;
+                                        if (itemnbt.ContainsKey("Damage"))
```
ORelio, line 89: Instead of implementing NBT parsing in the command, I think it would be better to implement a `Damage` getter in `Item.cs` just like [`DisplayName`](https://github.com/ORelio/Minecraft-Console-Client/blob/4cc29a6ee662613eb3e4d916cc5bda3f699a029a/MinecraftClient/Inventory/Item.cs#L56). The getter could return `0` or `-1` if no NBT damage is set for the item (as you prefer).

MinecraftClient/Commands/Inventory.cs
```diff
@@ -153,15 +164,16 @@ public override string Run(McClient handler, string command, Dictionary<string,
...
+                    {
                         response.Add(String.Format(" #{0}: {1}", inventory.Key, inventory.Value.Title + "§8"));
+                    }
                     response.Add(CMDDesc);
                     return String.Join("\n", response);
                 }
             }
             else return "Please enable inventoryhandling in config to use this command.";
         }
-
```
ORelio, line 164: This empty line should be kept for readability

PR #1214, opened 2020-08-17
Implement MC version 1.16.2 support

ReinforceZwei:
All packets ID update done
_Tested in 1.16.2 craftbukkit server_

Edited:
- [x] Entity Handling - done
- [x] Inventory Handling - done


ReinforceZwei:
https://wiki.vg/index.php?title=Pre-release_protocol&oldid=15990#Entity_Properties
The key naming style got changed in 1.16, from `oldStyle` to `new_style`. Since protocol-dependent stuff should be handled within the protocol handler, should we convert those key back to old style (or old style to new style)?

ReinforceZwei:
Entity equipment is fixed in the commit above.

ReinforceZwei:
Item ID got changed in 1.16+ so I added palette for item.

ORelio:
Good job! Seems like you are still working on it so for now I'm just adding a commit to fix megeability after merging #1211 to the main repository.

ORelio:
> The key naming style got changed in 1.16, from `oldStyle` to `new_style`. Since protocol-dependent stuff should be handled within the protocol handler, should we convert those key back to old style (or old style to new style)?

Were we using that anywhere in Minecraft Console Client?

ReinforceZwei:
Yep, I am working on entity properties and almost done.

> Were we using that anywhere in Minecraft Console Client?

@ORelio  AutoAttack for calculating attack speed.

ReinforceZwei:
Mojang also appended `minecraft:` to the start of the name and it is not documented. 
image

ORelio:
Ok, for now I'd say to just try both names in AutoAttack:
```C#
        public override void OnPlayerProperty(Dictionary<string, double> prop)
        {
            foreach (attackSpeedKey in new[] { "generic.attackSpeed", "minecraft:generic.attack_speed" })
            {
                // adjust auto attack cooldown for maximum attack damage
                if (prop.ContainsKey(attackSpeedKey))
                {
                    if (attackSpeed != prop[attackSpeedKey])
                    {
                        serverTPS = GetServerTPS();
                        attackSpeed = prop[attackSpeedKey];
                        attackCooldownSecond = 1 / attackSpeed * (serverTPS / 20.0); // server tps will affect the cooldown
                        attackCooldown = Convert.ToInt32(Math.Truncate(attackCooldownSecond / 0.1) + 1);
                    }
                }
            }
        }
```
Converting could be troublesome because some names did not change such as `generic.reachDistance` so you can't just change case everywhere and expect it to work. Regarding leading `minecraft:` that's the default behavior of identifiers when no namespace is specified:

https://wiki.vg/index.php?title=Pre-release_protocol&oldid=15990#Entity_Properties

image

https://wiki.vg/Protocol#Identifier

> Identifiers are a namespaced location, in the form of `minecraft:thing`. If the namespace is not provided, it defaults to `minecraft`  (i.e. `thing` is `minecraft:thing`.)

ReinforceZwei:
Obviously your solution is way better than mine. Updated using your code.

ReinforceZwei:
Everything are done.

ORelio:
Should be good now. Are the additional commits OK for you?

ReinforceZwei:
Yes, no problem.

PR #1215, opened 2020-08-17
Fix AutoCraft still being loaded when inventory handling is disabled

ReinforceZwei:


ReinforceZwei:
We are removing the same line in AutoCraft :laughing: 
I will remove mine from the commit.

ORelio:
Sorry, was working on #1214 and saw the compile Notice about unused variable so I though I could remove it by the way 😅

Issue #1144, opened 2020-07-23
SetSlot crash 1.15.2


Nekiplay:
**Prerequisites**
 - [x] I made sure I am running the latest [development build](https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/artifacts)
 - [x] I tried to [look for similar issues](https://github.com/ORelio/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one
 - [x] I have set `debugmessages=true` in config to diagnose my issue
 - [x] I have redacted session tokens and passwords before attaching screenshots

MC Version 1.15.2
Server Core 1.8.8 Paper Spigot
![image](https://user-images.githubusercontent.com/35975332/88343238-68f72700-cd5a-11ea-847a-42c7c833807a.png)

ORelio:
Maybe an issue with packet ID?

Nekiplay:
Now I will try to solve this problem

Nekiplay:
Something to do with checking ContainsKey judging by the error
![image](https://user-images.githubusercontent.com/35975332/90340708-88106e00-e013-11ea-8599-c92e19ab91b6.png)

Nekiplay:
There was no such problem before

Nekiplay:
Something tells me that she appeared after https://github.com/ORelio/Minecraft-Console-Client/pull/1107/commits/a901c7103151df84cad9fa2314d1d2cf3105d484#diff-0f0a6c52bf7703cdaa878785348ebb98

ORelio:
Seems likely, because `inventories[0]` is accessed without checking. Seems like a special case for storing item under cursor while moving items in inventories. Maybe the SetSlot() happens before the inventory is initialized.
https://github.com/ORelio/Minecraft-Console-Client/blob/a901c7103151df84cad9fa2314d1d2cf3105d484/MinecraftClient/McClient.cs#L1549

ReinforceZwei:
If it really the case that setslot happened before windowopen, maybe we could assume the inventory 0 is present and always initialize it ourselves before the windowopen event.

ORelio:
Yes or store the item at hand in a separate variable outside the inventory no. 0

ReinforceZwei:
But we still need to store it within the inventories list otherwise we cannot access to it with the current getter method. It might become more complicated if we store it separately.

ORelio:
[...]

PR #1217, opened 2020-08-18
Fix SetSlot (#1144)

ReinforceZwei:
#1144

### DIFF
--- a/MinecraftClient/McClient.cs
+++ b/MinecraftClient/McClient.cs
@@ -1723,10 +1723,23 @@ namespace MinecraftClient
             // Handle cursor item
             if (inventoryID == 255 && slotID == -1)
             {
-                if (item != null)
-                    inventories[0].Items[-1] = item;
-                else
-                    inventories[0].Items.Remove(-1);
+                if (inventories.ContainsKey(0))
+                {
+                    if (item != null)
+                        inventories[0].Items[-1] = item;
+                    else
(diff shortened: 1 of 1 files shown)


MinecraftClient/McClient.cs
```diff
@@ -1723,10 +1723,23 @@ public void OnSetSlot(byte inventoryID, short slotID, Item item)
...
-                    inventories[0].Items.Remove(-1);
+                if (inventories.ContainsKey(0))
+                {
+                    if (item != null)
+                        inventories[0].Items[-1] = item;
+                    else
+                        inventories[0].Items.Remove(-1);
+                }
+            }
+            else if (inventoryID == 254) // To add item to player inventory without animation
```
ORelio, line 1734: To limit copy and paste, how about doing this at the beginning of function:
```C#
if (inventoryID == 254)
    inventoryID = 0;
```
Handling for inventory ID 254 is slightly different than other inventories, not sure if that's intentional.

ReinforceZwei: Done.

ReinforceZwei: > Handling for inventory ID 254 is slightly different than other inventories, not sure if that's intentional.

It should be the same except no item adding animation.

ORelio: Ok, since we do not animate this... it's the same 😄

PR #1220, opened 2020-08-18
OnDeath event

Nekiplay:


ORelio:
So simple I'm wondering why it wasn't there before 👍

PR #1222, opened 2020-08-18
Add Entity.Item and OnEntityMetadata event and More

Nekiplay:
### Done
- [x] Entity.Item
- [x] EntityPose enum
- [x] Entity.Pose
- [x] Entity.Latency
- [x] Entity.Metadata
- [x] OnEntityMetadata event
- [x] Cmd: /entity upgrade
- [x] Pull request

### Tested
- [x] 1.15.2

image

### DIFF
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -342,7 +342,7 @@ namespace MinecraftClient
         /// </summary>
         /// <param name="entity">Entity</param>
         /// <param name="metadata">The metadata of the entity</param>
-        public virtual void OnEntityMetadata(Entity entity, Dictionary<int, object> metadata) { }
+        public virtual void OnEntityMetadata(int entityID, Dictionary<int, object> metadata) { }
 
         /* =================================================================== */
         /*  ToolBox - Methods below might be useful while creating your bot.   */
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -321,11 +321,6 @@ namespace MinecraftClient
         public virtual void OnPlayerLeave(Guid uuid, string name) { }
         
         /// <summary>
-        /// Called when the player deaths
-        /// </summary>
-        public virtual void OnDeath() { }
-        
-        /// <summary>
         /// Called when the player respawns
         /// </summary>
         public virtual void OnRespawn() { }
@@ -343,7 +338,7 @@ namespace MinecraftClient
         /// <param name="entity">Entity</param>
         /// <param name="metadata">The metadata of the entity</param>
         /// <param name="protocolversion">Ptotocol version</param>
-        public virtual void OnEntityMetadata(Entity entity, Dictionary<int, object> metadata, int protocolversion) { }
+        public virtual void OnEntityMetadata(Entity entity, Dictionary<int, object> metadata) { }
 
         /* =================================================================== */
         /*  ToolBox - Methods below might be useful while creating your bot.   */
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -342,7 +342,8 @@ namespace MinecraftClient
         /// </summary>
         /// <param name="entity">Entity</param>
         /// <param name="metadata">The metadata of the entity</param>
-        public virtual void OnEntityMetadata(Entity entity, Dictionary<int, object> metadata) { }
+        /// <param name="protocolversion">Ptotocol version</param>
+        public virtual void OnEntityMetadata(Entity entity, Dictionary<int, object> metadata, int protocolversion) { }
 
         /* =================================================================== */
         /*  ToolBox - Methods below might be useful while creating your bot.   */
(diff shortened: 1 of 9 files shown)


Nekiplay:
image
Strangely, for some reason it gives an error on the iron ingot.

Nekiplay (commented):
Done

MinecraftClient/ChatBot.cs
```diff
@@ -320,6 +320,11 @@ internal void ProcessQueuedText()
         /// <param name="name">Name of the player</param>
         public virtual void OnPlayerLeave(Guid uuid, string name) { }
         
+        /// <summary>
+        /// Called when the player deaths
+        /// </summary>
+        public virtual void OnDeath() { }
```
ORelio, line 326: I think this event is already present as per #1220. Could you remove it from the PR?

MinecraftClient/ChatBot.cs
```diff
@@ -328,9 +333,16 @@ internal void ProcessQueuedText()
...
         /// <param name="health">The health of the entity</param>
-        public virtual void OnEntityHealth(int entityID, float health) { }
+        public virtual void OnEntityHealth(Entity entity, float health) { }
+
+        /// <summary>
+        /// Called when the metadata of an entity changed
+        /// </summary>
+        /// <param name="entity">Entity</param>
+        /// <param name="metadata">The metadata of the entity</param>
+        public virtual void OnEntityMetadata(int entityID, Dictionary<int, object> metadata) { }
```
ORelio, line 345: Should be `OnEntityMetadata(Entity entity [...]` for consistency with other APIs

MinecraftClient/ChatBot.cs
```diff
@@ -328,9 +333,16 @@ internal void ProcessQueuedText()
         /// <summary>
         /// Called when the health of an entity changed
         /// </summary>
-        /// <param name="entityID">Entity ID</param>
+        /// <param name="entity">Entity</param>
         /// <param name="health">The health of the entity</param>
-        public virtual void OnEntityHealth(int entityID, float health) { }
+        public virtual void OnEntityHealth(Entity entity, float health) { }
```
ORelio, line 338: Good idea 👍

MinecraftClient/Mapping/EntityPose.cs
```diff
@@ -0,0 +1,13 @@
+namespace MinecraftClient.Mapping
+{
+    public enum EntityPose
+    {
+        Standing = 0,
+        Fall_Flying = 1,
```
ORelio, line 6: Case should be `FallFlying` (same for `SpinAttack` below)

MinecraftClient/McClient.cs
```diff
@@ -2027,6 +2028,7 @@ public void OnUpdateHealth(float health, int food)
                 {
                     ConsoleIO.WriteLogLine("You are dead. Type /respawn to respawn.");
                 }
+                DispatchBotEvent(bot => bot.OnDeath());
```
ORelio, line 2031: Already present since #1220, should be removed from this PR

MinecraftClient/McClient.cs
```diff
@@ -2066,8 +2068,17 @@ public void OnLatencyUpdate(Guid uuid, int latency)
             string playerName = null;
             if (onlinePlayers.ContainsKey(uuid))
             {
+                Dictionary<int, Entity> keyValuePairs = GetEntities();
```
ORelio, line 2071: (1/2) No need to call `GetEntities()` here, you can access `entities` directly since it's an internal field

MinecraftClient/McClient.cs
```diff
@@ -2066,8 +2068,17 @@ public void OnLatencyUpdate(Guid uuid, int latency)
             string playerName = null;
             if (onlinePlayers.ContainsKey(uuid))
             {
+                Dictionary<int, Entity> keyValuePairs = GetEntities();
                 playerName = onlinePlayers[uuid];
                 playersLatency[playerName] = latency;
+                foreach (KeyValuePair<int, Entity> ent in keyValuePairs)
```
ORelio, line 2074: (2/2)... so here the loop becomes `foreach (KeyValuePair<int, Entity> ent in entities)`

MinecraftClient/McClient.cs
```diff
@@ -2137,17 +2148,52 @@ public void OnUpdateScore(string entityname, byte action, string objectivename,
...
         {
             if (entities.ContainsKey(entityID))
             {
-                entities[entityID].Health = health;
-                DispatchBotEvent(bot => bot.OnEntityHealth(entityID, health));
+                Entity entity = entities[entityID];
+                try
+                {
+                    entity.Metadata = metadata;
+                    int healthField = protocolversion >= 477 ? 8 : 7; // Health is field no. 7 in 1.10+ and 8 in 1.14+
```
ORelio, line 2163: As discussed in #1205 the protocol handler should handle this, protocol versions should not be necessary in McClient.cs, hence the OnEntityHealth() event. ChatBots will likely not be able to exploit the event due to fields being different between versions: the `Dictionary<int, object> metadata` is dependant on protocol versions so it might not be very helpful as is. The only things I can think of to maintiain an `OnEntityMetadata` would be to make an `EntityMetadata` class with fields that will never change between versions and populate it from the protocol handler, or pass an Entity field with metadata already set by the protocol handler 🤔

ORelio (changes requested):
Thanks for your contribution! There are some things that would need to be addressed before merging:

ORelio (approved):
Manually fixed unadressed issues.

PR #1223, opened 2020-08-20
Update translation file indexes web URL to the latest

ReinforceZwei:
this fixed the below issue
image
image

**User need to delete their old lang file first in order to auto-download the newest lang file**

### DIFF
--- a/MinecraftClient/Settings.cs
+++ b/MinecraftClient/Settings.cs
@@ -73,7 +73,7 @@ namespace MinecraftClient
         }
 
         //Other Settings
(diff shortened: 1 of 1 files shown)


MinecraftClient/Settings.cs
```diff
@@ -74,7 +74,7 @@ public static byte MCSettings_Skin_All
 
         //Other Settings
         public static string TranslationsFile_FromMCDir = Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData) + @"\.minecraft\assets\objects\ed\eda1518b15c711cf6e75d99003bd87753f67fac4"; //MC 1.10 en_GB.lang
```
ORelio, line 76: How about updating this one as well? :)

ReinforceZwei: Not sure how. I will take a look at it.

ReinforceZwei:
Done.

PR #1224, opened 2020-08-20
Add Respawn Player

Nekiplay:
### Done
- [x] Respawn player (Do not call this method often, it may affect the CPU)
- [x] Pull request


ORelio:
What is the purpose or sending a respawn packet when you are not dead?

Nekiplay:
> What is the purpose or sending a respawn packet when you are not dead?


I think it would be useful in chat bots, after the OnDeath event, the user would use the condition to check whether or not the bot should be resurrected.

PR #1230, opened 2020-08-23
Update hostile mobs list

### LINKED ISSUE #1229: Why the AutoAttack won't attack Enderman
Why the AutoAttack won't attack Enderman

ReinforceZwei:
Fix #1229

### DIFF
--- a/MinecraftClient/Mapping/EntityTypeExtensions.cs
+++ b/MinecraftClient/Mapping/EntityTypeExtensions.cs
@@ -17,16 +17,21 @@ namespace MinecraftClient.Mapping
             switch (e)
             {
                 case EntityType.Blaze:
+                case EntityType.CaveSpider:
                 case EntityType.Creeper:
                 case EntityType.Drowned:
+                case EntityType.Enderman:
+                case EntityType.Endermite:
                 case EntityType.Evoker:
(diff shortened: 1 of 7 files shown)


MinecraftClient/Mapping/EntityTypeExtensions.cs
```diff
@@ -41,7 +46,9 @@ public static bool IsHostile(this EntityType e)
                 case EntityType.WitherSkeleton:
                 case EntityType.Zoglin:
                 case EntityType.Zombie:
+                case EntityType.ZombiePigman:
```
ORelio, line 49: Is there a difference between `ZombiePigman` and `ZombifiedPiglin` or did the name simply change in 1.6?
If so, it might be interesting to remove `ZombiePigman` from `EntityTypes` and use `ZombifiedPiglin` instead.
Does it make sense to you? Or should we keep both?

ReinforceZwei: They are the same mob but the name got changed. Let's just convert the old name to the new one.

ORelio: OK, doing this.

PR #1233, opened 2020-08-24
Fix unknown item id cause crashes

### LINKED ISSUE #1232: Current version of MCC can't handle modded item
the older version can handle, now it can't
image

ReinforceZwei:
Fix #1232
For handling non-vanilla item (i.e. forge item)


ORelio:
Good idea, added `ItemType.Unknown` for that.
Converting unknown items back to IDs will of course not work so inventory actions for unknown items will not work.

PR #1236, opened 2020-08-25
Packet palette

ReinforceZwei:
The concept is that the packet IDs are consecutive numbers. We can put the packet type into a list in a correct order, then we can access the list by index and it will return the correct packet type. Moreover, we can create the reverse mapping by looping the list and create both `getById` and `getByType` method.

This covered all packet (in and out).

Tested on 1.9, 1.10, 1,11.2, 1.12.2, 1.13.2, 1.14.4, 1.15.2 and 1.16.2 server.


MinecraftClient/Protocol/Handlers/PacketTypes/Palettes/PacketTypePalette.cs
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        /// <summary>
+        /// Dump the inbound packet ID mapping to console
+        /// </summary>
+        public void DumpInboundPacketId()
+        {
+            for (int i = 0; i < GetListIn().Count; i++)
+            {
+                ConsoleIO.WriteLine("0x" + i.ToString("X2") + " " + GetListIn()[i]);
```
ORelio, line 76: Good idea to put in hexadecimal 👍

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -60,6 +62,7 @@ class Protocol18Handler : IMinecraftCom
         IMinecraftComHandler handler;
         EntityPalette entityPalette;
         ItemPalette itemPalette;
+        PacketTypePalette PacketType;
```
ORelio, line 65: Maybe name the variable `packetPalette` like the others?

MinecraftClient/Protocol/Handlers/PacketTypes/PacketTypeHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+        public PacketTypePalette GetTypeHandler()
+        {
+            return GetTypeHandler(this.protocol);
+        }
+        /// <summary>
+        /// Get the packet type palette
+        /// </summary>
+        /// <param name="protocol">Protocol version to use</param>
+        /// <returns></returns>
+        public PacketTypePalette GetTypeHandler(int protocol)
```
ORelio, line 40: Other palette selections are in `Protocol18.cs`. It's fine to put it here but at least throw an exception as with other palette selections when trying to initialize the palette for a new Minecraft version that is not implemented yet. This helps people trying to implement new Minecraft versions by immediately telling them they forgot the palette 😄

Also, how did you generate the palettes? People implementing newer versions will need guidance for this, so some comments would be useful here to explain how to generate new palettes. (And replace the line `If packet IDs were changed, also update getPacketIncomingType() and getPacketOutgoingID() inside Protocol18PacketTypes.cs` in `Protocol18.cs` to point to the new instructions)

MinecraftClient/MinecraftClient.csproj
```diff
@@ -154,6 +154,19 @@
     <Compile Include="Protocol\Handlers\Forge\FMLVersion.cs" />
     <Compile Include="Protocol\Handlers\PacketIncomingType.cs" />
     <Compile Include="Protocol\Handlers\PacketOutgoingType.cs" />
+    <Compile Include="Protocol\Handlers\PacketTypes\Palettes\PacketPalette111.cs" />
```
ORelio, line 157: To avoid too many directories and for consistency, it would be better to use the same architecture as other namespaces:

```
MinecraftClient.Protocol.Handlers
 +- PacketTypesIn.cs
 +- PacketTypesOut.cs
 +- PacketTypeHandler.cs (or replace Protocol18PacketTypes.cs with it)
MinecraftClient.Protocol.Handlers.PacketPalettes
 +- PacketPalette111.cs
 +- [...]
```

MinecraftClient/MinecraftClient.csproj
```diff
@@ -327,6 +340,7 @@
     <Content Include="Protocol\Dns\Records\totla.txt" />
     <Content Include="Resources\AppIcon.ico" />
   </ItemGroup>
+  <ItemGroup />
```
ORelio, line 343: Seems like your Visual Studio added some useless empty node there.

ORelio (changes requested):
Thanks for working in packet palettes! This will definitely help in implementing packets for both older versions (all IDs defined even for packets not implemented yet) and newer versions (maybe you found a way to automate palette generation?)

Some remarks:

MinecraftClient/Protocol/Handlers/PacketTypes/Palettes/PacketPalette111.cs
```diff
@@ -0,0 +1,137 @@
...
+namespace MinecraftClient.Protocol.Handlers.PacketTypes.Palettes
+{
+    /// <summary>
+    /// For Minecraft version 1.10 - 1.11.2
+    /// </summary>
+    class PacketPalette111 : PacketTypePalette
+    {
+        private List<PacketTypesIn> typeIn = new List<PacketTypesIn>()
+        {
+            PacketTypesIn.SpawnEntity,
```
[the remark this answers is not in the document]
ReinforceZwei: Currently those palette are created by hand. I think it is painful to hardcode every packet ID into the code so I decided to use index for replacement. But without the ID clearly stated it might make more mistakes is a fact. So I added a dumper method to dump out the packet ID for debugging.

MinecraftClient/MinecraftClient.csproj
ReinforceZwei: Then I am going to delete the old implementation of packet type.

MinecraftClient/Protocol/Handlers/PacketTypes/Palettes/PacketPalette111.cs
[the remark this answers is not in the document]
ReinforceZwei: Seems like it is easy to generate a packet palette with ID in the source code after creating the packet type list. Just need to write a script for doing that.
i.e. from List
```cs
{
    PacketTypesIn.SpawnEntity,
    ...
}
```
to Dict with ID
```cs
{
    { 0x00, PacketTypesIn.SpawnEntity },
    ...
}
```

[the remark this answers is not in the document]
ReinforceZwei: I wrote a method to convert the existing packet List to the packet Dictionary with IDs hard-coded. New Dict will look like this:
image
If this is okay, I will convert all palette into this format.

ReinforceZwei:
All requests done. The document might be a bit difficult to understand because my English is poor. Please correct me if there is any mistake or room of improvement.

ORelio (approved):

ORelio:
Looks good! Regarding 1.16.2 support... well, lately the project gained some very invested contributors who worked hard to implement new features. 6 months ago, I would not have expected MCC to be able to handle inventories, entities, and build stuff, but see all the work done now 🥇 👏

PR #1243, opened 2020-08-29
Create item moving helper class

ReinforceZwei:
From #1208

Also found a bug (i guess) while testing new class. Player inventory isn't available after the `OnAfterGameJoined` event.

### DIFF
--- a/MinecraftClient/Inventory/ItemMovingHelper.cs
+++ b/MinecraftClient/Inventory/ItemMovingHelper.cs
@@ -1,0 +1,187 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace MinecraftClient.Inventory
+{
+    /// <summary>
+    /// Class that contains useful methods to move item around in a container
+    /// </summary>
+    public class ItemMovingHelper
+    {
+        private Container c;
+        private Func<int, int, WindowActionType, bool> windowAction;
+
+        /// <summary>
+        /// Create a helper that contains useful methods to move item around in container
+        /// </summary>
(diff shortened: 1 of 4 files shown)


MinecraftClient/Inventory/ItemMovingHelper.cs
```diff
@@ -0,0 +1,187 @@
...
+
+namespace MinecraftClient.Inventory
+{
+    /// <summary>
+    /// Class that contains useful methods to move item around in a container
+    /// </summary>
+    public class ItemMovingHelper
+    {
+        private Container c;
+        private Func<int, int, WindowActionType, bool> windowAction;
```
ORelio, line 15: Any reason for storing a callback to windowAction here and not just a reference to McClient itself? 🤔

ReinforceZwei: Yes. Since ChatBot cannot access to McClient

ORelio: Oh, OK. In that case, how adding this to the ChatBot API just after the `WindowAction` API method?
```C#
        /// <summary>
        /// Get inventory action helper
        /// </summary>
        /// <param name="container">Inventory Container</param>
        /// <returns>ItemMovingHelper instance</returns>
        protected ItemMovingHelper GetItemMovingHelper(Container container)
        {
            return new ItemMovingHelper(Handler, container);
        }
```
Would make the ItemMovingHelper more visible and easy to use for people using the ChatBot API.

ReinforceZwei: Okay.

ORelio (approved):

Issue #1249, opened 2020-09-02
Proxy Switcher


Nekiplay:
**Prerequisites**
 - [x] I have read and understood the [user manual](https://github.com/ORelio/Minecraft-Console-Client/tree/master/MinecraftClient/config)
 - [x] I made sure I am running the latest [development build](https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/artifacts)
 - [x] I tried to [look for similar feature requests](https://github.com/ORelio/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

I think it would be useful to have such a function that itself changes the Proxy if it becomes unavailable, and in order for the proxy to be taken from txt, it will also have to remember the last good Proxy found from txt

ORelio:
I'm not very fond of that kind of addition (it has been asked before already), because it will very likely be abused. Some people already abuse MCC to spam chat on servers, but MCC won't automatically reconnect if a server admin puts an IP ban or tempban. Adding a random proxy or proxy switching feature would make it super easy to bypass IP bans and keep spamming 😕

Nekiplay:
> I'm not very fond of that kind of addition (it has been asked before already), because it will very likely be abused. Some people already abuse MCC to spam chat on servers, but MCC won't automatically reconnect if a server admin puts an IP ban or tempban. Adding a random proxy or proxy switching feature would make it super easy to bypass IP bans and keep spamming 😕

Spamery eto konechno plokho, no po suti MCC eto otkrytyy proyekt i pol'zovateli ispol'zuyut yego tak kak schitayut nuzhnym, dannaya funktsiya byla-by polezna, i na drugikh pol'zovateley by ona vryadli by povliyala tak-kak v osnovnom krupnyye prokuty pokupayut dlya vyzhnykh tselets proksi.
Spammers are certainly bad, but in fact, MCC is an open project and users use it as they see fit, this function would be useful, and it would hardly affect other users, since mostly large proxies buy proxies for the most effective targets.

Mhowser:
I'm in favor of this, mostly because this can help with taking down Pay2Win
servers that violate Mojang's EULA.

OverHash:
> but in fact MCC is an open project and users use it as they see fit
This is not a valid reason to add this feature into the official build.

I can't think of any good reasons as to why this feature needs to be implemented. Why would a proxy suddenly become unavailable?

ReinforceZwei:
[...]

PR #1250, opened 2020-09-08
ProxySwitcher v1

Nekiplay:
#1249 
Example:
image


ORelio:
Hi,
I'm sorry, especially since you spent some time working on it, but as stated in #1249 I cannot merge features that may backfire (i.e. where the potential negative consequences outweigh the added convenience). However, I think you can easily wrap this up into a ChatBot for your personal use using the `OnDisconnect` event.

Issue #1263, opened 2020-09-18
Please enable Terrain and Movements to use this command.


LilSadPanda:
Okay so its my first time sending a bug request. I double checked everything in the MinecraftClient.ini file and also enabled **terrainandmovements=true** and when i want to use /dig or /move it tells me to enable **terrainandmovements=true** but it is enabled

ReinforceZwei:
What Minecraft version are you using? Terrain and movement do not support version 1.16+

ORelio:
This issue is likely a duplicate of #1090

mxamber:
Same issue here, I had to turn on debug messages to learn that 1.16 doesn't support terrain handling. Would be helpful if the program returned different errors for 'terrainandmovements=false' and 'this version is not supported' tho, the way it is now, it's a wee bit misleading.

ReinforceZwei:
MCC will inform the user that the terrain and movements is not supported even though without debug message is on.
![image](https://user-images.githubusercontent.com/39955851/93843652-222b9c00-fccd-11ea-9f40-c246b4543a4b.png)
But indeed the message isn't eye-catching and user may not see it. It has rooms of improvement.

PR #1264, opened 2020-09-22
Change color of important log to red

ReinforceZwei:
From #1263 
Result:
image


ORelio:
Thanks 👍

I merged #1265 instead although posted later than yours because it also made the change to `Protocol16.cs`.

PR #1265, opened 2020-09-22
More transparent notifications about unsupported features

mxamber:
Changed the "Terrain & Movements/Inventories/Entities are currently not handled for that MC version" message display upon startup from dark gray to red to more obviously point out the issue.

image


ReinforceZwei:
Hi, there is a same pull request #1264 with exactly same changes already exist.

ORelio:
I'll merge this one because it also changes the lines in `Protocol16.cs`.

PR #1307, opened 2020-11-02
Fix interact entity not being correctly implemented 1.16+

### LINKED ISSUE #1304: Crash when using InteractEntity(entity.ID, 0, 0);
**Prerequisites**
 - [x] I tried to [look for similar issues](https://github.com/ORelio/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one
 - [x] I have set `debugmessages=true` in config to diagnose my issue
 - [x] I have redacted session tokens and passwords before attaching screenshots

**Bug description**
A bug report should contain the following items:
* A clear and concise description of what the bug is
* Steps to reproduce the behavior
* What I expected to happen instead
* If applicable, add screenshots to help explain your problem
* Specify which kind of operating system you are running

**Description**
Crash when using InteractEntity(entity.ID, 0, 0); in script.

-add ''BotLoad(new AutoTrade());'' in McClient.cs at line 199
- save this code in AutoTrade.cs in ChatBots folder
```
using MinecraftClient.Inventory;
using MinecraftClient.Mapping;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;

//MCCScript 1.0
//MCC.LoadBot(new AutoTrade());
//MCCScript Extensions

namespace MinecraftClient
{
    public class AutoTrade : ChatBot
    {
        private bool sent = false;
        public override void OnEntityMove(Entity entity)
        {
            if (entity.Type == EntityType.Villager)
            {
                if (!sent)
                {
                    LogToConsole("1");
                    //InteractEntity(entity.ID, 2, 0, 0.3F, 1.4F, -0.01F);   (actually what the client sends before a normal interact but not supported yet)
                    InteractEntity(entity.ID, 0, 0);
                    LogToConsole("2");
                    SendAnimation(Inventory.Hand.MainHand);
                    //InteractEntity(entity.ID, 1); (attacking works so it's selecting the correct entity?)
                    LogToConsole("3");
                    sent = true;
                }

            }
        }
    }
}
```
- compile and enable entity handling and maybe also inventory handling.
- place bot/player next to a villager with a workstation and wait a second for the villager to move.

I expect to happen that the villager inventory is opened and the server sends back some packets.
console:
image
not that much to see sadly
server console:
image

Running windows 10 tried it from within visual studio 2019 and with building it as described in Readme.
My best guess is that the packet sent to the server isn't correctly formatted. 

(also there is the option to use InteractEnitity with a 2 for interact at but in McClient.cs and ChatBot.cs there is no option to pass the x y z for that while there is a handler for it in IMinecraftCom.cs. When a real client interacts with a villager it sends the interact at and directly after that a normal interact (implementing this does not fix the crash tho))

ReinforceZwei:
Fix #1304

### DIFF
--- a/MinecraftClient/Protocol/Handlers/Protocol18.cs
+++ b/MinecraftClient/Protocol/Handlers/Protocol18.cs
@@ -1573,7 +1573,7 @@ namespace MinecraftClient.Protocol.Handlers
                 // Currently hardcoded to false
                 // TODO: Update to reflect the real player state
                 if (protocolversion >= MC116Version)
-                fields.AddRange(dataTypes.GetBool(false));
+                    fields.AddRange(dataTypes.GetBool(false));
 
                 SendPacket(PacketTypesOut.InteractEntity, fields);
                 return true;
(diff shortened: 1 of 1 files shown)


ORelio:
Thanks for the fix! 👍

PR #1298, opened 2020-10-29
Fix null reference when trying to get protocol version

ReinforceZwei:
Fix #1297

Issue #1296, opened 2020-10-26
Want to be able to trade with villagers by chatbot


mexiscool:
**Prerequisites**
 - [x] I made sure I am running the latest [development build](https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/artifacts)
 - [x] I have read and understood the [user manual](https://github.com/ORelio/Minecraft-Console-Client/tree/master/MinecraftClient/config)
 - [x] I tried to [look for similar feature requests](https://github.com/ORelio/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

**Feature description**
A feature request should contain the following items:
* A concise description of your problem (I'm always frustrated when [...])
* A clear description of what you want to happen (I'd like the app to do this to solve my problem [...])
* Any alternative solutions or features you've considered.

**Description**
- Want to be able to trade with villagers by chatbot (in 1.16.3).
- Want to be able to access the information contained in the packet TradeList (0x26) sent by the server after sending use/InteractEntity  (0x0e). Also being able to send the SelectTrade (0x22) packet.
here are two screenshots of what's in the packets for opening a farmer villager and selecting the pumpkin trade: 
TradeList received from the server:
![trade_list](https://user-images.githubusercontent.com/27775925/97184550-564c1e00-179f-11eb-8e88-5185cda622d6.PNG)
SelectTrade sent to the server:
![select_trade](https://user-images.githubusercontent.com/27775925/97184576-5c41ff00-179f-11eb-9085-f1b8a1433b1e.PNG)
(used [pakkit](https://github.com/Heath123/pakkit) to analyze this)
the server also sends various other packets but I believe those are already supported.
From what I understand these need to be added in MinecraftClient/Protocol/Handlers/Protocol18.cs but I'm new to c# so I can only kinda understand what's already there.
- For receiving the TradeList I have considered using the OnNetworkPacket in the chatbot API and translating it.
I don't think there is a possible way to send a SelectTrade packet with the api?

mexiscool:
Update got handling the tradelist packet implemented and working. Gonna do the selectTrade one next and try and make a pull request after that.

ORelio:
Would be great! Thanks for investigating that 👍

PR #1316, opened 2020-11-06
Adds support for villager trading

mexiscool:
Adds OnTradeList, Adds SelectTrade as described in #1296 ,
I would love some feedback and suggested improvements, I'm new to c# and actually making pull requests. There was an error in vs that has to do with pull request #1298 I think, I don't really understand why it shows up, and everything still works. 

For example to get the item type of the first trade you would do in OnTradeList, trades[0].inputItem1.Type. And to select a trade SelectTrade(2).


MinecraftClient/ChatBot.cs
```diff
@@ -220,7 +220,19 @@ private McClient Handler
...
+        /// Called when tradeList is received from server
+        /// </summary>
+        /// <param name="windowID">Window ID</param>
+        /// <param name="size">Total Amount of trades the villager has.</param>
+        /// <param name="trades">List of trades.</param>
+        /// <param name="villagerLevel">The level the villager is.</param>
+        /// <param name="experience">The amount of experience the villager has.</param>
+        /// <param name="isRegularVillager">True if regular villagers and false if the wandering trader.</param>
+        /// <param name="canRestock">If the villager can restock his trades at a workstation, True for regular villagers and false for the wandering trader.</param>
+        public virtual void OnTradeList(int windowID, int size, List<Trade> trades, int villagerLevel, int experience, bool isRegularVillager, bool canRestock) { }
```
ReinforceZwei, line 234: How about removing the `int size` parameter as the size can be obtained from `List<Trade> trades`?

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -430,6 +430,25 @@ internal bool HandlePacket(int packetID, Queue<byte> packetData)
                         }
                         int iconcount = dataTypes.ReadNextVarInt(packetData);
                         handler.OnMapData(mapid, scale, trackingposition, locked, iconcount);
+                        break;
+					case PacketTypesIn.TradeList:
```
ORelio, line 434: MCC is indented with spaces, please replace tabs with spaces here 😉

MinecraftClient/Protocol/Handlers/Protocol16.cs
```diff
@@ -845,5 +845,10 @@ public static bool doPing(string host, int port, ref int protocolversion)
             }
             catch { return false; }
         }
+
+        public bool SelectTrade(int selectedSlot)
+        {
+            return false; //Currently not implemented
```
ORelio, line 851: Maybe don't add `SelectTrade` to the ChatBot API for now if it's not implemented?

MinecraftClient/ChatBot.cs
ORelio: Indeed. Size is not necessary because C# `List` has a `Count` attribute.

Additionally, since villager properties may change in newer Minecraft versions, adjusting the parameters in this ChatBot event would beak existing scripts. To make this event more future-proof, how about moving these to a `VillagerInfo` class? The method would look like this:

```C#
public virtual void OnTradeList(int windowID, List<Trade> trades, VillagerInfo villagerInfo) { }
```
```C#
public class VillagerInfo
{
    public int Level { get; set; }
    public int Experience { get; set; }
    public bool IsRegularVillager { get; set; }
    public bool CanRestock { get; set; }
}
```
This way, new fields can be added without breaking existing scripts, and existing getters can be adjusted if necessary.

ORelio (changes requested):
Thanks for your contribution! Everything is looking good.
Here is some feedback for improving your pull request before merging:

MinecraftClient/Protocol/Handlers/Protocol16.cs
```diff
@@ -845,5 +845,10 @@ public static bool doPing(string host, int port, ref int protocolversion)
             }
             catch { return false; }
         }
+
+        public bool SelectTrade(int selectedSlot)
+        {
+            return false; //MC 1.13+
```
mexiscool, line 851: I suspect there is a better way to avoid even needing this in protocol16 file, I can't figure out what part requires this to be in here?

MinecraftClient/ChatBot.cs
```diff
@@ -220,7 +220,18 @@ private McClient Handler
...
+        /// <summary>
+        /// Called when tradeList is received from server
+        /// </summary>
+        /// <param name="windowID">Window ID</param>
+        /// <param name="trades">List of trades.</param>
+        /// <param name="villagerLevel">The level the villager is.</param>
+        /// <param name="experience">The amount of experience the villager has.</param>
+        /// <param name="isRegularVillager">True if regular villagers and false if the wandering trader.</param>
+        /// <param name="canRestock">If the villager can restock his trades at a workstation, True for regular villagers and false for the wandering trader.</param>
+        public virtual void OnTradeList(int windowID, List<Trade> trades, int villagerLevel, int experience, bool isRegularVillager, bool canRestock) { }
```
mexiscool, line 233:
```
public virtual void OnTradeList(int windowID, List<Trade> trades, VillagerInfo villagerInfo) { }
public class VillagerInfo
{
    public int Level { get; set; }
    public int Experience { get; set; }
    public bool IsRegularVillager { get; set; }
    public bool CanRestock { get; set; }
}
```

I can't figure out how to get the actual value into the set part. What I could do is add a dataType just like trade but that is probably not what you intended right?

MinecraftClient/ChatBot.cs
mexiscool: I can't figure out how to get the actual value into the set part. What I could do is add a dataType just like trade but that is probably not what you intended right?

MinecraftClient/Protocol/Handlers/Protocol16.cs
mexiscool: It's only 1.13+ and implemented in protocol18, how would I make it so that it doesn't throw errors if I don't have it in Protocol16?

MinecraftClient/ChatBot.cs
ORelio: Assuming the class is the one presented above, you can access each field directly:
```C#
VillagerInfo villagerInfo = new VillagerInfo();
villagerInfo.Level = 1;
villagerInfo.Experience = 1;
villagerInfo.IsRegularVillager = true;
villagerInfo.CanRestock = true;
```
Or, if you want, even initialize all fields at once:
```C#
VillagerInfo villagerInfo = new VillagerInfo()
{
    Level = 1,
    Experience = 1,
    IsRegularVillager = true,
    CanRestock = true
};
```

MinecraftClient/Protocol/Handlers/Protocol16.cs
ORelio: My bad, sorry

MinecraftClient/ChatBot.cs
mexiscool: I made a new file for the class I don't know if that's the best way to do it.

ORelio (approved):
Looks good! Thanks for your contribution.

Issue #1090, opened 2020-06-27
1.16(.1) support


Bas950:
Minecraft 1.16 implementation status:

☑️ Basic Chat Support
☑️ Terrain Handling
☑️ Inventories
☑️ Entities

ORelio:
OK so basic chat is now handled thanks to your pull request 👍
Updated your post with check list of features that still need work for 1.16 support.

ORelio:
Entity handling for 1.16 done in #1149

ORelio:
Mojang released 1.16.2 a few days ago. Need to update protocol support for 1.16.2 also.

ORelio:
1.16.2 and inventory support done in #1214 - Only Terrain is missing for full 1.16 support.

JackTYM:
Any updates on ETA of Terrain and Movements?

ReinforceZwei:
I am working on it and I have a good news. I am able to move around correctly on 1.16+
But there is an issue that processing chunk data took too much time and causing the server to kick me for timeout. (some server did kick me, but some didn't)

It would be great if someone could have a look of my code and maybe improve it for me. Here is the commit that contain the code [b1beafd](https://github.com/ReinforceZwei/Minecraft-Console-Client/commit/b1beafd5462f95e299d59dcefeec618b26e176eb)

ReinforceZwei:
Update:
Seems like the issue is not related to chunk reading but reading an array of VarInt. In 1.16.2+, the biomes field in `ChunkData` packet changed from Int to VarInt array. Therefore the time costed is 10-12 times longer than before. But since the data length is variable, I can't "skip" it by reading a fixed amount of data. @ORelio any help would be appreciated :pray:

ORelio:
Thanks @ReinforceZwei for working on this I'll review your PR straight away 👍

ORelio:
Terrain support has been added, so now MCC fully supports Minecraft 1.16 🎉

PR #1353, opened 2020-11-26
1.16+ Terrain and Movement support

ReinforceZwei:
related #1090


ReinforceZwei:
Seems like `MultiBlockChange` packet also got changed but I haven't updated it yet.

ReinforceZwei:
Terrain and movement is fully functional now.

Mhowser:
Excellent job! Thank you for doing this!!

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -413,8 +417,26 @@ internal bool HandlePacket(int packetID, Queue<byte> packetData)
...
+                                {
+                                    if (protocolversion >= MC1162Version)
+                                    {
+                                        for (int i = 0; i < biomesLength; i++)
+                                        {
+                                            // Biomes - 1.16.2 and above
+                                            // Don't use ReadNextVarInt because it cost too much time
+                                            while (true)
+                                            {
+                                                if ((dataTypes.ReadNextByte(packetData) & 0x80) != 128) break;
```
ORelio, line 434: Clever! Seems worth to move into a dedicated method `dataTypes.SkipNextVarInt(packetData);` 👍

MinecraftClient/Protocol/Handlers/Protocol18Terrain.cs
```diff
@@ -79,45 +80,97 @@ public void ProcessChunkColumnData(int chunkX, int chunkZ, ushort chunkMask, ush
 
                         Chunk chunk = new Chunk();
 
-                        if (dataArray.Length > 0)
+                        if (protocolversion >= Protocol18Handler.MC116Version)
```
ORelio, line 83: I see no changes for 1.16 in https://wiki.vg/Chunk_Format#Chunk_Section_structure ; so I'm not 100% sure what is done here specifically for 1.16. If it's a more optimized chunk parsing routine, it should work also below 1.16 so no need to duplicate the code here, right?

ORelio (commented):
Good job in implementing 1.16 chunk data and multi-block changes! I've a question about `ProcessChunkColumnData` before merging:

MinecraftClient/Protocol/Handlers/Protocol18Terrain.cs
ReinforceZwei: No. The document for chunk format is outdated and doesn't work on 1.16.

ORelio: Thanks for pointing that out, everything makes sense 👍
I adjusted the parsing code to use your approach instead and disable the padding part when parsing 1.15 chunk data.

ORelio (approved):

ORelio:
Thanks again for your work in adding terrain support for Minecraft 1.16. Now merged 👍

yougeshlama:
awwwwwwwwwwwwww iam so happy finally its implemented :)

PR #1354, opened 2020-11-28
Update VkMessager add Example

Nekiplay:


MinecraftClient/config/ChatBots/VkMessager.cs
```diff
@@ -28,6 +28,27 @@
...
+        VkLongPoolClient = new VkLongPoolClient(vkToken, botCommunityId, ProcessMsgFromVk);
+        ChatId = chatId;
+    }
+
+    public override void Initialize()
+    {
+        LogToConsole("Bot enabled!");
+    }
+
+    private void ProcessMsgFromVk(string senderId, string peer_id, string text, string conversation_message_id, string id, string event_id)
```
ORelio, line 47: According to the description, the bot forward "messages between Minecraft and VKonrakte chats". Should this method do a `SendText()` call here to forward the message to Minecraft chat?

Also, should the VkMessager ChatBot override the `GetText()` method to capture messages and forward them to the VkLongPoolClient if they "start with dot (".")"?

Nekiplay: I think it should be left to users

ORelio: Ok, added a minimal working example just in case.

PR #1355, opened 2020-11-28
New GUIVersion

Nekiplay:


ORelio:
Thanks for your work but I'm not sure what to do with this pull request:
 - It contains a copy of the `MinecraftClient` project (should be kept separately to avoid copy and paste of the whole code)
 - It contains functionality specific to your server (mc.megacraft.org)
 - Maybe other things but with so many changed files I could not fully review it

Nekiplay:
This pool is an example of how you can make a GUI version

ORelio:
Thanks for your contribution, which is great, but cannot be merged in its current state due to https://github.com/ORelio/Minecraft-Console-Client/pull/1355#issuecomment-735362591.

PR #1417, opened 2021-01-15
Update to MC version 1.16.5

ReinforceZwei:
_Note: According to [Minecraft wiki](https://minecraft.gamepedia.com/Java_Edition_1.16.5), `1.16.4` and `1.16.5` share the same protocol number `754`._


ORelio:
This is perfect. When multiple version share the same protocol number, just use the highest one including in variable names 👍
Although before merging you can remove the comments as they will likely be of no use afterwards.

PR #1422, opened 2021-01-17
Fix unknown packet ID cause crash

### LINKED ISSUE #1419: Unable to connect to 1.7.10 server
I tried to connect to a 1.7.10 server.
When I successfully connected to the server, the program crashed. 
It's the same server and game version as #1199 .But reporting errors is not the same.....

ReinforceZwei:
Fix #1419

_I am doubt about ignoring unknown packet ID is a good idea... since a bad/incorrect packet palette might also cause `IndexOutOfBound`_

### DIFF
--- a/MinecraftClient/Protocol/Handlers/PacketPalettes/PacketTypePalette.cs
+++ b/MinecraftClient/Protocol/Handlers/PacketPalettes/PacketTypePalette.cs
@@ -57,7 +57,12 @@ namespace MinecraftClient.Protocol.Handlers.PacketPalettes
         /// <returns>Packet type</returns>
         public PacketTypesIn GetIncommingTypeById(int packetId)
         {
-            return GetListIn()[packetId];
+            PacketTypesIn p;
+            if (GetListIn().TryGetValue(packetId, out p))
+            {
+                return p;
+            }
+            else return PacketTypesIn.Unknown;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


ORelio:
Mods should not use custom packets because they can use plugin messages. If the packet ID is unknown, maybe the client has gotten out of sync and is using some random bytes as packet ID. However, if you can successfully connect to the server and continue doing things e.g. read chat messages, move around, then maybe the server indeed has custom packets.

ReinforceZwei:
@MCwenka said this PR fixed his problem. Should we merge this into master?

ORelio:
OK so in that case I'd say to ignore unknown packets but only if Forge support is enabled.
Non-Forge servers should not have unknown packets unless we did something wrong when defining the palette 🙂

OverHash:
Perhaps a warn should be thrown in the console when an unknown packet was received?

ReinforceZwei:
> OK so in that case I'd say to ignore unknown packets but only if Forge support is enabled.

Done. Please make a review again, thank you.

ORelio:
> 
> 
> Perhaps a warn should be thrown in the console when an unknown packet was received?

@OverHash Seems like it's the case when `debugmessages` is enabled.

Mhowser:
Will this show the unknown packet in chat? I think it would be quite useful to see the raw packet.

ReinforceZwei:
It will show the packet ID but not the whole packet.

PR #1459, opened 2021-02-05
Dynamic sleep time

ReinforceZwei:
I had this idea in my mind a long time ago. I think it could help some low-end device not to timeout easily under heavy load.


ORelio:
Good idea! This should give a more accurate TPS 👍

I've made some slight adjustments, including a change in the check to avoid calling `Update()` twice if `Update()` takes more than 100 milliseconds and returns `false`.

The PR is ready to merge if you're OK with the change.

PR #1488, opened 2021-03-04
Add lock to chatlog file

### LINKED ISSUE #1484: Sharing violation - Chatlog
Every few hours the bot keeps spamming Errors in the cinsole which say (shortened):
Sharing violation on path [path of chatlog] 
at MCC.Chatbot. ChatLog

In my understanding that would mean that another program tries to use the same file on the same time or something about similar. But I dont think there is so it must have to do with a problen coming from MCC. Do you know what is wrong?
Thanks

ReinforceZwei:
try to resolve #1484

### DIFF
--- a/MinecraftClient/ChatBots/ChatLog.cs
+++ b/MinecraftClient/ChatBots/ChatLog.cs
@@ -19,6 +19,7 @@ namespace MinecraftClient.ChatBots
         private bool savePrivate = true;
         private bool saveInternal = true;
         private string logfile;
+        private object logfileLock = new object();
 
         /// <summary>
         /// This bot saves the messages received in the specified file, with some filters and date/time tagging.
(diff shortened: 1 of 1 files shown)


ORelio:
I think the issue is caused by having multiple MinecraftClient.exe trying to access the same lock file. Unless the user wants to load multiple ChatLog bots into the same MinecraftClient.exe writing to a single log file, having this kind of lock will have no effect.

There is a FileMonitor class for handling concurrency, although it does not have a method for appending text yet:
https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/FileMonitor.cs

ReinforceZwei:
Wouldn't it be the thread-safe issue like the one in web access, executing command cause client to crash?

ORelio:
You can do both internal lock and retry on write just to be safe. Internal lock ensures thread safety within a process and retries ensure proper writing when several processes want to write to the same file.

ORelio:
(although I'm not sure why someone would want to log different accounts to the same log file)

ORelio (approved):
It can't hurt to have a lock file here. OK with this PR 👍

Issue #1472, opened 2021-02-19
Ability to filter messages in chatlog


joey1313:
Hello,

I've been trying to filter both my console and chatlog in a certain way such that certain messages get filtered out. I've been successful in filtering my console messages with regex, but I'm not sure how to do the same with chatlog (even after fiddling with chatformat).

How would I apply the same regex filter that I apply on my console output to my chatlog output?

ReinforceZwei:
`ChatLog` currently doesn't support regex message filtering. You can only filter messages by message type, e.g. private message, internal message, etc.

ReinforceZwei:
With the newer version, you can enable writing console output to log file. Both content would be the same. 

You may copy the settings below to your `MinecraftClient.ini` file `[Logging]` section if you don't want to delete your old `.ini` file.
```ini
[Logging]
logtofile=false                    # Write log messages to file
logfile=console-log-%username%-%serverip%.txt    # Log file name
prependtimestamp=false             # Prepend timestamp to messages in log file
```

PR #1494, opened 2021-03-06
Implement log to file logger

ReinforceZwei:
related: #1472

Other changes:
- Logger moved to it's own namespace
- New logger abstract class


MinecraftClient/MinecraftClient.csproj
```diff
@@ -158,7 +160,7 @@
     <Compile Include="Mapping\EntityPalettes\EntityPaletteGenerator.cs" />
     <Compile Include="Mapping\EntityTypeExtensions.cs" />
     <Compile Include="Mapping\MaterialExtensions.cs" />
-    <Compile Include="MCLogger.cs" />
+    <Compile Include="Logger\FilteredLogger.cs" />
```
ORelio, line 163: For your information when you rename an existing file, it's better to use `git mv` as well so that the diff is visible instead of file creation + file deletion as in this pull request. You can do this this way:
1. Rename file in Visual Studio
2. Exit Visual Studio
3. Rename file back to the old name
4. git mv oldname.cs newname.cs
5. The name change is now applied in Git as well

ORelio (approved):
Looks good to me 👍

PR #1503, opened 2021-03-13
New method for getting looking location

### LINKED ISSUE #1364: Get which block player is looking
I am not sure if this is useful, but I wrote a method to do that. The following code require .Net Framework 4.6+ or .Net Core 1+ so I can't put it into the project now. I will leave the code here and wait until the project is ported to .Net 5. (or someone write a simple Vector3 class for me :P)
```cs
// Designed to work in ChatBot class
public Location LookingAtBlock(double yaw, double pitch)
{
    double rotX = (Math.PI / 180) * yaw;
    double rotY = (Math.PI / 180) * pitch;
    double x = -Math.Cos(rotY) * Math.Sin(rotX);
    double y = -Math.Sin(rotY);
    double z = Math.Cos(rotY) * Math.Cos(rotX);
    Vector3 vector = new Vector3((float)x, (float)y, (float)z);
    for (int i = 0; i < 5; i++)
    {
        Vector3 v = Vector3.Multiply(vector, i);
        Location l = GetCurrentLocation().EyesLocation() + new Location(v.X, v.Y, v.Z);
        l.X = Math.Floor(l.X);
        l.Y = Math.Floor(l.Y);
        l.Z = Math.Floor(l.Z);
        Block b = GetWorld().GetBlock(l);
        if (b.Type != Material.Air)
        {
            return l;
        }
    }
    return new Location();
}
```

@ORelio  Also I have a question about the yaw and pitch in `McClient.cs` Class. I tried to get the yaw and pitch value by using `yaw.GetValueOrDefault(0)` but they seems to be empty all the time, even after the `/look` command was executed. I got zero only. Am I doing something wrong?

ReinforceZwei:
Resolve #1364 

_Please ignore the branch name :P_


MinecraftClient/Mapping/Entity.cs
```diff
@@ -54,6 +54,18 @@ public class Entity
         /// </summary>
         public Location Location;
 
+        /// <summary>
+        /// Entity head yaw
+        /// </summary>
+        /// <remarks>Untested</remarks>
```
ORelio, line 60: Is it still untested? 😜

MinecraftClient/Mapping/Entity.cs
```diff
@@ -54,6 +54,18 @@ public class Entity
...
+        public float Yaw = 0;
+
+        /// <summary>
+        /// Entity head pitch
+        /// </summary>
+        /// <remarks>Untested</remarks>
```
ORelio, line 66: Same here

MinecraftClient/Mapping/Entity.cs
```diff
@@ -95,6 +107,23 @@ public Entity(int ID, EntityType type, Location location)
             this.Item = new Item(ItemType.Air, 0, null);
         }
         /// <summary>
```
ORelio, line 109: Missing empty line between `}` and `/// <summary>`

MinecraftClient/Mapping/Entity.cs
```diff
@@ -95,6 +107,23 @@ public Entity(int ID, EntityType type, Location location)
...
+            this.ID = ID;
+            this.Type = type;
+            this.Location = location;
+            this.Health = 1.0f;
+            this.Equipment = new Dictionary<int, Item>();
+            this.Item = new Item(ItemType.Air, 0, null);
+            this.Yaw = yaw * (1 / 256) * 360; // to angle in 360 degree
+            this.Pitch = pitch * (1 / 256) * 360;
+        }
+        /// <summary>
```
ORelio, line 126: same here

MinecraftClient/Mapping/Vector3.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace MinecraftClient.Mapping
+{
+    /// <summary>
+    /// Simple vector3 class
+    /// </summary>
+    public class Vector3
```
ORelio, line 11: This class looks like a duplicate of [Location](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Mapping/Location.cs) which already has the Add and Multiply properties with `+` and `*` operators.

ORelio (changes requested):
Good Job! Here's some feedback:

MinecraftClient/Mapping/Entity.cs
ReinforceZwei: Yes. I am too tired to test it right now.

ReinforceZwei:
Done.

ORelio (approved):

PR #1504, opened 2021-03-13
Feature: Dockerfile

Tandashi:
**What is included in this Pull Request:**
Docker support using a Dockerfile

**Why a Dockerfile?**
For people who don't want to install all the dependencies on Linux for example but already have docker installed.

**What the Dockerfile does:**
It downloads the latest build version and runs it

To download the latest version it uses the appveyor API and [jq](https://stedolan.github.io/jq/) to parse the json in the `sh` file.

**Future improvements:**
Host the docker image publicly so one doesn't have to build it themselves e.g. on [DockerHub](https://hub.docker.com)

### DIFF
--- a/Docker/start-latest.sh
+++ b/Docker/start-latest.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+cd /opt/data || exit 1
+
+# Download latest version
+curl -L https://ci.appveyor.com/api/buildjobs/51an4dvs204ak6tq/artifacts/MinecraftClient%2Fbin%2FRelease%2FMinecraftClient.exe --output MinecraftClient.exe
+
+# Start Client
+mono MinecraftClient.exe
--- a/Docker/start-latest.sh
+++ b/Docker/start-latest.sh
@@ -2,8 +2,11 @@
 
 cd /opt/data || exit 1
 
+# Get latest Job ID
+JOB_ID=$(curl -L https://ci.appveyor.com/api/projects/ORelio/minecraft-console-client | jq -r .build.jobs[0].jobId)
+
 # Download latest version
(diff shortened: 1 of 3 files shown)


Docker/start-latest.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
ORelio, line 1: This script is run with `bash` by the docker file, maybe use `/bin/bash` here for consistency or run with `sh`?

Docker/start-latest.sh
```diff
@@ -0,0 +1,9 @@
...
+
+cd /opt/data || exit 1
+
+# Download latest version
+curl -L https://ci.appveyor.com/api/buildjobs/51an4dvs204ak6tq/artifacts/MinecraftClient%2Fbin%2FRelease%2FMinecraftClient.exe --output MinecraftClient.exe
```
ORelio, line 6: That `51an4dvs204ak6tq` part is randomized by AppVeyor on every new build so getting the latest release would be slightly more involved (request https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/artifacts first to determine the latest download link)

ORelio:
Hi @Tandashi, thanks for contributing. Adding a way to set up quickly with Docker is a good idea 👍

ORelio (changes requested):
Here are a few remarks about your current code:

Docker/start-latest.sh
Tandashi: Should be fixed. (957dc6e8f882517049be136094b12fd557143da5)

Docker/start-latest.sh
Tandashi: Should be fixed. (957dc6e8f882517049be136094b12fd557143da5)

I use the appveyor API and parse the JSON using `jq` to get the latest Job ID  to curl the artifact.

ORelio (approved):

ORelio:
Looks good now! Merged, thanks again 😉

PR #1506, opened 2021-03-16
New entity status packet event

### LINKED ISSUE #1485: Detect when the player is hit
**Prerequisites**
 - [x] I have read and understood the [user manual](https://github.com/ORelio/Minecraft-Console-Client/tree/master/MinecraftClient/config)
 - [x] I tried to [look for similar feature requests](https://github.com/ORelio/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

---
## Describe your problem
Present a concise description of the problem to be addressed by this feature request. 

---
## Suggest a possible solution
A clear description of what you want to happen (I'd like the app to do this to solve my problem [...])

---
## Screenshot to help explain your request (If applicable)

ReinforceZwei:
resolve #1485

### DIFF
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -336,6 +336,12 @@ namespace MinecraftClient
         public virtual void OnEntityMetadata(Entity entity, Dictionary<int, object> metadata) { }
 
         /// <summary>
+        /// Called when the status of client player have been changed
+        /// </summary>
+        /// <param name="statusId"></param>
+        public virtual void OnPlayerStatus(byte statusId) { }
+
+        /// <summary>
         /// Called when a network packet received or sent
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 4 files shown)


ORelio:
I guess that `byte statusId` might change on future versions?
If so, maybe you should define an enum instead just in case we need to change the values in the future?

ReinforceZwei:
After some comparison between an old version of document and the current version of document, I didn't see any change in `statudId` value and their meaning. So I guess it is less likely they will change on future versions.

Pokechu22:
The values are unlikely to change in the future as (as far as I know) Mojang just has a bunch of hardcoded values, hence the somewhat jank partial re-use of values.  Possibly at some point they'll change it to a more typical registry, but that can be handled if/when it happens.

ORelio:
OK, fine by me. I've no other remarks. If you're done, feel free to merge.

PR #1512, opened 2021-03-26
Add setting for overriding attack speed for autoattack

### LINKED ISSUE #1511: How to change attack speed?
How to change attack speed? Attack Speed is low, sorry my English very bad :(

ReinforceZwei:
Resolve #1511

### DIFF
--- a/MinecraftClient/ChatBots/AutoAttack.cs
+++ b/MinecraftClient/ChatBots/AutoAttack.cs
@@ -16,13 +16,14 @@ namespace MinecraftClient.ChatBots
         private int attackCooldownCounter = 6;
         private Double attackSpeed = 4;
         private Double attackCooldownSecond;
+        private bool overrideAttackSpeed = false;
         private int attackRange = 4;
         private Double serverTPS;
         private float health = 100;
         private bool singleMode = true;
         private bool priorityDistance = true;
 
-        public AutoAttack(string mode, string priority)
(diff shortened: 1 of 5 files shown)


ORelio:
Hi,

How about `cooldownsecond=auto` when not set?
This way the user does not need to set `overrideattackspeed` in config:

```ini
 cooldownsecond=auto                # How long to wait between attacks in seconds, or "auto" to auto-adjust.
```
```C#
case "cooldownsecond":
    if (argValue.ToLower() == "auto")
    {
        AutoAttack_OverrideAttackSpeed = false;
    }
    else
    {
        AutoAttack_OverrideAttackSpeed = true;
        AutoAttack_CooldownSecond = str2int(argValue);
    }
    break;
````

Also, what happens if the user sets `cooldownsecond=0`?
Or a negative value? Would a minimum delay be relevant or not?

ReinforceZwei:
> How about cooldownsecond=auto when not set?

Thanks for the nice idea :D

> Also, what happens if the user sets cooldownsecond=0?

I forgot to handle such case when implementing this last night. Now it will use `auto` if invalid value was given.

ORelio:
I only realize it now, but there is a slight grammar issue in the variable/setting name:
usually you'd write `cooldownseconds` and `attackCooldownSeconds` with as `s`
Apart from that the code is good 😄 👍

ReinforceZwei:
Ok, I will fix it before merging.

PR #1532, opened 2021-04-07
Send chat messages to Discord via webhook.

Daenges:
With corresponding Minecraft skins as profilepicture.


MinecraftClient/config/ChatBots/discordWebhook.cs
```diff
@@ -0,0 +1,369 @@
...
+
+        currentSkinMode = newCurrentSkinMode;
+        fallbackSkin = newFallbackSkin;
+        size = newSize;
+        scale = newScale;
+        overlay = newOverlay;
+    }
+
+    // Always get a an Answer to a valid name!
+    public string getUUIDFromMojang(string name)
```
ReinforceZwei, line 30: Method naming in C# should be `PascalCase`, which the first letter of each word in a compound word is capitalised. The method name would be `GetUUIDFromMojang()`.

MinecraftClient/config/ChatBots/discordWebhook.cs
```diff
@@ -0,0 +1,369 @@
...
+        return playerList.FirstOrDefault(x => x.Value == name).Key.Replace("-", "");
+    }
+
+    public string getSkinURL(string UUID)
+    {
+        string parameters = string.Join("&", "size=" + size, "scale=" + scale, "default=" + fallbackSkin, (overlay ? "overlay" : ""));
+        return string.Format(skinModes[currentSkinMode], UUID + "?" + parameters);
+    }
+
+    public void setSize(int newSize) { if (newSize <= 512) { size = newSize; } }
```
ReinforceZwei, line 48: You may use setter/getter (Auto-Property) here.
```cs
public int Size 
{ 
    set { if (value <= 512) size = value; } 
    get { return size; }
}
```

MinecraftClient/config/ChatBots/discordWebhook.cs
```diff
@@ -0,0 +1,369 @@
...
+    public void setScale(int newScale) { if (newScale <= 10) { scale = newScale; } }
+    public void setCurrentSkinMode(string newSkinMode) { currentSkinMode = newSkinMode; }
```
ReinforceZwei, line 50: Same as above.

MinecraftClient/config/ChatBots/discordWebhook.cs
```diff
@@ -0,0 +1,369 @@
...
+    public void toggleOverlay() { overlay = !overlay; }
+    public void toggleFallbackSkin() { fallbackSkin = (fallbackSkin == "MHF_Steve" ? "MHF_Alex" : "MHF_Steve"); }
+
+    public int getSize() { return size; }
```
ReinforceZwei, line 54: You don't need extra getter method here if you use Auto-Property above.

MinecraftClient/config/ChatBots/discordWebhook.cs
```diff
@@ -0,0 +1,369 @@
...
+        getUUIDFromMojang = true;
+        sendServer = true;
+        togglesending = false;
+        allowMentions = false;
+        onlyPrivate = false;
+    }
+
+    public override void Initialize()
+    {
+        base.Initialize();
```
ReinforceZwei, line 96: You do not have to call `base.Initialize()` as it doesn't do any thing.

MinecraftClient/config/ChatBots/discordWebhook.cs
```diff
@@ -0,0 +1,369 @@
...
+                sendWebhook(username, "[Private Message]: " + message);
+            }
+            else if (sendServer)
+            {
+                sendWebhook("[Server]", text);
+            }
+        }
+    }
+
+    public void setWebhook(string newWebhook) { webhookURL = newWebhook; }
```
ReinforceZwei, line 126: Use getter and setter here as above.

MinecraftClient/config/ChatBots/discordWebhook.cs
```diff
@@ -0,0 +1,369 @@
...
+
+    public void setSkinAPI(string newCurrentSkinType = "flatFace", string newFallbackSkin = "MHF_Steve", int newSize = 512, int newScale = 10, bool newOverlay = true)
```
ReinforceZwei, line 128: Please capitalize the first letter of each method to match `PascalCase` naming style.

ReinforceZwei (changes requested):

Daenges (commented):
Thank you for pointing out some flaws. I fixed most of them. I just had some trouble with the Auto Property. See https://stackoverflow.com/questions/367192/why-does-property-set-throw-stackoverflow-exception

Daenges:
**Things still to do:**

- [ ] Add a message cache to store messages of the same user temporarily to avoid too many requests
- [ ] Clean up the CommandHandler
- [ ] Rename all functions the right way
- [ ] Add an ignore list

Optional Features

- [ ] Integrate https://minotar.net/ API (No need for UUID)
- [ ] Several webhooks at once

ORelio:
Looks cool! That will definitely make many users happy as it was requested more that once in issues 👍

MinecraftClient/config/ChatBots/discordWebhook.cs
ORelio: I'd even say
```C#
set
{
    size = value;
    if (size > 512)
        size = 512;
}
```
So that size changes to max if provided value is too big.

ReinforceZwei:
> Thank you for pointing out some flaws. I fixed most of them. I just had some trouble with the Auto Property. See https://stackoverflow.com/questions/367192/why-does-property-set-throw-stackoverflow-exception

If you do it like this:
```cs
public int size { set { if (value <= 512) { this.size = value; } } }
```
Would be equivalent to:
```cs
public void SetSize(int size)
{
    if (size <= 512)
        SetSize(size);
}
```
and causing stack overflow.


You should have two variables, private and public. 
```cs
private int size;
public int Size {
    get { return size; }
    set {
        size = value;
        if (size > 512)
            size = 512;
    }
}
```

---

For public properties, you should use `PascalCase` naming style.
```cs
public string currentSkinMode
{ ... }
```
Should be
```cs
public string CurrentSkinMode
{ ... }
```

ReinforceZwei:
I removed some codes that were commented out and renamed the file from `discordWebhook.cs` to `DiscordWebhook.cs`.

ReinforceZwei (approved):

Daenges:
Okay, I have now added all features I wanted to have in this script. If you want to look over it feel free to do so. The next commits will most likely be improvements of exsisting stuff. @ReinforceZwei

Daenges:
@ReinforceZwei @ORelio I think I finished working here. Feel free to spot some bugs and merge it.

ReinforceZwei:
Looks good. Merged.

ORelio:
Hi,
Just read the PR and I've a quick remark: `GetUUIDDirectlyFromMojang`: The Mojang API is rate-limited and you are expected to cache the result (e.g. using a Dictionary), or you'll be denied access if you request the same UUID several times in a row. That setting is disabled by default so the bot should work fine with the default settings. God job! 👍

Daenges:
@ORelio Thank you for your compliment, it was a pleasure for me to work on it!
Regarding the mojang API there would actually be a workaround. If you would use the minotar API (mentioned in checklist), you could get around obtaining the UUID directly from mojang, since the API also works with ingame names and presumably checks for the UUID itself. Therefore you would push the potential cache problem to the API host.
But there are a few problems regarding this API:

1. The code would grow significantly if you try to implement both methods at once, hence they use a totally different way of link structure
2. You can only define the total size of the result picture, not scale etc.
3. The API is currently unable to render cubic heads with overlay

In the end both of them would do the same job. Each one of them with its own advantages and flaws. :)

ORelio:
Oh, you don't need to do something overly complicated, just use a dictionary at runtime:
```C#
private Dictionary<string, string> namesToUuidMojangCache = new Dictionary<string, string>();
```
And store only in memory:
```C#
public string GetUUIDFromMojang(string name)
{
    if (namesToUuidMojangCache.ContainsKey(name))
        return namesToUuidMojangCache[name];

    WebClient wc = new WebClient();
    try
    {
        string uuid = Json.ParseJson(wc.DownloadString("https://api.mojang.com/users/profiles/minecraft/" + name)).Properties["id"].StringValue;
        namesToUuidMojangCache[name] = uuid;
        return uuid;
    }
    catch (Exception) { return "00000000000000000000000000000000"; }
}
```
When the bot is relaunched, you can just start over, the API cooldown is not very long (10 minutes).

Daenges:
I got what you suggested there and this is a reasonable edge case. What I suggested was just a way to get around the mojang API completely. I had never planned to implement it by this means in any way.

This is a really rare edgecase. I mean how unlikely is it, that a player writes a message and disconnects faster than the bot could check the playerlist. However, I can make a PR tomorrow to add this to the script!

ORelio:
Oh, I see. Indeed, MCC does not keep the info in player list as soon as the player disconnects.
Could be worth keeping that info in the dictionary using the `OnPlayerJoin` and `OnPlayerLeave` methods of the ChatBot API?

PR #1544, opened 2021-04-14
Improvements Discord Webhook

Daenges:
Improve caching and remove unused functions

### DIFF
--- a/MinecraftClient/config/ChatBots/DiscordWebhook.cs
+++ b/MinecraftClient/config/ChatBots/DiscordWebhook.cs
@@ -22,6 +22,7 @@ class WebhoookSettings
     private Dictionary<string, List<string>> messageCache = new Dictionary<string, List<string>>();
     private Dictionary<string, string> messageContains = new Dictionary<string, string>();
     private Dictionary<string, string> messageFrom = new Dictionary<string, string>();
+    private Dictionary<string, string> namesToUuidMojangCache = new Dictionary<string, string>();
     private List<string> ignoredPlayers = new List<string>();
     #endregion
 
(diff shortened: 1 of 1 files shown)


ORelio:
Oops, did not see you already opened a new PR so I kept posting in #1532. So, as I said, maybe you could get rid of the Mojang API part by registering players in the dictionary inside `OnPlayerJoin`/`OnPlayerLeave` event of the ChatBot API?

Daenges:
I just have some concerns about overloading this Dict.
Lets say, you have a big server, where several hundred players join and leave each day. If you run this bot for several days, the dictionary could have several thousand entries. Do you think that I should implement a clear of this list after a certain time? Problem there is that players who got enlisted one second too early, could be deleted and a new mojang request must be started. My sollution for this would be a list of a new player class which can hold a date additionally to the UUID and the name. Then you could clear all entries older than 10 min. Do you think this is necessary? @ORelio

ORelio:
Well, overestimating the size of one entry to 512 bytes and taking 1000 players * 10 days * 512, that's still only like ~5MB of RAM.

Daenges:
But this will also implies a search through more and more data, just because one Player was not found, which could take some time.

ORelio:
A dictionary is a data structure made to index the values by key and lookup the data immediately regardless of its size.

> Each addition to the dictionary consists of a value and its associated key. Retrieving a value by using its key is very fast, close to O(1), because the Dictionary<TKey,TValue> class is implemented as a hash table.

Source: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2?view=net-5.0#remarks

> In a well-dimensioned hash table, the average cost (number of instructions) for each lookup is independent of the number of elements stored in the table.

Source: https://en.wikipedia.org/wiki/Hash_table

Daenges:
I was actually a bit concerned because of the lookup of .ContainsKey(). But as described below, it is nothing else than a try-catch with the value.
Source: https://stackoverflow.com/questions/9382681/what-is-more-efficient-dictionary-trygetvalue-or-containskeyitem

Okay. So in your opinion I can leave everything as it is?

ORelio:
Good catch, TryGetValue is indeed better because you do only one lookup instead of two.
Yes, just adding mappings in the dictionary in memory seems fine to me since it'll clear when closing the client.
You may also clear the dictionary on connection lost, as servers may restart from time to time you can use this to perform cleanup.

MinecraftClient/config/ChatBots/DiscordWebhook.cs
```diff
@@ -101,6 +101,7 @@ public WebhoookSettings()
     public Dictionary<string, List<string>> GetCachedMessages() { return this.messageCache; }
     public Dictionary<string, string> GetSkinModes() { return this.skinModes; }
     public Dictionary<string, string> GetNamesToUuidMojangCache() { return this.namesToUuidMojangCache; }
+    public void resetUUIDCache() { namesToUuidMojangCache = new Dictionary<string, string>(); }
```
ORelio, line 104:
```suggestion
    public void resetUUIDCache() { namesToUuidMojangCache.Clear(); }
```

ORelio (approved):
If you are done with the changes let me know if I can merge the PR

Daenges:
@ORelio Everything is done here. :)

Issue #1519, opened 2021-03-30
What can cause my console client to freeze when reconnecting?


Tnavillacs:
I always lose my connection at around 2am and I run 30+ alts with console client. I already added the random delay when reconnecting which helped. But sometimes I still have a few alts freezing.

```
02:14:00 [MCC] Connection has been lost.
02:14:00 [MCC] [AutoRelog] Waiting 63 seconds before reconnecting...
Restarting Minecraft Console Client...
02:15:05 Cached session is still valid for XXXXXXX.
02:15:05 Using Minecraft version 1.8.9 (protocol v47)
02:15:05 [MCC] Version is supported.
Logging in...
```
It's just stuck at "Logging in..." and I also can't type anything in the chat. I can only close the window.

ORelio:
I'm not sure either. Try `debugmessages=true` in config

n00b69:
This happens if a server soft crashes (server appears online and you can see online users in the menu screen, however you can't connect etc.) The only way to relog is to manually close and reopen the MCC window. I hope there'll be a fix for this soon.

ORelio:
I'm not sure there is a timeout on server login so I'll convert this to an enhancement request.

ReinforceZwei:
Tcp timeout is set to 30000 before calling `Login()` so there is timeout. It doesn't seems to work though.

ReinforceZwei:
Server side is using `netcat` to listen TCP connection: `nc -l 25565`
Client: 
![image](https://user-images.githubusercontent.com/39955851/116551121-2c301000-a92a-11eb-8957-8aa97c439671.png)
We can see TCP timeout is working if server doesn't send anything.

My question is: How to create a server that can 100% reproduce this issue?

n00b69:
The server I play on automatically force restarts if it doesn't respond for more than 30 seconds. It has a weird crashing problem, apparently caused by old mineshaft data (the map is from 2016).

Aternos servers also seem to soft crash if you paste a schematic that's too big, this could perhaps be used.

ReinforceZwei:
It doesn't seems to be relevant but would setting `TcpClient.SendTimeout` help? @ORelio

ORelio:
I don't think so. We'd probably need to start the timeout detector straight when starting login, not after.

PR #1560, opened 2021-04-30
Add timeout detect on server login

ReinforceZwei:
Solve #1519

Able to detect timeout when connecting to dummy netcat server
image


ORelio:
Good job! 👍

PR #1566, opened 2021-05-03
Remove thread sleep in disconnect method

ReinforceZwei:
_I don't see any reason why it is there :P_
_And It slows down the shutdown speed_


ORelio:
I added that line so long ago that the repo did not exist yet so I have no idea why I put that Thread.Sleep() 😅

ORelio:
Well, we'll see if anything breaks 😜

PR #1570, opened 2021-05-04
Add method to schedule main thread task

ReinforceZwei:
Taking command input as example, scheduling a task to run is better than creating a queue for everything that need to be run on the main thread.

#

My next plan is to create a method to schedule a delayed task for `ChatBot` API so that delaying a task will be easier.
https://github.com/ORelio/Minecraft-Console-Client/blob/0400ad89acf2aa3fad985ce45afbefda2d90db1d/MinecraftClient/ChatBots/AutoAttack.cs#L70-L73
https://github.com/ORelio/Minecraft-Console-Client/blob/0400ad89acf2aa3fad985ce45afbefda2d90db1d/MinecraftClient/ChatBots/AutoFishing.cs#L35-L42

### DIFF
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -61,7 +61,7 @@ namespace MinecraftClient
         /// <remarks>
         /// <see cref="Update"/> method can be overridden by child class so need an extra update method
         /// </remarks>
-        public void UpdateInternal()
+        public sealed void UpdateInternal()
         {
             lock (delayTasksLock)
             {
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -41,6 +41,8 @@ namespace MinecraftClient
         private ChatBot master = null;
         private List<string> registeredPluginChannels = new List<String>();
         private List<string> registeredCommands = new List<string>();
+        private object delayTasksLock = new object();
+        private List<DelayedTask> delayTasks = new List<DelayedTask>();
         private McClient Handler
         {
             get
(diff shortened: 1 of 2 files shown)


ORelio:
That's a good idea as it could be a simple way to make the ChatBot API thread-safe. In case the API is called from a different thread, the call would need to be queued, then block the API call until the task is properly executed on main thread, then retrieve and return the result. The command prompt could also use that feature rather than a queue, indeed.

(Just in case: don't touch the chat queue as it has a rate-limit mechanism)

ORelio:
Not sure if useful, but there is a similar mechanism in Windows Forms. Forms need some actions to be invoked in the main thread managing the UI of the form, with methods `InvokeRequired` and `Invoke`.

https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.invokerequired
https://docs.microsoft.com/fr-fr/dotnet/api/system.windows.forms.control.invoke
https://stackoverflow.com/questions/6650691/invoke-in-windows-forms

ReinforceZwei:
This PR is basically done. Optionally we can document a bit of how to use this method.

ORelio:
How do you wait for task to be executed and retrieve the result from the other thread?

ReinforceZwei:
> How do you wait for task to be executed and retrieve the result from the other thread?

It would be more complex to implement blocking and returning result. I guess I will take a look at that in a new git branch.

ORelio:
OK, why not.

MinecraftClient/ChatBot.cs
```diff
@@ -53,6 +55,39 @@ private McClient Handler
             }
         }
 
+        /// <summary>
+        /// Will be called every ~100ms.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="Update"/> method can be overridden by child class so need an extra update method
+        /// </remarks>
+        public void UpdateInternal()
```
ORelio, line 64: How about protecting this method using the `sealed` attribute?
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/sealed

ORelio (approved):

ReinforceZwei:
Sorry I forgot to mention, the `sealed` attribute seems not to be working and I am about to delete that commit.

ORelio:
Never tried this attribute before. My bad too.

PR #1571, opened 2021-05-05
Implement hardcoding the WebhookURL

Daenges:
This is not a very smooth way of implementing the automatic start of the script with a URL. But as it was requested, I added it. I would rather prefere doing this with a script, which calls the bot and after that executes the "/dw changeurl" command.

A better way of implementation would be to add start parameters like "/script discordwebhook WebhookURL", which automatically starts the script with the URL. Unfortunately I could not find a way to read the arguments provided while calling the script.

### DIFF
--- a/MinecraftClient/config/ChatBots/DiscordWebhook.cs
+++ b/MinecraftClient/config/ChatBots/DiscordWebhook.cs
@@ -78,7 +78,7 @@ class WebhoookSettings
         skinModes.Add("fullSkin", "https://crafatar.com/renders/body/{0}");
 
         // Define standard values for main class
-        WebhookURL = "Your URL";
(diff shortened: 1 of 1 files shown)


ReinforceZwei:
> I would rather prefere doing this with a script, which calls the bot and after that executes the "/dw changeurl" command.

You may want to add a short description at the top of the script to give some example for users as they might not know how to do it exactly.

ORelio:
Could be nice to put the expected format instead of "Your URL" e.g. "https://domain.com/endpoint/YOURTOKENHERE"

Daenges:
So something like "https://discord.com/api/webhooks/SomeNumber/RandomStuff"?

ORelio:
Yes. I don't how how the URL is supposed to look it it should help people configuring the bot if they know how it looks like 😄

Daenges:
@ORelio opened a new PR ;D

Issue #1362, opened 2020-11-30
TheAltening Support


HeyItskPan1c:
I was curious if you would be able to add rhealtening support for premium altering users

Mhowser:
There is no point as mojang is forcing everyone to migrate to a microsoft
account in the future.

ORelio:
We do not add support for services offering stolen/for-rent accounts are they are too often misused.

Tiebe:
I added MCLeaks support (similar to TheAltening) #1577

PR #1577, opened 2021-05-08
Add MCLeaks and TheAltening support

Tiebe:
This adds MCLeaks support, this pr is probably not going to be accepted but I'm just putting it out there!

### DIFF
--- a/MinecraftClient/Protocol/ProtocolHandler.cs
+++ b/MinecraftClient/Protocol/ProtocolHandler.cs
@@ -370,47 +370,79 @@ namespace MinecraftClient.Protocol
             try
             {
                 string result = "";
-                string json_request = "{\"agent\": { \"name\": \"Minecraft\", \"version\": 1 }, \"username\": \"" + JsonEncode(user) + "\", \"password\": \"" + JsonEncode(pass) + "\", \"clientToken\": \"" + JsonEncode(session.ClientID) + "\" }";
-                int code = DoHTTPSPost("authserver.mojang.com", "/authenticate", json_request, ref result);
-                if (code == 200)
+                int code = 0;
+                if (Settings.UseMCLeaks)
                 {
-                    if (result.Contains("availableProfiles\":[]}"))
+                    string json_request = "{\"token\": \""+user+"\"}";
+                    code = DoHTTPSPost("auth.mcleaks.net", "/v1/redeem", json_request, ref result);
+                    if (code == 200)
                     {
-                        return LoginResult.NotPremium;
+                        result = result.Split('\n')[1];
+                        Json.JSONData loginResponse = Json.ParseJson(result);
+                        if (loginResponse.Properties["success"].StringValue == "true")
+                        {
+                            session.ID = loginResponse.Properties["result"].Properties["session"].StringValue;
+                            session.PlayerName = loginResponse.Properties["result"].Properties["mcname"].StringValue;
+                            string newResult = "";
+                            DoHTTPSGet("api.mojang.com", "/users/profiles/minecraft/"+session.PlayerName, "", ref newResult);
+                            Json.JSONData UUIDJSON = Json.ParseJson(newResult);
+                            session.PlayerID = UUIDJSON.Properties["id"].StringValue;
+                            return LoginResult.Success;
+                        }
+                        else
+                        {
+                            return LoginResult.WrongPassword;
+                        }
                     }
                     else
                     {
-                        Json.JSONData loginResponse = Json.ParseJson(result);
-                        if (loginResponse.Properties.ContainsKey("accessToken")
-                            && loginResponse.Properties.ContainsKey("selectedProfile")
-                            && loginResponse.Properties["selectedProfile"].Properties.ContainsKey("id")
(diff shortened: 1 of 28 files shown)


MinecraftClient/Protocol/ProtocolHandler.cs
```diff
@@ -767,7 +827,7 @@ public static string GetRealmsWorldServerAddress(string worldId, string username
         /// <param name="cookies">Cookies for making the request</param>
         /// <param name="result">Request result</param>
         /// <returns>HTTP Status code</returns>
-        private static int DoHTTPSGet(string host, string endpoint, string cookies, ref string result)
+        public static int DoHTTPSGet(string host, string endpoint, string cookies, ref string result)
```
ORelio, line 830:
```suggestion
        private static int DoHTTPSGet(string host, string endpoint, string cookies, ref string result)
```
No need to switch this method to `public` except maybe for testing.

MinecraftClient/Resources/config/MinecraftClient.ini
```diff
@@ -11,6 +11,7 @@ password=
 serverip=
 type=mojang                        # Account type. mojang or microsoft
 method=mcc                         # Microsoft Account sign-in method. mcc OR browser
+mcleaks=false
```
ORelio, line 14: Maybe you could allow `type=mcleaks` in addition to `mojang` and `microsoft` instead of adding another setting.

MinecraftClient/Protocol/ProtocolHandler.cs
```diff
@@ -370,47 +370,79 @@ private static LoginResult MojangLogin(string user, string pass, out SessionToke
...
                 }
-                else if (code == 503)
-                {
-                    return LoginResult.ServiceUnavailable;
-                }
-                else
-                {
-                    ConsoleIO.WriteLineFormatted(Translations.Get("error.http_code", code));
-                    return LoginResult.OtherError;
+                else {
```
ORelio, line 402:
```suggestion
                else
                {
```

MinecraftClient/Protocol/ProtocolHandler.cs
```diff
@@ -370,47 +370,79 @@ private static LoginResult MojangLogin(string user, string pass, out SessionToke
...
+                    {
+                        return LoginResult.ServiceUnavailable;
+                    }
+                    else
+                    {
+                        ConsoleIO.WriteLineFormatted(Translations.Get("error.http_code", code));
+                        return LoginResult.OtherError;
+                    }
                 }
+
```
ORelio, line 445:
```suggestion
```

ORelio:
Hi,

Thanks for your contribution.
The pull request is good, especially for a first-time contribution 👍

However, as you said and as stated in #1362 this kind of service offering disposable accounts do not really let you invest time on a server because you'll lose access to the account anytime. Although it can be fine for single player mode or short multiplayer sessions e.g. on mini-games servers, people will use MCC to AFK and automate things, not for actual playing. Automation combined with disposable accounts is a perfect recipe for misuses and we don't want to encourage that.

For this reason I cannot merge your pull request but nonetheless I'll provide some review.

ORelio (commented):
Here are some changes I'd suggest if merging to the main codebase:

MinecraftClient/Protocol/ProtocolHandler.cs
Tiebe: Ahh yes, I switched this to public because I needed to use it in another class, but I found a smarter way to do it afterwards

MinecraftClient/Resources/config/MinecraftClient.ini
Tiebe: Done!

Tiebe:
Thanks for reviewing it, even though it isn't going to get merged!

MinecraftClient/Resources/config/MinecraftClient.ini
Tiebe: Oh wait haha I did it in the method instead of the type gimme a sec

Tiebe: There that should work now

Tiebe:
How does that look @ORelio?

MinecraftClient/Settings.cs
```diff
@@ -281,12 +281,12 @@ public static void LoadSettings(string settingsfile)
                                                 case "login": Login = argValue; break;
                                                 case "password": Password = argValue; break;
                                                 case "type": AccountType = argValue == "mojang" 
-                                                        ? ProtocolHandler.AccountType.Mojang 
+                                                        ? ProtocolHandler.AccountType.Mojang
```
ReinforceZwei, line 284: Should probably use `switch` here for readability.

Tiebe: Something like what I did now? (Haven't tested yet though)

ORelio: Seems better indeed 👍

Tiebe:
Just added TheAltening support as well!

ORelio:
Since the feature seems complete, I'll close the pull request as per https://github.com/ORelio/Minecraft-Console-Client/pull/1577#issuecomment-835252202.

Issue #1510, opened 2021-03-26
Loading and unloading of bots are unreliable


Vort:
There are multiple issues I want to describe in this report.

First of all, it looks like it is not possible to maintain single instance of script without hacks.
If you put in `tasks.ini` `triggerOnFirstLogin`, then after some amount of relogins script will unloaded and not loaded anymore.
If you put `triggerOnLogin`, then after several relogins you will end up with lots of script copies loaded.

To solve this problem I have added such code in my bot (`OreMiner2`):
```csharp
        if (GetLoadedChatBots().Count(x => x is OreMiner2) > 1)
        {
            LogToConsole("OreMiner 2 is already loaded.");
            UnloadBot();
            return;
        }
```

But it did not helped too much since MCC despite `UnloadBot` being called, feels ok to invoke handlers like `Update` for uninitialized and unloaded bot.
I have added checks to return from that functions ASAP since if I allow them to throw exception, MCC often crashes.

But it did not solved problems too!
Because loading and unloading process can produce crashes by itself!

First thing that I have noticed is that sometimes in `DispatchBotEvent` function last element of `selectedBots` array becomes null, which means crash.
Then I decided to make minimized script which can reproduce this problem.
And got another crash:

![image](https://user-images.githubusercontent.com/1242858/112592832-48163280-8e0f-11eb-8a58-801d6c088820.png)

Here is the script:

```csharp
//MCCScript 1.0

MCC.LoadBot(new LoadUnloadBug());

//MCCScript Extensions

public class LoadUnloadBug : ChatBot
{
    public override void Initialize()
    {
        LogToConsole("Load");
        if (GetLoadedChatBots().Count(x => x is LoadUnloadBug) > 1)
        {
            UnloadBot();
            LogToConsole("Unload");
        }
    }
}
```

As you can see at screenshot, I was testing latest version with LoadUnloadBug script.

Please try to improve reliability of loading and unloading process.

I can try to get additional information if you will start working on this problem and will have a need for it.

Vort:
More useful test script:

```csharp
//MCCScript 1.0

MCC.LoadBot(new LoadUnloadBug());

//MCCScript Extensions

public class LoadUnloadBug : ChatBot
{
    List<int> list;

    public override void Initialize()
    {
        LogToConsole("Load");
        if (GetLoadedChatBots().Count(x => x is LoadUnloadBug) > 1)
        {
            UnloadBot();
            LogToConsole("Unload");
            return;
[...]

PR #1579, opened 2021-05-08
Implement schedule main thread task with return value

ReinforceZwei:
I tried my best to implement this. Unfortunately for delayed task there is no way to get the returned value or the calling thread will have to wait until the task has executed.

image

### DIFF
--- a/MinecraftClient/ChatBot.cs
+++ b/MinecraftClient/ChatBot.cs
@@ -72,7 +72,7 @@ namespace MinecraftClient
                     {
                         if (delayTasks[i].Tick())
                         {
-                            Handler.ScheduleTask(delayTasks[i].Task);
+                            delayTasks[i].Task.DynamicInvoke();
                             tasksToRemove.Add(i);
                         }
                     }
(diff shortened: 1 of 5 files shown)


ORelio:
Hi,
Of course if you want to retrieve the value you need to wait. That seems normal. And it will be useful for scripts (#1510)
You can launch the task asynchronously only if you don't care about the result 😉

ReinforceZwei:
There is a new class `TaskWithResult` but I think the class name is not good enough to reflect the class functionality. Is there any better idea?

ReinforceZwei:
_Oops I accidentally commited my test code._

ORelio:
Hi, I hadn't enough time on my side to fully review this PR before it got merged (sorry), so I reworked the feature afterwards in commit c1cfaf520d04dc939376700dab5f7bbe1d743c95. Basically it works as before but with various improvements such as type handling for return value, exception handling, and suppressing the need to write `new Action()` each time we want to invoke a task on main thread. Now I'll look into making the ChatBot API thread-safe using this new mechanism. Thanks again for your contribution 😉

PR #1585, opened 2021-05-12
Use dedicated thread to process world data

### LINKED ISSUE #1583: Handle terrain and movement in separated thread/task
**Prerequisites**
 - [x] I have read and understood the [user manual](https://github.com/ORelio/Minecraft-Console-Client/tree/master/MinecraftClient/config)
 - [x] I tried to [look for similar feature requests](https://github.com/ORelio/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

## Describe your problem
It is not possible to use `terrainandmovement` on low-end devices such as raspberry pi as handling world data hang-up the main thread and causing connection timeout because of not able to response keep alive packet.

## Suggest a possible solution
Process world data packet and movement calculation in a new thread/task, and make use of the new `ScheduleTask()` method to handle the result on main thread.

***

Related issue: #219, #970

ReinforceZwei:
This change allow low-end device to join server without being timeout. I decided not to change `/move` command as it doesn't cause connection timeout.

Tested with Raspberry Pi 4 4GB. It is possible to join server with `terrainandmovement` enabled now.

Resolve #1583

### DIFF
--- a/MinecraftClient/McClient.cs
+++ b/MinecraftClient/McClient.cs
@@ -699,6 +699,7 @@ namespace MinecraftClient
         /// Schedule a task to run on the main thread
         /// </summary>
         /// <param name="task">Task to run</param>
+        /// <returns>Any result returned from delegate</returns>
         public object ScheduleTask(Delegate task)
         {
             var taskAndResult = new TaskWithResult(task);
--- a/MinecraftClient/Protocol/Handlers/Protocol18.cs
+++ b/MinecraftClient/Protocol/Handlers/Protocol18.cs
@@ -18,6 +18,7 @@ using System.Diagnostics;
 using MinecraftClient.Inventory.ItemPalettes;
 using MinecraftClient.Protocol.Handlers.PacketPalettes;
 using MinecraftClient.Logger;
+using System.Threading.Tasks;
 
 namespace MinecraftClient.Protocol.Handlers
 {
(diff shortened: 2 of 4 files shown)


ORelio:
Could be nice indeed. I think we could do this once the cross-thread invoke mechanism is done. Before that I'd want to make the ChatBot API thread-safe to solve #1510 and many other issues encountered by people trying to use threads without knowing that the API is not thread-safe.

ORelio (commented):
Since packet parsing happens in the main thread, and the task is scheduled using task scheduler that runs tasks in the main thread, how this change will make terrain parsing faster?

The heaviest part is parsing chunk column data in `ProcessChunkColumnData`, I think this part could be offloaded in another thread, and the task inserting final parsed terrain data in the world structure should be scheduled back in the main thread to be safe.

Other parts (single block changes) do not need to be offloaded in another thread.

ReinforceZwei:
> The heaviest part is parsing chunk column data in ProcessChunkColumnData, I think this part could be offloaded in another thread, and the task inserting final parsed terrain data in the world structure should be scheduled back in the main thread to be safe.

@ORelio This is exactly what I did in this PR.

ORelio:
I that case I have misread your PR.

ORelio (approved):

PR #1612, opened 2021-06-01
Invalid path name to resource causing build to fail

alex-huff:
Cannot build with msbuild on ubuntu because it cannot find some resources that have invalid path names.


ORelio:
Thanks for the fix!

PR #1617, opened 2021-06-02
Prevent errors from OnPlayerLeave() returning NULL

Daenges:
I suffered from some errors when I started the script. I soon found out, that the client detects some players leaving and I was wondering why, because there was not even a leaving message. I soon found out that the errors came from OnPlayerLeave() which returned a UUID but the name parameter was NULL. I don't know whether this is a bug in the client, but this should fix it for the script.

### DIFF
--- a/MinecraftClient/config/ChatBots/DiscordWebhook.cs
+++ b/MinecraftClient/config/ChatBots/DiscordWebhook.cs
@@ -317,17 +317,33 @@ class DiscordWebhook : ChatBot
 
     public override void OnPlayerJoin(Guid uuid, string name)
     {
-        if (settings.GetNamesToUuidMojangCache().ContainsKey(name))
+        if (uuid.ToString() != string.Empty && name != string.Empty &&
+            uuid.ToString() != null && name != null)
         {
-            settings.GetNamesToUuidMojangCache().Remove(name);
+            if (settings.GetNamesToUuidMojangCache().ContainsKey(name))
+            {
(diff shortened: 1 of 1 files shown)


Daenges:
This was the error I was getting, including a line, where I print out the values of the variables UUID and Name:
```
[MCC] [DiscordWebhook] UUID: f924c66a-be5a-4267-979f-f66fe436d2e4, Playername:
[ERROR] OnPlayerLeave: Got error from ScriptLoader.Script+DiscordWebhook: System.ArgumentNullException: Der Wert darf nicht NULL sein.
Parametername: key
   bei System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   bei ScriptLoader.Script.DiscordWebhook.OnPlayerLeave(Guid uuid, String name)
   bei MinecraftClient.McClient.<>c__DisplayClass148_0.<OnPlayerLeave>b__0(ChatBot bot)
   bei MinecraftClient.McClient.DispatchBotEvent(Action`1 action, IEnumerable`1 botList)
```

ORelio:
Hi @Daenges and thanks for the fix. The UUID is always set but the name may be missing if MCC itself does not know what name belongs to this UUID:
https://github.com/ORelio/Minecraft-Console-Client/blob/dd992dc4eea848d71d313100b9e94348e02b5a0d/MinecraftClient/McClient.cs#L2060-L2074
That's because in the protocol, MCC only gets the UUID, it finds the name from its own internal dictionary.

Issue #1313, opened 2020-11-05
Porting MCC to .NET Framework 5


downloadpizza:
The project works well in Mono on Linux, but the `.Net Framework`  is slowly dying out. `.Net core` was also "killed" and moved to `.Net 5.0` (and higher).

`.Net 5.0` is apparently almost at the first release, so it may be good to consider porting the project to unified `.Net` once it is released.

I don't see this as an important Issue, but it would be quite nice. I'll start looking for namespaces that dont exist in `.Net 5.0` to see what needs to be replaced. I assume the GUI stuff will completely break, but no idea.

downloadpizza:
Non Standard Namespaces: `System.Deployment`, `System.Core`, `System.Data.DataSetExtensions`, 

Ill see if they are available on NuGet for `.Net core` as that probably means that they will get ported to `5.0`

ORelio:
Discussed about it in several issues before (e.g #1306 recently) but there is not proper issue for .NET 5 yet so I'll leave it open as it is a bit soon for porting yet.

breadbyte:
Hi! 

I just started porting this project to .NET 5.0. 
There are still some issues, but I've done the initial work to get it up and running on a Linux machine.
You can see my work over at https://github.com/breadbyte/Minecraft-Console-Client under the `bread` branch.
I might make a pull request once I've tested it more thoroughly and I'm certain about the stability of the code, but for now I have only tested it with 1.17.

![image](https://user-images.githubusercontent.com/14045257/124256389-d06a4900-db5d-11eb-9e2a-e355d49e27cb.png)

ORelio:
Looks very promising! The code contains may workaround for Mono that you will be able to remove once ported to .NET 5. This should also resolve weird issues such as the mysterious silent disconnection bug (https://github.com/ORelio/Minecraft-Console-Client/issues/1306#issuecomment-720159571)

PR #1661, opened 2021-07-02
.NET 5.0 Support

breadbyte:
This is a very far-reaching pull request, and is a continuation of #1313
Due to the size and amount of changes of this pull request, it is marked as draft.

I am making this pull request as a tracker on what things need to be tested and what things need to be worked on.
I am also planning to make even more radical changes to the code that may be out of scope for this PR, which may be incompatible with this project due to differing views, and this PR will help limit that.
There are TODO lines on the code where a universal non-windows implementation is required.

I currently have an issue tracking my testing and implementation progress over at https://github.com/breadbyte/Minecraft-Console-Client/issues/1 in order to not pollute this project with my testing.

This code currently can connect to a 1.17 and 1.16.5 server (with Terrain Handling), but most other functions and/or versions are untested.

Comments on the code and any ideas on what to test are welcome.

# Changes

## .csproj
70c10d6, f53767c, 594cf22, cbd6206, b4ce551 change the csproj files. The project is now built under `AnyCpu` instead of `x86`, and the project is now building under C# 9.0, with nullables enabled. 

Additionally, 594cf22 adds an XUnit Project to help with testing.

## HejidenDNS
b5cdc3b and 1f886fb replaces HejidenDNS with [DnsClient](https://github.com/MichaCo/DnsClient.NET). **This may be a controversial change.** 
The reason for this change is because HejidenDNS makes use of `AsyncResult` from `System.Runtime.Remoting.Messaging`, which does not exist in .NET 5.0., and [DnsClient](https://github.com/MichaCo/DnsClient.NET) is an alternative DNS Client which supports .NET 5.0.

I have tested that DnsClient works with Minecraft SRV Records, but I only have one testing server and one SRV Record, so more testing would be ideal.

## Thread.Abort
e7608f2 replaced `Thread.Abort` with `CancellationTokens`. **This may be a controversial change.**
The reason for this change is because .NET 5.0 no longer implements this function and throws an exception instead.
See https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/thread-abort-obsolete.

I have written `CancellationToken` support in the least invasive way I can, but thoughts are always welcome.

## Miscellaneous Changes
2a6ca7f updates the gitignore files and can be rebased off.
5d5bfa9 and e411eb1 removes some unused System.Windows.Forms usings, which prevented the project from building.


ORelio:
First of all, congrats for your work 👏

> b5cdc3b and 1f886fb replaces HejidenDNS with DnsClient. This may be a controversial change.

HejidenDNS seems unmaintained so any replacement that has support for SRV records is fine.
Just make sure your pull request updates [README.md#credits](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/config/README.md#credits) to list the correct library.

> e7608f2 replaced Thread.Abort with CancellationTokens. This may be a controversial change.

Since it's now impossible to use Thread.Abord and CancellationTokens is the recommended method by Microsoft, it's OK to me.

I'll review your PR more thoroughly when I have more time for that 😉

breadbyte:
The latest update adds scripting support.
This has been broken due to the fact that CodeDOM no longer provides a dynamic compiler, and throws a `PlatformNotSupportedException` instead.
This has been replaced with code from https://github.com/laurentkempe/DynamicRun which leverages Roslyn instead.

I have tested the sample script from https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/config/sample-script.cs and it is working fine.

image

breadbyte:
The latest commits add a check for `(RuntimeInformation.IsOSPlatform(OSPlatform.Windows))` on windows-specific code.

Unfortunately, this is the last commit that I will be able to do before having to rewrite most of the underlying code to dig deeper into other problems.

I am experiencing some rather odd issues with the console and network dropping on Linux particularly, so I would have to look deeper into the code and rework it, most likely.

PR #1674, opened 2021-07-08
NET 5.0 Support

breadbyte:
Continuation of #1661, I just renamed the branch, the same commits apply here.


breadbyte:
Sorry for the hiatus! Life took over for a couple of weeks, but I am now free to work on this project. Just a quick update on a few issues:

# Clipboard Functionality
https://github.com/ORelio/Minecraft-Console-Client/pull/1674/commits/5f75a55ef1301bd3ffe86be576c37628d79cdb51 reintroduces the clipboard pasting mechanism through the nuget package [TextCopy](https://github.com/CopyText/TextCopy), licensed under MIT License.

It requires `xsel` to be installed for Linux users.

Note: `xsel` not being installed is not a problem for people connecting through terminals (i.e. PuTTY, Windows Terminal) / over SSH as the client-side terminal handles the clipboard operations themselves. 

For people using Linux with a desktop environment, this is a requirement. 

I have not tested it with WSL (Windows Subsystem for Linux), but TextCopy says it should work out of the box without any issue.

# Threads and Tasks

https://github.com/ORelio/Minecraft-Console-Client/pull/1674/commits/f8fb26294bcfb010537d81604ef7b97d3a8d068a has a bug where, after reconnecting, you have to press enter twice for the first chat message to be sent off. Subsequent chat messages are not affected.

## Why? And what's with the title?
The reason why https://github.com/ORelio/Minecraft-Console-Client/pull/1674/commits/f8fb26294bcfb010537d81604ef7b97d3a8d068a has this bug is because of the way MCC works with threads.

The current threading system that Minecraft Console Client has forcefully kills the [existing `ConsoleIO.ReadLine` blocking operation](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/McClient.cs#L524) with `Thread.Abort` whenever a client restart is requested.

Unfortunately, there is no way to stop `ConsoleIO.ReadLine` in the middle of it's execution through CancellationTokens or Tasks, because reading from the console is [inherently a synchronous operation](https://docs.microsoft.com/en-us/dotnet/api/system.console.in?view=net-5.0#remarks), so it will keep blocking until a message is received.

So we simply need to let the previous `ConsoleIO.ReadLine` complete itself by entering another keypress. The previous command handler will not proceed as a Cancellation is already requested by the time it completes, so it will terminate itself because of the [CancellationToken checks I added in](https://github.com/breadbyte/Minecraft-Console-Client/blob/net5.0/MinecraftClient/McClient.cs#L528-L529).

I also encountered this issue when trying to kill the [offline command prompt](https://github.com/ORelio/Minecraft-Console-Client/blob/349a3de8e4a5fb5d2af3fe9b74aa700e2de2a1a8/MinecraftClient/Program.cs#L466).

`Reco` and `Connect` was completing before the CancellationTokens received a cancellation (because threads), so the offline command handler runs again, while a new `Client` is created. The offline command handler never finishes.

The way this currently works is by `Thread.Abort`ing the offline command prompt thread as soon as a restart was requested.

I looked deeper into the `Reco` and `Connect` commands and figured out that they both return an empty string on success, so I added an empty string check to safely exit from the thread.

Unfortunately this is not possible with the Command Handler as the Command Handler must run indefinitely.

A complete fix for this issue would be to rewrite the entire command handler to properly accommodate for one and only one `Console.ReadLine` call at a time. 

The current solution of having the user press the enter key twice is not too invasive to the current codebase, but might be annoying for other users.

## Linux

On Linux, however, the situation is more dire and it is unusable. 

A short list of issues that I have noticed on Linux:
- Sometimes, chat messages require an enter key and a keypress event (a keypress event specifically, any key pressed) for a message to be sent.
- Sometimes, chat messages fail to be shown to the client, and a key must be pressed in order to show the chat message.
- Disconnection messages will not be sent to the user until a key has been pressed, printing each individual line per keypress.
- Command messages are unaffected, however.

I am not too sure why the program execution differs from Windows to Linux, but I digress. (`Works on my machine :^)`)

The gist of it is, because of how threading is used in MCC, it causes a plethora of issues, and the best course of action would be to rewrite the entire thread-based system to work with Tasks better.

Because rewriting the entire system is a monumental task, I am also looking into other ways of making it work with minimal intrusion to the existing codebase.

I will still need to do extensive work before I can deem this PR ready.

breadbyte:
Thanks for the input.

I would just like to ask,
What is `BasicIO` and what is it's intended use case?
Looking at the code, it looks like something that is meant for non-interactive, unattended use.
Using the flag however, yields an interactive mode.

ReinforceZwei:
@breadbyte You may want to take a look at the [wiki about command arguments](https://github.com/ORelio/Minecraft-Console-Client/wiki/Command-arguments#explanation)

ORelio:
`BasicIO` is for disabling terminal escape sequences, colors are outputted as `§c` escape codes, and `BasicIO-NoColor` is the same without `§c` escape codes. These modes are to be used when launching MCC from a wrapper program like the GUI or when using a dumb terminal / when not looking at the output e.g. when using `screen` or launching MCC as a service.

breadbyte:
Hello! I come back from the grave after writing some more code.

I decided to write my own console mechanism to figure out what is going wrong with MCC's.
It turns out threading was not the problem at all with the console issues! Well, not exactly.

It happens because on non-Windows systems, reading and writing to the console is disallowed by this [lock](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Console/src/System/ConsolePal.Unix.cs#L464-L469) in `System.Console`.

Therefore, I had to replace the entire console mechanism with something else.
I used [gui.cs](https://github.com/migueldeicaza/gui.cs) to replace the current console mechanism. It is a terminal GUI library licensed under the MIT License.

I've replaced the entire console handler with great results.
The video shows an older iteration of the UI, but it is functionally the same.

https://user-images.githubusercontent.com/14045257/130371653-85f37e8e-e876-4ff9-bbde-71969270cf26.mp4

However, when the application crashes on Linux, a `reset` of the terminal is required due to the way it handles the terminal.
So far, I have used it without any problems whatsoever.

The only problem it has so far is, there is no word wrapping.
I am still thinking about how to implement word wrapping, but so far it is at a usable state.

I had to introduce a change where I had to replace the `Write` method with `WriteLine` due to the way the console output works.
Any other usage of `Console.Write` behind `BasicIO` is untouched.

I haven't put xmldoc comments in the code yet, but I will go over the code with a fine-toothed comb to catch anything I might have missed.

MinecraftClient/Program.cs
```diff
@@ -35,59 +39,75 @@ static class Program
...
-            if (isUsingMono || WindowsVersion.WinMajorVersion >= 10)
-            {
-                Console.OutputEncoding = Console.InputEncoding = Encoding.UTF8;
+            if (ConsoleIO.BasicIO) {
+                if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows)) {
+                    Console.WriteLine("BasicIO is not supported on non-Windows systems.");
+                    Console.WriteLine("Press any key to continue...");
+                    Console.ReadKey();
+                    Environment.Exit(1);
+                }
```
ORelio, line 78: Any reason to prevent BasicIO on Linux? The BasicIO mechanism should rely on ReadLine() and WriteLine() only without console formatting or manipulating the terminal, so that should work on all OS 🤔
... Or is that related to the lock in System.Console you talked about before?

breadbyte: Yes. We can't `Console.ReadLine` and `Console.WriteLine` at the same time on non-windows systems because of the lock.
Sorry, my wording wasn't really clear about that :P

ReinforceZwei:
For me, I think the new console GUI is a bit over-killed. I would prefer the "original-style" MCC console. If the console read/write lock is the problem, could [this](https://stackoverflow.com/questions/7464448/how-to-read-write-on-console-at-the-same-time-using-threads) solves the problem?

Daenges:
I know that I am not qualified enough to judge any code related stuff of this PR, but from a visual point I really need to agree with ORelio.
It may be overkill currently, but this provides many new opportunities to make the client more visually appealing. 
If you configure GUI correctly this could make the client way more accessible, also for new users.
On top of that, you gain a really professional look through all those new design elements.
This professional look might also liquidate a new users first doubts of legitimacy, created by the big warning of the Windows Defender on the first startup.

breadbyte:
Upon further inspection as directed by @ReinforceZwei, I have had a major lapse of understanding of the issue.

It is, in fact, possible to `Read` and `Write` on Linux at the same time.
This means that `BasicIO` is actually possible on all OSes.

Where the `lock` comes into play however, is when accessing `ConsoleLeft` and `ConsoleTop`.

Trying to create a custom console handler IS the one that is causing lock contention between Read and Write.
The `lock` prevents access to `ConsoleLeft` and `ConsoleTop` while a `Read` operation is in progress.
With the way I have implemented `Write`, it uses `ConsoleLeft` and `ConsoleTop` to move around the console to write, while at the same time another thread does a `ReadKey` operation.

This causes the lock to engage and wait for `ReadKey`, making me think that `Read` and `Write` are blocking each other.

Implementing the [option ReinforceZwei offered](https://stackoverflow.com/questions/7464448/how-to-read-write-on-console-at-the-same-time-using-threads), fixes this locking issue.

I might have to read a bit more into it to see why console performance in Linux is quite slow, so I am going to continue working on a pure `System.Console` alternative as an option for MCC.

I implemented a small `System.Console`-based console handler [here](https://github.com/breadbyte/ConsoleInteractive) if you would like to take it for a spin and give feedback.

---

I am not opposed to a simpler user interface without the fancy bells and whistles.
The TUI is good because it allows a pathway for better user accessibility and interactivity, especially with the mouse.
For new users who are inexperienced, we could create an experience that is more interactive than just typing on a console.

Personally, I would like the bottom bar in my UI demo to tell me server status and/or time just to tell me that it's still alive and connected.

But developing this experience will take time.

If we take a step backwards, this application is only going to be used the majority of the time inside cloud instances and virtual private servers. Is it really worth the time to create a fancy user interface for something that the user will only interact with once in a while?

Not to mention the amount of extra code that we will need to introduce. Granted, the user can use BasicIO for a more simple experience, but just for something basic that also gives you the extra features of formatted text and uninterrupted typing, a TUI would indeed be overkill.

Take the Minecraft Server software for example, it only gives you a console output and an input. Simple and familiar for people who run minecraft servers.

Either option is fine with me. I would like the TUI of course, but that could introduce a new set of issues and would have to spend more time maintaining it.

I accept everyone's viewpoint, but of course, it is not up to me to decide.
I am simply just one person trying to improve MCC on my own time. If a TUI would be too much for MCC, I have no problem with it.

ORelio:
Thanks for your work! A simple prompt like the server console is completely fine, that's what I tried to do when implementing ConsoleIO initially. Your library looks good: similar to ConsoleIO but simpler, cleaner 👍

xXjojaXx:
Are you still working on this?

breadbyte:
> Are you still working on this?

Absolutely! Progress is taking slower than what I would have liked or expected due to personal issues.

xXjojaXx:
> > Are you still working on this?
> 
> Absolutely! Progress is taking slower than what I would have liked or expected due to personal issues.

Don!'t worry, take your time! Was just wondering, because the last message was months ago.

breadbyte:
I do apologize for the massive hiatus. 

The recent months have not been very kind to me, as my household has contracted covid, and it took a while to recover, but I am back on my feet now and able to work on projects again.


On to the outstanding issues.

# Recap

To recap, accessing `Console.Left` and `Console.Top` is a blocking operation for `Console.Read/ReadLine` as they operate over the same lock to prevent returning the wrong results from both functions simultaneously running.

A `Console.Read` can run waiting for a read while an access to `Console.Left/Top` is happening in the background, which can cause a deadlock, so we need to fix this issue. 

The [fix](https://stackoverflow.com/questions/7464448/how-to-read-write-on-console-at-the-same-time-using-threads) suggested earlier fixes this issue.

A good example of this is if we are printing to the console, which requires setting the `Console.Left/Top` position, and then a `Console.Read` is happening simultaneously in the background.

There also used to be a major performance hit when trying to access `Console.Left` / `Top` in .NET 5, which I had to fix by keeping track of the horizontal and vertical position on our own. 

This actually seems to have fixed itself in the recent .NET updates, so I might have to either rework my code or keep it as-is, considering the amount of time I spent on the manual tracking code.

# What's next?

The only final issue I have to tackle is the use of `Console.ReadLine`.

According to the [official documentation for Console.ReadLine](https://docs.microsoft.com/en-us/dotnet/api/system.console.readline?view=net-5.0#remarks),
> If the standard input device is the keyboard, the ReadLine method blocks until the user presses the Enter key.

This unfortunately means that we have to rewrite all usages of `Console.ReadLine`, as a ReadLine will lock the console until it is completed.

The aforementioned fix for the deadlock is `Console.IsKeyAvailable`. This only corresponds if there is a _key_ available, not whenever a complete input is available. So we have to build our own input buffer using `Console.ReadKey`.

This comes at a very expensive cost of having to rewrite how the console handler works to have it work off buffering `Console.ReadKey` instead of `Console.ReadLine`.

This will not affect `BasicIO` as it does not use any console positioning code.

# The fix

I have written a new [library](https://github.com/breadbyte/ConsoleInteractive) that takes all this into account, as well as `CancellationToken` support so the console can be terminated at any point within the program's execution.

A lot more difficult to implement, but a lot more performant.

This should also have the potential to fix the outstanding issues about CJK/Unicode input/output. Admittedly I am in no way qualified to test this, since I do not speak/read non-latin languages.

I do understand however that taking on a new library, especially from an outside contributor like me is a major commitment, as any issues that may arise from using this library will be therefore my responsibility, and if I am unavailable, it will be a significant issue.

ORelio:
Hi,
Thank you for working that hard on a major upgrade. We are all volunteers here, so you don't need to apologise.
I think that solving the console issues with a custom libraries is not really a concern. The ConsoleIO class was already custom made to begin with, and people other than me already had issues dealing with it. And it had long standing issues as you already know 😅

breadbyte:
This should be good for review now.

The console mechanism is the same as before, with some small changes.

https://user-images.githubusercontent.com/14045257/142914315-57fa36b9-80ff-4aad-a62f-450b2b12584a.mp4


This might be too large for a code review, should I split it up?

ReinforceZwei:
I have a suggestion regarding the code style. I saw you've used 
```c#
if (...) { // <- the curly brace here
    // something
}
```
instead of
```c#
if (...)
{ // <- the curly brace here
    // something
}
```
Since we are using the second style all the time in the code, would you mind moving the curly brace to the next line so that the style would be more consistent with the existing code?

breadbyte:
Fixed the code style for the code that i've written.

ORelio:
Looks good, will take some time to review. For consistency with older releases and convenience, do you think it would be possible to put back the `>` character that shows the user there's is a chat prompt and they can start typing?

breadbyte:
It shouldn't be too hard to add in (fingers crossed), but i'll work it out.

MinecraftClient/Protocol/Handlers/Protocol16.cs
```diff
@@ -194,11 +200,10 @@ private bool processPacket(byte id)
             return true; //packet has been successfully skipped
         }
 
-        private void StartUpdating()
-        {
-            netRead = new Thread(new ThreadStart(Updater));
-            netRead.Name = "ProtocolPacketHandler";
-            netRead.Start();
+        private void StartUpdating() {
```
ReinforceZwei, line 203:
```suggestion
        private void StartUpdating() 
        {
```

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -150,14 +149,16 @@ public Protocol18Handler(TcpClient Client, int protocolVersion, IMinecraftComHan
         /// <summary>
         /// Separate thread. Network reading loop.
         /// </summary>
-        private void Updater()
-        {
-            try
-            {
+        private void Updater(object? o) {
```
ReinforceZwei, line 152:
```suggestion
        private void Updater(object? o) 
        {

```

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -150,14 +149,16 @@ public Protocol18Handler(TcpClient Client, int protocolVersion, IMinecraftComHan
...
         /// </summary>
-        private void Updater()
-        {
-            try
-            {
+        private void Updater(object? o) {
+            if (((CancellationToken) o!).IsCancellationRequested)
+                return;
+
+            try {
```
ReinforceZwei, line 156:
```suggestion
            try 
            {

```

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -150,14 +149,16 @@ public Protocol18Handler(TcpClient Client, int protocolVersion, IMinecraftComHan
...
                 bool keepUpdating = true;
                 Stopwatch stopWatch = new Stopwatch();
-                while (keepUpdating)
-                {
+                while (keepUpdating) {
```
ReinforceZwei, line 159:
```suggestion
                while (keepUpdating) 
                {

```

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -1135,11 +1140,10 @@ internal bool HandlePacket(int packetID, Queue<byte> packetData)
         /// <summary>
         /// Start the updating thread. Should be called after login success.
         /// </summary>
-        private void StartUpdating()
-        {
-            netRead = new Thread(new ThreadStart(Updater));
-            netRead.Name = "ProtocolPacketHandler";
-            netRead.Start();
+        private void StartUpdating() {
```
ReinforceZwei, line 1143:
```suggestion
        private void StartUpdating() 
        {

```

MinecraftClient/WinAPI/ConsoleIcon.cs
```diff
@@ -32,44 +31,46 @@ public enum WinMessages : uint
             SETICON = 0x0080,
         }
 
-        private static void SetWindowIcon(System.Drawing.Icon icon)
-        {
-            IntPtr mwHandle = System.Diagnostics.Process.GetCurrentProcess().MainWindowHandle;
-            IntPtr result01 = SendMessage(mwHandle, (int)WinMessages.SETICON, 0, icon.Handle);
-            IntPtr result02 = SendMessage(mwHandle, (int)WinMessages.SETICON, 1, icon.Handle);
+        private static void SetWindowIcon(System.Drawing.Icon icon) {
```
ReinforceZwei, line 34:
```suggestion
        private static void SetWindowIcon(System.Drawing.Icon icon) 
        {
```

MinecraftClient/WinAPI/ConsoleIcon.cs
```diff
@@ -32,44 +31,46 @@ public enum WinMessages : uint
...
-                    try
-                    {
-                        using (HttpWebResponse httpWebReponse = (HttpWebResponse)httpWebRequest.GetResponse())
-                        {
-                            try
-                            {
-                                Bitmap skin = new Bitmap(Image.FromStream(httpWebReponse.GetResponseStream())); //Read skin from network
-                                SetWindowIcon(Icon.FromHandle(skin.GetHicon())); // Windows 10+ (New console)
-                                SetConsoleIcon(skin.GetHicon()); // Windows 8 and lower (Older console)
+                Thread t = new Thread(new ThreadStart(delegate {
```
ReinforceZwei, line 49:
```suggestion
                Thread t = new Thread(new ThreadStart(delegate 
                {
```

MinecraftClient/WinAPI/ConsoleIcon.cs
```diff
@@ -32,44 +31,46 @@ public enum WinMessages : uint
...
+                        HttpWebRequest httpWebRequest =
+                            (HttpWebRequest) HttpWebRequest.Create(
+                                "https://minotar.net/helm/" + playerName + "/100.png");
+                        try {
```
ReinforceZwei, line 53:
```suggestion
                        try 
                        {
```

MinecraftClient/WinAPI/ConsoleIcon.cs
```diff
@@ -32,44 +31,46 @@ public enum WinMessages : uint
...
+                            using (HttpWebResponse httpWebReponse = (HttpWebResponse) httpWebRequest.GetResponse()) {
+                                try {
```
ReinforceZwei, line 55:
```suggestion
                                try 
                                {
```

MinecraftClient/WinAPI/ConsoleIcon.cs
```diff
@@ -32,44 +31,46 @@ public enum WinMessages : uint
...
+                                    Bitmap skin = new Bitmap(Image.FromStream(httpWebReponse.GetResponseStream())); //Read skin from network
+                                    SetWindowIcon(Icon.FromHandle(skin.GetHicon())); // Windows 10+ (New console)
+                                    SetConsoleIcon(skin.GetHicon()); // Windows 8 and lower (Older console)
+                                }
+                                catch (ArgumentException) {
```
ReinforceZwei, line 60:
```suggestion
                                catch (ArgumentException)
                                {
```

MinecraftClient/FileMonitor.cs
```diff
@@ -28,15 +28,14 @@ public FileMonitor(string folder, string filename, FileSystemEventHandler handle
...
-            try
-            {
-                monitor = new FileSystemWatcher();
-                monitor.Path = folder;
-                monitor.IncludeSubdirectories = false;
-                monitor.Filter = filename;
-                monitor.NotifyFilter = NotifyFilters.LastWrite;
-                monitor.Changed += handler;
-                monitor.EnableRaisingEvents = true;
+            try {
```
ReinforceZwei, line 31:
```suggestion
            try
            {
```

ReinforceZwei (commented):
Also use curly brace at new line for the `try-catch` block and methods.

MinecraftClient/ConsoleIO.cs
```diff
@@ -322,71 +112,39 @@ public static void Write(char c)
         /// If true, "hh-mm-ss" timestamp will be prepended.
         /// If unspecified, value is retrieved from EnableTimestamps.
         /// </param>
-        public static void WriteLineFormatted(string str, bool acceptnewlines = true, bool? displayTimestamp = null)
-        {
+        public static void WriteLineFormatted(string str, bool acceptnewlines = false, bool? displayTimestamp = null) {
```
ReinforceZwei, line 115:
```suggestion
        public static void WriteLineFormatted(string str, bool acceptnewlines = false, bool? displayTimestamp = null)
        {
```

MinecraftClient/ConsoleIO.cs
```diff
@@ -406,137 +164,23 @@ public static void WriteLogLine(string text, bool acceptnewlines = true)
...
-        /// </summary>
-        private static void GoLeft()
-        {
-            if (buffer.Length > 0)
-            {
-                buffer2 = "" + buffer[buffer.Length - 1] + buffer2;
-                buffer = buffer.Substring(0, buffer.Length - 1);
-                GoBack();
-            }
+        private static void ClearLineAndBuffer() {
```
ReinforceZwei, line 167:
```suggestion
        private static void ClearLineAndBuffer()
        {
```

MinecraftClient/ConsoleIO.cs
```diff
@@ -406,137 +164,23 @@ public static void WriteLogLine(string text, bool acceptnewlines = true)
...
-                        clipdata = Clipboard.GetText();
-                    }
-                    catch { }
-                }
-            ));
-            staThread.SetApartmentState(ApartmentState.STA);
-            staThread.Start();
-            staThread.Join();
-            return clipdata;
+        public static void AutocompleteHandler(object? sender, ConsoleKey e) {
```
ReinforceZwei, line 175:
```suggestion
        public static void AutocompleteHandler(object? sender, ConsoleKey e)
        {
```

MinecraftClient/McClient.cs
```diff
@@ -503,32 +517,33 @@ public void OnConnectionLost(ChatBot.DisconnectReason reason, string message)
...
         #region Command prompt and internal MCC commands
-
-        /// <summary>
-        /// Allows the user to send chat messages, commands, and leave the server.
-        /// </summary>
-        private void CommandPrompt()
-        {
-            try
+        
+        private void ConsoleReaderOnMessageReceived(object? sender, string e) {
```
ReinforceZwei, line 533:
```suggestion
        private void ConsoleReaderOnMessageReceived(object? sender, string e)
        {

```

MinecraftClient/Program.cs
```diff
@@ -147,8 +157,17 @@ static void Main(string[] args)
...
                 if (useBrowser)
                     ConsoleIO.WriteLine("Press Enter to skip session cache checking and continue sign-in with browser");
-                Console.Write(ConsoleIO.BasicIO ? Translations.Get("mcc.login_basic_io") + "\n" : Translations.Get("mcc.login"));
-                Settings.Login = Console.ReadLine();
+
+                if (ConsoleIO.BasicIO) 
+                {
+                    ConsoleIO.WriteLine(Translations.Get("mcc.login_basic_io"));
+                }
+                else {
```
ReinforceZwei, line 165:
```suggestion
                else
                {

```

MinecraftClient/Protocol/Handlers/Protocol16.cs
```diff
@@ -60,20 +60,26 @@ private Protocol16Handler(TcpClient Client)
...
         {
+            if (((CancellationToken) o!).IsCancellationRequested)
+                return;
+            
             try
             {
-                do
-                {
-                    Thread.Sleep(100);
+                while (!((CancellationToken) o!).IsCancellationRequested) {
```
ReinforceZwei, line 70:
```suggestion
                while (!((CancellationToken) o!).IsCancellationRequested)
                {
```

MinecraftClient/Protocol/Handlers/Protocol16.cs
```diff
@@ -60,20 +60,26 @@ private Protocol16Handler(TcpClient Client)
...
+                    do {
```
ReinforceZwei, line 71:
```suggestion
                    do
                    {
```

DenaryDev:
Is it possible to reduce the number of files that are required for work?

explorer_fJDq7lH4Ir

breadbyte:
> Is it possible to reduce the number of files that are required for work?

`dotnet publish -r win-x64 -p:PublishSingleFile=true`
will reduce everything to a single executable and dll file.

image

breadbyte:
It definitely has been quite a while since I was last here, two .NET releases later!
I definitely experienced some burnout on the way back.

I'm currently throwing together a small test server to iron out any issues and update it to .NET 7 (shouldn't take a while).

Also finally implemented the `>` character in my console helper.

ORelio:
I recognize that the amount of work of this PR is indeed exceptional. Great job 😄
It seems like .NET7 is still in preview state, is it necessary to download a preview version of Visual Studio to compile it?

breadbyte:
Visual Studio is not necessary. Just get the SDK installer or binaries [here](https://dotnet.microsoft.com/en-us/download/dotnet/7.0) and run `dotnet build MinecraftClient.sln` on a terminal. The output should be in `MinecraftClient/bin/net7.0/MinecraftClient.exe`.

Don't forget to run `git submodule update --init` before building, to pull the InteractiveConsole library.
I've been contemplating throwing the library to nuget, but I kept it as a submodule for now so I can throw updates on it faster.

ORelio:
Looks like Offline Mode is not behaving properly (tries to authenticate against session servers).

image

Could not debug this at the moment since my Visual Studio supports .NET6 at most 🤔

breadbyte:
I can't seem to replicate this. I'll bring the version down to .NET 6 so you can potentially debug what's going on.

ReinforceZwei:
@ORelio Would it be better to start a new git branch for .NET 6 or whatever and keep an old branch for .Net Framework 4 (current version)?

ORelio:
Given your dedication to this pull request, the quality of your work, and your motivation, I'm adding you as a maintainer so you could finish polishing the changes and merge them into the repository 🙂

I'd suggest to "archive" the existing code to an alternate branch, provide .NET 5+ releases as default in the master branch (so that contributors and testers would use this version in priority) and keep .NET Framework 4 releases on alternate branch as fallback while ironing out the remaining issues, so that people are not blocked immediately by the new requirements.

breadbyte:
Reviewed the code again and everything LGTM 👍

MinecraftClient/Program.cs
```diff
@@ -401,12 +399,13 @@ private static void InitializeClient()
         /// Disconnect the current client from the server and restart it
         /// </summary>
         /// <param name="delaySeconds">Optional delay, in seconds, before restarting</param>
-        public static void Restart(int delaySeconds = 0)
+        public static void  Restart(int delaySeconds = 0)
```
ORelio, line 402:
```suggestion
        public static void Restart(int delaySeconds = 0)
```

ORelio:
Just saw the above item in last 3 commits, apart from that, looks good to me.

MinecraftClient/Program.cs
breadbyte: I thought I removed this space, but somehow this wasn't a problem in the first place? 
That's what I get for writing code at 3 am 🥴

breadbyte:
I've cloned the `master` branch to a new branch named `previous`. The branch can always be renamed if desired.

Squash merged the PR to `master`. mmm. squash.

breadbyte:
Oops. Need to change the Github Actions builder to .NET too.

ORelio:
Indeed. I'll move the existing action to the legacy branch and look into creating a new one.

Good job! 👏

PR #1677, opened 2021-07-11
Excavate cubes through a command

Daenges:
The purpose of this command is to mimic the "/fill" Minecraft vanilla command. You give the client two corners of a cube and it starts excavating it (also in survival).
Although this sounds pretty simple, there are some problems which make the current status of this script "WIP, buggy but usable". (Most of the time it does its job.)
It can only mine from top to bottom and easily fail mining blocks due to wrong coordinates from GetCurrentLocation() and the fact, that the MoveTo() function does not allow approximation to blocks. It either walks, or fails, although getting in range of this block would be enough. (Not to forget all the bugs that might be created through my own coding skills ;D)

This project could be an essential step to the Baritone bot for minecraft, which has a similar feature named "sel".
I am looking foreward to get feedback and improvements. :)


ReinforceZwei:
I thing it would be better to use `ChatBot` API to implement your function and register a command to control it.

ORelio:
Looks nice! As suggested by @ReinforceZwei a command should not block the thread, because it may just hang the client. It would be best to move that code to a ChatBot class so that it can run in parallel of the rest. 😉

Daenges:
Where can I get the current McClient instance from? In the commands it is a parameter, but where can I find obtain it from the Chatbot class?

Daenges:
Features that could be added:

- [ ] Better pathfinding (maybe with approximation)
- [ ] Reversing order if given cube is the wrong way around
- [ ] Combination with inventory handling (Pick the right tools for the right block)

I am open for more suggestions ;)

Daenges:
@ReinforceZwei @ORelio Okay, I added two functions `getCorrectToolForBlock(Material block)` and `selectCorrectSlotInHotbar(ItemType[] tools)` which should be able to evaluate the right tool for each block. 
(This was a LOT of  work o.o)
I am not sure whether these functions belong in this script, since they make it way too long and are more general purpose. Do you know any better place for them or a better way to solve this issue?

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+        {
+            get { return rowsToMine; }
+        }
+
+        /// <summary>
+        /// Add a new row to this layer
+        /// </summary>
+        /// <param name="givenRow"> enter a row that should be added </param>
+        /// <returns> Index of the last row </returns>
+        public int addRow(Row givenRow = null)
```
ReinforceZwei, line 48: Please use `PascalCase` naming style for C# method.
```suggestion
        public int AddRow(Row givenRow = null)
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+        {
+            get { return layersToMine; }
+        }
+
+        /// <summary>
+        /// Add a new layer to the cube
+        /// </summary>
+        /// <param name="givenLayer"> Enter a layer that should be added </param>
+        /// <returns> Index of the last layer </returns>
+        public int addLayer(Layer givenLayer = null)
```
ReinforceZwei, line 81:
```suggestion
        public int AddLayer(Layer givenLayer = null)
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+    }
+
+    class MineCube : ChatBot
+    {
+        public override void Initialize()
+        {
+            if (!GetTerrainEnabled())
+            {
+                LogToConsole(Translations.Get("extra.terrainandmovement_required"));
+                UnLoadBot(this);
```
ReinforceZwei, line 104: Use this instead of `UnLoadBot(this)`
https://github.com/ORelio/Minecraft-Console-Client/blob/982f358bf37b69ac1d99887887d71fe86226a2fb/MinecraftClient/ChatBot.cs#L836-L839

Also remember to `return` after calling `UnloadBot()` or the `RegisterChatBOtCommand()` below would be called.
```suggestion
                UnloadBot();
                return;
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+            return cubeToMine;
+        }
+
+        /// <summary>
+        /// Get all numbers between from and to.
+        /// </summary>
+        /// <param name="start">Number to start</param>
+        /// <param name="end">Number to stop</param>
+        /// <returns>All numbers between the first, including the stop number</returns>
+        public List<int> getNumbersFromTo(int start, int stop)
```
ReinforceZwei, line 214:
```suggestion
        public List<int> GetNumbersFromTo(int start, int stop)
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+            {
+                for (int i = start; i >= stop; i--)
+                {
+                    tempList.Add(i);
+                }
+            }
+            return tempList;
+        }
+
+        public Func<Location, Location> getHeadLocation = locFeet => new Location(locFeet.X, locFeet.Y + 1, locFeet.Z);
```
ReinforceZwei, line 234:
```suggestion
        public Func<Location, Location> GetHeadLocation = locFeet => new Location(locFeet.X, locFeet.Y + 1, locFeet.Z);
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+
+        /// <summary>
+        /// Checks whether a material is minable
+        /// </summary>
+        /// <param name="block">Block that should be checked</param>
+        /// <returns>Is block minable</returns>
+        private bool isMinable(Material block)
```
ReinforceZwei, line 241:
```suggestion
        private bool IsMinable(Material block)
```

***

You may want to move this method to [MaterialExtensions.cs](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Mapping/MaterialExtensions.cs)

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+        {
+            return (
+                block != Material.Air &&
+                block != Material.Bedrock &&
+                !block.IsLiquid()
+                    );
+        }
+
+        private void selectCorrectSlotInHotbar(ItemType[] tools)
```
ReinforceZwei, line 250:
```suggestion
        private void SelectCorrectSlotInHotbar(ItemType[] tools)
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Evaluates the right tool for the job
+        /// </summary>
+        /// <param name="block">Enter the Material of a block</param>
+        /// <returns>The right tool cathegory as string</returns>
+        private ItemType[] getCorrectToolForBlock(Material block)
```
ReinforceZwei, line 271:
```suggestion
        private ItemType[] GetCorrectToolForBlock(Material block)
```

***

You may want to move this method to [MaterialExtensions.cs](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Mapping/MaterialExtensions.cs)

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+                        block.ToString().Contains("Warped"))
+                 {
+                     return new ItemType[] { ItemType.NetheriteAxe, ItemType.DiamondAxe, ItemType.IronAxe, ItemType.GoldenAxe, ItemType.StoneAxe, ItemType.WoodenAxe };
+                 }
+            else if (shovel.Contains(block) || block.ToString().Contains("ConcretePowder"))
+            {
+                return new ItemType[] { ItemType.NetheriteShovel, ItemType.DiamondShovel, ItemType.IronShovel, ItemType.GoldenShovel, ItemType.StoneShovel, ItemType.WoodenShovel};
+            }
+          
+            return null;
```
ReinforceZwei, line 446: I think it would be better to return an empty array here instead of null to prevent null reference.

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+        }
+
+        /// <summary>
+        /// Prints a whole cube to the console. Separated in layers and rows.
+        /// </summary>
+        /// <param name="cubeToPrint">Some cube</param>
+        private void printCubeToConsole(Cube cubeToPrint)
```
ReinforceZwei, line 453:
```suggestion
        private void PrintCubeToConsole(Cube cubeToPrint)
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,517 @@
...
+                    {
+                        generatedRow += loc.ToString() + "; ";
+                    }
+                    LogToConsole(generatedRow);
+                }
+            }
+            LogToConsole("End of cube.");
+        }
+
+        private string evaluateCommand(string command, string[] args)
```
ReinforceZwei, line 472:
```suggestion
        private string EvaluateCommand(string command, string[] args)
```

MinecraftClient/McClient.cs
```diff
@@ -216,6 +216,7 @@ private void StartClient(string user, string uuid, string sessionID, string serv
 
                     //Add your ChatBot here by uncommenting and adapting
                     //BotLoad(new ChatBots.YourBot());
+                    BotLoad(new MineCube());
```
ReinforceZwei, line 219: Remember to remove this line after you have done your work.

ReinforceZwei (changes requested):
Thanks for your amazing work. I have made an rough review mainly related to coding style.

Daenges:
@ReinforceZwei Okay, so I did two things. First I improved some comments and the naming of public variables. Second I removed the workaround in the tool selection, where I converted the item to a string to avoid writing down all the different colors. Well... I now added all items to all tools. (Was again alot of work.) Nevertheless, this makes this pretty simple script very very long. Any idea on where to move it? I think other scripts would also profit from it.

MinecraftClient/McClient.cs
Daenges: I think I will make this a script in the config folder later. It is not necessary to implement a full bot to the client.

Daenges:
@ReinforceZwei @ORelio 
Okay, I implemented an intelligent tool switching, which works for all tools in the hotbar.

-  While implementing this I noticed that the inventory slots for the hotbar are labeled from 36 - 44 but the `ChangeSlot()` function only takes numbers from 0 - 8. I solved this by just subtracting 36 from the value. Nevertheless, this behaviour seemed strange to me.

Furthermore, I implemented a position switching, if the cube would be upsite down.

- Just checking for the Y value of the coordinates and switch them if needed.

I put the material-tool assertion in a seperate .cs file and made a class out of it. If you have suggestions in which folder I should move this file, I am open for it!

- At the moment it is struggeling from relative long loading times, since it needs to search through several (partly very long) lists, just to find a single block. Any suggestions appreciated. :)

Last but not least I found out, that the walking often fails on blocks that should normally be walkable to. I am not sure what (besides position blocking) can prevent the walking process, but I will work on that.

ReinforceZwei:
Sorry for not reviewing your PR. I am having lots of work to do recently.

> I put the material-tool assertion in a seperate .cs file and made a class out of it. If you have suggestions in which folder I should move this file, I am open for it!

I think you can move `Material2Tool.cs` into `Mapping` folder.

> While implementing this I noticed that the inventory slots for the hotbar are labeled from 36 - 44 but the ChangeSlot() function only takes numbers from 0 - 8.

Hotbar slots are 36 - 44 only if the inventory type is player. And for changing hotbar slot at protocol level, the slot are marked as 0 - 8. [wiki.vg](https://wiki.vg/Protocol#Held_Item_Change_.28serverbound.29)

Daenges:
@ReinforceZwei @ORelio I am currently desparating. It is VERY frustrating to work with the terrain and movements API.
Sometimes blocktypes are detected completely wrong and sometimes the player position is not accurate. This leads to strange bugs, which half of the time make the script fail completely. How am I supposed to avoid breaking blocks which could result in the bot getting harmed, if all the blocks around it are detected as air, eventhough sand is directly above it?
On top of that the move() function sometimes fails to walk to locations which are 5 blocks away and perfectly reachable.
Do you have an Idea on how to fix this? Do I have a bug in my script?

Daenges:
I was able to implement the new command. Now the bot is also able to mine from bottom to top partially. :)

Daenges:
@ReinforceZwei I think I am out of big ideas for this bot (maybe small additions will come). Could you maybe review the script again?
Then I can make sure that the major part is flawless.

MinecraftClient/Mapping/Material2Tool.cs
```diff
@@ -0,0 +1,706 @@
...
+                Material.OrangeWool,
+                Material.PinkWool,
+                Material.PurpleWool,
+                Material.RedWool,
+                Material.SpruceLeaves,
+                Material.WhiteWool,
+                Material.YellowWool,
+        });
+        // Every block that is mined with a sword.
+        private List<Material> sword = new List<Material>(new Material[]
```
ReinforceZwei, line 538: Missing `readonly` here.
```suggestion
        private readonly List<Material> sword = new List<Material>(new Material[]
```

MinecraftClient/Mapping/Material2Tool.cs
```diff
@@ -0,0 +1,706 @@
+﻿using MinecraftClient.Inventory;
+using MinecraftClient.Mapping;
+using System.Collections.Generic;
+
+namespace MinecraftClient.Mapping
+{
+    class Material2Tool
```
ReinforceZwei, line 7: Would be better to mark this class as `static`.
```suggestion
    static class Material2Tool
```
And change everything within the class to `static` as well.

MinecraftClient/Mapping/Material2Tool.cs
```diff
@@ -0,0 +1,706 @@
...
+    {
+        // Made with the following ressources: https://minecraft.fandom.com/wiki/Breaking
+        // Sorted in alphabetical order.
+        // Minable by Any Pickaxe.
+        private readonly List<Material> pickaxe_tier0 = new List<Material>(new Material[]
```
ReinforceZwei, line 12: As a example,
```suggestion
        private static readonly List<Material> pickaxe_tier0 = new List<Material>(new Material[]
```

MinecraftClient/Mapping/Material2Tool.cs
```diff
@@ -0,0 +1,706 @@
...
+                Material.StructureBlock,
+                Material.StructureVoid
+        });
+
+        /// <summary>
+        /// Evaluates the right tool for the job
+        /// </summary>
+        /// <param name="block">Enter the Material of a block</param>
+        /// <returns>Returns a list of tools that can be used, best to worst</returns>
+        public ItemType[] GetCorrectToolForBlock(Material block)
```
ReinforceZwei, line 597:
```suggestion
        public static ItemType[] GetCorrectToolForBlock(Material block)
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        /// <summary>
+        /// Dig out a 2 Block high cube and let the bot walk through it
+        /// mining all blocks above it that it can reach.
+        /// </summary>
+        /// <param name="walkingArea">Area that the bot should walk through. (The lower Y coordinate of the 2 high cube.)</param>
+        public void MineUp(Cube walkingArea)
+        {
+            Material2Tool m2t = new Material2Tool();
```
ReinforceZwei, line 30: Please read my feedback about static class below first.

As the class is static, you do not need to create a class instance before using the class.

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,340 @@
...
+            }
+            LogToConsole("End of cube.");
+        }
+
+        private string EvaluateMineUpCommand(string command, string[] args)
+        {
+            Location startBlock = new Location(
+                        double.Parse(args[0]),
+                        double.Parse(args[1]),
+                        double.Parse(args[2])
```
ReinforceZwei, line 262: It is a good practice to validate user input before using them. You may take a look at how `AutoTree` does it.
https://github.com/ORelio/Minecraft-Console-Client/blob/349a3de8e4a5fb5d2af3fe9b74aa700e2de2a1a8/MinecraftClient/config/ChatBots/AutoTree.cs#L142-L144
https://github.com/ORelio/Minecraft-Console-Client/blob/349a3de8e4a5fb5d2af3fe9b74aa700e2de2a1a8/MinecraftClient/config/ChatBots/AutoTree.cs#L146-L158

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,340 @@
...
+                        );
+
+            Location stopBlock = new Location(
+                    double.Parse(args[3]),
+                    double.Parse(args[4]),
+                    double.Parse(args[5])
```
ReinforceZwei, line 268: Same as above.

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,340 @@
...
+                    );
+
+            if (Math.Round(startBlock.Y) != Math.Round(stopBlock.Y))
+            {
+                return "Both blocks must have the same Y value!";
+            }
+
+            CubeFromWorld CFW = new CubeFromWorld();
+            List<Material> materialWhitelist = new List<Material>(new Material[] { Material.Air });
```
ReinforceZwei, line 277: To populate values to a list,
```suggestion
            List<Material> materialWhitelist = new List<Material>() 
            {
                Material.Air
            };
```

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,340 @@
...
+                    startBlock = new Location(
+                    double.Parse(args[0]),
+                    double.Parse(args[1]),
+                    double.Parse(args[2])
+                    );
+
+                    stopBlock = new Location(
+                    double.Parse(args[3]),
+                    double.Parse(args[4]),
+                    double.Parse(args[5])
```
ReinforceZwei, line 302: Validate user input before using them. You may add a `try/catch` block to catch invalid input.

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        /// <summary>
+        /// Initialize a row of blocks
+        /// </summary>
+        /// <param name="bIL"> Enter a list of blocks </param>
+        public Row(List<Location> bIL = null)
+        {
+            blocksInRow = bIL ?? new List<Location>();
+        }
```
ReinforceZwei, line 25: Maybe use full name here for better readability?
```suggestion
        public Row(List<Location> blocksInRow = null)
        {
            this.blocksInRow = blocksInRow ?? new List<Location>();
        }
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+        /// Initialize a cube
+        /// </summary>
+        /// <param name="lTM"> Enter a list of layers </param>
+        public Cube(List<Layer> lTM = null)
+        {
+            layersToMine = lTM ?? new List<Layer>();
+        }
+    }
+
+    public class CubeFromWorld
```
ReinforceZwei, line 94:
```suggestion
    public static class CubeFromWorld
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        /// <summary>
+        /// Creates a cube of blocks out of two coordinates.
+        /// </summary>
+        /// <param name="startBlock">Start Location</param>
+        /// <param name="stopBlock">Stop Location</param>
+        /// <returns>A cube of blocks consisting of Layers, Rows and single blocks</returns>
+        public Cube GetBlocksAsCube(World currentWorld, Location startBlock, Location stopBlock, List<Material> materialList = null, bool isBlacklist = true)
```
ReinforceZwei, line 102:
```suggestion
        public static Cube GetBlocksAsCube(World currentWorld, Location startBlock, Location stopBlock, List<Material> materialList = null, bool isBlacklist = true)
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+            return cubeToMine;
+        }
+
+        /// <summary>
+        /// Get all numbers between from and to.
+        /// </summary>
+        /// <param name="start">Number to start</param>
+        /// <param name="end">Number to stop</param>
+        /// <returns>All numbers between the start and stop number, including the stop number</returns>
+        private List<int> GetNumbersFromTo(int start, int stop)
```
ReinforceZwei, line 164:
```suggestion
        private static List<int> GetNumbersFromTo(int start, int stop)
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+using System.Collections.Generic;
+
+namespace MinecraftClient.Mapping
+{
+    /// <summary>
+    /// A row of blocks that will be mined
+    /// </summary>
+    public class Row
+    {
+        private List<Location> blocksInRow;
```
ReinforceZwei, line 11: As the variable here will not be changed once created, you may use `readonly` here and remove the public getter.
```suggestion
        public readonly List<Location> BlocksInRow;
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+            blocksInRow = bIL ?? new List<Location>();
+        }
+    }
+
+    /// <summary>
+    /// Several rows are summarized in a layer
+    /// </summary>
+    public class Layer
+    {
+        private List<Row> rowsToMine;
```
ReinforceZwei, line 33: Same as above.
```suggestion
        public readonly List<Row> RowsToMine;
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+            rowsToMine = rTM ?? new List<Row>();
+        }
+    }
+
+    /// <summary>
+    /// Several layers result in a cube
+    /// </summary>
+    public class Cube
+    {
+        private List<Layer> layersToMine;
```
ReinforceZwei, line 66:
```suggestion
        public readonly List<Layer> LayersToMine;
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+            return layersToMine.Count - 1;
+        }
+
+        /// <summary>
+        /// Initialize a cube
+        /// </summary>
+        /// <param name="lTM"> Enter a list of layers </param>
+        public Cube(List<Layer> lTM = null)
+        {
+            layersToMine = lTM ?? new List<Layer>();
```
ReinforceZwei, line 90:
```suggestion
        public Cube(List<Layer> layersToMine = null)
        {
            LayersToMine = layersToMine ?? new List<Layer>();
```

MinecraftClient/Mapping/CubeFromWorld.cs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        /// <summary>
+        /// Creates a cube of blocks out of two coordinates.
+        /// </summary>
+        /// <param name="startBlock">Start Location</param>
+        /// <param name="stopBlock">Stop Location</param>
+        /// <returns>A cube of blocks consisting of Layers, Rows and single blocks</returns>
+        public Cube GetBlocksAsCube(World currentWorld, Location startBlock, Location stopBlock, List<Material> materialList = null, bool isBlacklist = true)
+        {
+            Material2Tool m2t = new Material2Tool();
```
ReinforceZwei, line 104: You can remove this once you have converted the class to static class.

MinecraftClient/ChatBots/MineCube.cs
```diff
@@ -0,0 +1,340 @@
...
+            LogToConsole("Mining finished.");
+        }
+
+        /// <summary>
+        /// Get all numbers between from and to.
+        /// </summary>
+        /// <param name="start">Number to start</param>
+        /// <param name="end">Number to stop</param>
+        /// <returns>All numbers between the first, including the stop number</returns>
+        public List<int> GetNumbersFromTo(int start, int stop)
```
ReinforceZwei, line 164: Seems like this method is not used anywhere. Maybe delete it?

ReinforceZwei (changes requested):

Daenges:
@ReinforceZwei Okay, I think I have finished this project. If you have any further ideas let me know.

ReinforceZwei:
In your last update, I saw `MinecraftClient/ChatBots/MineCube.cs` duplicated to `MinecraftClient/config/ChatBots/MineCube.cs`, leaving two identical ChatBot. Do you mean to keep the bot at `MinecraftClient/config/ChatBots/` and delete the old one?

Daenges:
@ReinforceZwei Yup, this was my intention, but the whole reformating got a bit chaotic in the end. I hope the last push fixed it. :)

ReinforceZwei (approved):

ReinforceZwei:
Thanks for your contribution. Looks good to me.

PR #1691, opened 2021-07-17
Create bug report GitHub issue form

ReinforceZwei:
This will create new bug report form in `.yaml` format and delete the old template. __This feature is still in beta.__

_Form Preview_
image


ORelio:
Brilliant! Much better than existing template 👍

PR #1700, opened 2021-07-27
Is it possible to add support for Labymod?

1Dilva:
I have no clue if this is possible or not but I wanted to ask if we could add that


Daenges:
Uhm... Is this a general feature request or do you want to code and add Labymod support yourself into the client?

1Dilva:
oh right I didn't know I do code a bit but not skilled enough yet thanks. and yes it's a feature request. thanks.

PR #1709, opened 2021-08-04
Integration of the Mojang API

Daenges:
I wanted to improve the mail script and implement UUID support. Therefore I need the Mojang API to request those.
I thought why not invest a few more minutes and improve the experience for everyone by implementing some request types mentioned in the [wiki](https://wiki.vg/Mojang_API).

I will make a seperate pull request for the mail script later.

### DIFF
--- a/MinecraftClient/MinecraftClient.csproj
+++ b/MinecraftClient/MinecraftClient.csproj
@@ -75,7 +75,7 @@
     <Compile Include="AutoTimeout.cs" />
     <Compile Include="ChatBots\AutoDrop.cs" />
     <Compile Include="ChatBots\Mailer.cs" />
-    <Compile Include="MojangAPI.cs" />
+    <Compile Include="Protocol\MojangAPI.cs" />
     <Compile Include="Mapping\CubeFromWorld.cs" />
     <Compile Include="Mapping\Material2Tool.cs" />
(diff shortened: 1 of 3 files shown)


ORelio:
That's a good idea. How about placing these methods in [ProtocolHandler.cs](https://github.com/ORelio/Minecraft-Console-Client/blob/master/MinecraftClient/Protocol/ProtocolHandler.cs) which already has authentication-related API calls?
Or, alternatively move existing API calls from ProtocolHandler to your new MojangAPI class (both approaches are valid).

Daenges:
@ORelio I moved it to the protocol namespace. I think it is easier to leave the seperate MojangAPI.cs instead of hiding these methods inside the ProtocolHandler.cs. 
I am not sure whether people who want to code Chatbots will look deply into the protocol handler, especially when they have less experience. I mean even I just partly understand what is happening in there.

Apart from this I do not have any idea of which methods to move out of the ProtocolHandler.cs and which files would changes due to this action.

ORelio:
Okay, let's keep it as is. Is it ready to merge?

Daenges:
@ORelio Yup. Everything is ready to merge.

ReinforceZwei:
> Yup. Everything is ready to merge.

I will merge it for you.

PR #1714, opened 2021-08-07
Addition to MojangAPI.cs

Daenges:
It seems like I have overseen something. In the Mojang API there is a way to request a players profile. Between several useless information, there is a Base64 encrypted JSON string, which reveals plenty information about the players skin and look. I implemented this feature, but I was a bit confused when it comes to a certain edge case:

There is a possibility, that a a player has no skin data. According to the [wiki](https://wiki.vg/Mojang_API) this happens when a Player has not set a custom skin. Seemingly this only applies until the first skin change of an account, since I was not able to recreate this, by changing my skin to Steve. If these cases occur, you can still evaluate their skin by checking their UUID, since even UUIDs have a Steve skin and odd ones an Alex skin.

I am not sure whether I implemented the odd/even check correctly, as I was mainly orienting on a JS [source](https://github.com/crafatar/crafatar/blob/9d2fe0c45424de3ebc8e0b10f9446e7d5c3738b2/lib/skins.js#L90-L108) recommended by the wiki.

### DIFF
--- a/MinecraftClient/Protocol/MojangAPI.cs
+++ b/MinecraftClient/Protocol/MojangAPI.cs
@@ -142,5 +142,56 @@ namespace MinecraftClient.Protocol
 
             return tempDict;
         }
+
+        /// <summary>
+        /// Obtain links to skin, skinmodel and cape of a player.
+        /// </summary>
+        /// <param uuid="uuid">UUID of a player</param>
+        /// <returns>Dictionary with a link to the skin and cape of a player.</returns>
+        public static Dictionary<string, string> SkinInfo(string uuid)
+        {
+            Dictionary<string, string> tempDict = new Dictionary<string, string>();
+            Dictionary<string, Json.JSONData> textureDict;
+            string base64SkinInfo;
+            Json.JSONData decodedJsonSkinInfo;
+
+            // Perform web request
+            try
+            {
+                // Obtain the Base64 encoded skin information from the API. Discard the rest, since it can be obtained easier through other requests.
+                base64SkinInfo = Json.ParseJson(wc.DownloadString("https://sessionserver.mojang.com/session/minecraft/profile/" + uuid)).Properties["properties"].DataArray[0].Properties["value"].StringValue;
+            }
+            catch (Exception) { return tempDict; }
+
+            // Parse the decoded string to the JSON format.
+            decodedJsonSkinInfo = Json.ParseJson(System.Text.Encoding.UTF8.GetString(Convert.FromBase64String(base64SkinInfo)));
+
+            // Assert temporary variable for readablity.
+            // Contains skin and cape information.
+            textureDict = decodedJsonSkinInfo.Properties["textures"].Properties;
+
+            // Can apparently be missing, if no custom skin is set.
+            // Probably for completely new accounts.
+            if (textureDict.ContainsKey("SKIN"))
+            {
+                // Add the URL leading to the texture of the ingame skin.
+                tempDict.Add("SkinURL", textureDict["SKIN"].Properties.ContainsKey("url") ? textureDict["SKIN"].Properties["url"].StringValue : string.Empty);
+
+                // Detect whether the playermodel is based on Steve or Alex.
+                // If the skin property contains metadata, wich always contains "slim", it is an Alex based skin.
+                tempDict.Add("PlayerModel", textureDict["SKIN"].Properties.ContainsKey("metadata") ? "Alex" : "Steve");
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


ORelio:
Hi,

Looks good to me.

Regarding the hashcode thing, according to the source file you linked to, you need to compute the hashcode first, and only check at the end if the result is even or not. So no need to apply isEven everytime, only check it at the end.

Seems like you need to generate each int16 by taking pairs of bytes, then xor them together. For instance, the first Int16 would be read from `BitConverter.ToInt16(uuid, 7)` but maybe you need to reverse pairs of byte due to endianness, I'm not sure.
https://github.com/ORelio/Minecraft-Console-Client/blob/39b7d878dd3cc29fc0179346cd4edea62168428c/MinecraftClient/Protocol/GuidExtensions.cs#L10-L33

Hope this help. You can check the java code against yours to see if it computes the same results for a set of sample UUIDs.

Daenges:
@ORelio I do not have much experience with hex values and their behaviour, but I rechecked my code against the Java examples and it worked fine.
I did not quite understand what to change on the odd/even detection. :/
I also oriented myself on [this](https://github.com/LapisBlue/Lapitar/blob/30fee6c71716bdb0bbb1d78632cdf9f95463c7c5/lapitar/mc/uuid.go#L35), which is linked in the comment above the mentioned code example.

ReinforceZwei:
We could follow the implementation of Java `UUID.hashCode()` to get the UUID hash and mod the hash to determine the skin is "Steve" or "Alex".
```cs
int hashCode(string hash) {
    byte[] data = ToLittleEndian(new Guid(hash)).ToByteArray();
    ulong msb = 0;
    ulong lsb = 0;
    for (int i = 0; i < 8; i++)
        msb = (msb << 8) | (uint)(data[i] & 0xff);
    for (int i = 8; i < 16; i++)
        lsb = (lsb << 8) | (uint)(data[i] & 0xff);
    var hilo = msb ^ lsb;
    
    return ((int)(hilo >> 32)) ^ (int) hilo;
}

if (hashCode("5d82b1b37c00466d9f5b18633ee66575") % 2 == 1) 
{
    // Alex
}
else
{
    // Steve
}
```

[Java UUID class](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/default/src/share/classes/java/util/UUID.java)
[The _algorithm_](https://pastebin.com/9B3wmt9J)

Daenges:
@ReinforceZwei I tinkered something together, in the hope that this the correct approach. Also tested it against the control values and it works (but my previous version also passed this test).

MinecraftClient/Protocol/MojangAPI.cs
```diff
@@ -142,5 +143,91 @@ public static string UuidToCurrentName(string uuid)
...
+                // Probably a completely new account.
+                tempDict.Add("SkinURL", string.Empty);
+                tempDict.Add("PlayerModel", DefaultModelAlex(uuid) ? "Alex" : "Steve");
+            }
+
+            // If a cape exists, add it, otherwise leave string empty.
+            tempDict.Add("CapeURL",
+                textureDict.ContainsKey("CAPE") ? textureDict["CAPE"].Properties["url"].StringValue : string.Empty);
+
+            return tempDict;
```
ReinforceZwei, line 199: I think it would better not to return a string dictionary because it would be harder for the user (method caller) to understand what is inside the dictionary. One solution to this problem is to create a class for holding the result and return the class to the user.
For example,
```cs
class SkinInfo
{
    public readonly string SkinUrl;
    ...
}
```
and return it,
```cs
return new SkinInfo()
{
    SkinUrl = theUrl,
    ...
};
```
So that user can know what is inside by looking at the class definition.

The same thing for `GetMojangServiceStatus()` method above.

Daenges:
@ReinforceZwei new return types implemented.

MinecraftClient/Protocol/MojangAPI.cs
```diff
@@ -6,10 +6,56 @@
...
+            SkinModel = skinModel;
+        }
+    }
+
+    /// <summary>
+    /// Status of the single Mojang services
+    /// </summary>
+    public class MojangServiceStatus
+    {
+        public readonly string MinecraftNet;
```
ReinforceZwei, line 37: How about creating an enum to replace string?
```cs
public enum ServiceStatus
{
    Green,
    Yellow,
    Red
}
```

Daenges:
@ReinforceZwei implemented it.

ReinforceZwei:
Thanks! Merging your PR.

PR #1825, opened 2021-11-17
Spectator-Teleport Implementation

xXjojaXx:
I just implemented what I requested in an Issue. Hope I did it right, because this is my first pull request. Feel free to change anything or tell me what should be different, if there is any kind of Enhancement or Problem to take care of. Basically this change allows MCC-Users to use `SpectatorTeleport(Player from GetEntities())` and `SpectatorTeleport(new Guid(String of Player or Entity-UUID))` int their Chatbots.

### DIFF
--- a/MinecraftClient/ChatBots/ScriptScheduler.cs
+++ b/MinecraftClient/ChatBots/ScriptScheduler.cs
@@ -9,7 +9,7 @@ namespace MinecraftClient.ChatBots
     /// <summary>
     /// Trigger scripts on specific events
     /// </summary>
-    
+
     public class ScriptScheduler : ChatBot
     {
         private class TaskDesc
--- a/MinecraftClient/ChatBots/ScriptScheduler.cs
+++ b/MinecraftClient/ChatBots/ScriptScheduler.cs
@@ -9,6 +9,7 @@ namespace MinecraftClient.ChatBots
     /// <summary>
     /// Trigger scripts on specific events
     /// </summary>
+    
     public class ScriptScheduler : ChatBot
     {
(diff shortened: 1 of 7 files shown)


MinecraftClient/ChatBots/ScriptScheduler.cs
```diff
@@ -9,7 +9,7 @@ namespace MinecraftClient.ChatBots
     /// <summary>
     /// Trigger scripts on specific events
     /// </summary>
-
+
```
ReinforceZwei, line 12: Remove the indentation here.

ReinforceZwei (approved):
Thank you for your contribution! The work is perfect for a first time contribution. Just remove the extra spaces and you are good to go.

xXjojaXx:
Short question about making pull request cleaner: can I delete commits of an pr to undo their changes and not having to do another commit to undo? Also thanks for reacting so quick!

ReinforceZwei:
You may reset work directory or rebase branch and push again. But deleting commit is like rewriting the history. It is not recommended when you are working on the same branch with others.

It's okay to leave the commits here as is. Github can combine multiple commits into single commit before merging into master branch.

ORelio:
Good job! Nice contribution 👍

Just in case, for the sake of compatibility between Minecraft versions did you check on wiki.vg what is the minimum version supporting the Spectator teleport and made sure the packet has not changed between version?

xXjojaXx:
The packet was introduced in 1.8 and has remained unchanged since then.

ORelio:
OK, added a version check 😉

Issue #1830, opened 2021-11-22
Any way to make a random delay on the script scheduler?


throwaway4206:
I searched the readme and tried some stuff myself, but I can't find anyway to make a random delay using the script scheduler. Is this not a feature?

ORelio:
Indeed that's not possible at the moment, but it's a good idea. Could be implemented like AutoRelog with delay expressed as a `min-max` interval. Converting your issue to a enhancement request 👍

milutinke:
I'll be looking into adding this in the following days.

PR #1846, opened 2021-12-10
Add setrnd command

Daenges:
Hello there! As #1830 requested a random delay feature for the script scheduler, I had an idea how to implement it.
Therefore I added the "setrnd" command. As the name suggests it is pretty similar to the already existing "set" command.
The only difference is the syntax, you use it as the following:

| Command | Effect |
|---|---|
|`setrnd varname -10to10` | This sets the variable to a value between -10 and 10.|
|`setrnd varname string1,string2,string3,string4...` | This sets the variable value to one of the strings. <br> You can use as many comma seperated strings as you want.|

Due to the fact, that we are using local variables here, it can be used in all kinds of scripts. An example:
```
send Hello server.
setrnd waittime 50to100
wait %waittime%
send Waited: %waittime%
setrnd name ORelio,ReinforceZwei,Daenges
send Hello %name%, how are you?
```

### DIFF
--- a/MinecraftClient/Commands/SetRnd.cs
+++ b/MinecraftClient/Commands/SetRnd.cs
@@ -30,7 +30,7 @@ namespace MinecraftClient.Commands
                         }
                         catch (Exception)
                         {
-                            return Translations.Get("cmd.setrnd.format");
+                            return Translations.Get("cmd.setrndnum.format");
                         }
 
                         if (num2 < num1)
@@ -44,18 +44,18 @@ namespace MinecraftClient.Commands
                         {
                             return string.Format("Set %{0}% to {1}.", args[0], Settings.GetVar(args[0])); //Success
                         }
-                        else return Translations.Get("cmd.setrnd.format");
+                        else return Translations.Get("cmd.setrndnum.format");
                     }
                     else
                     {
                         string argString = command.Substring(command.IndexOf(args[0]) + args[0].Length, command.Length - 8 - args[0].Length);
                         List<string> values = parseCommandLine(argString);
 
-                        if (Settings.SetVar(args[0], values[rand.Next(0, values.Count)]))
+                        if (values.Count > 0 && Settings.SetVar(args[0], values[rand.Next(0, values.Count)]))
                         {
                             return string.Format("Set %{0}% to {1}.", args[0], Settings.GetVar(args[0])); //Success
                         }
-                        else return Translations.Get("cmd.setrnd.format");
+                        else return Translations.Get("cmd.setrndstr.format");
                     }
                 }
                 else return GetCmdDescTranslated();
--- a/MinecraftClient/Commands/SetRnd.cs
+++ b/MinecraftClient/Commands/SetRnd.cs
@@ -20,16 +20,21 @@ namespace MinecraftClient.Commands
 
                 if (args[0] == "player")
                 {
+                    // remember to use no further arguments
                     if (args.Length > 1)
                         return Translations.Get("cmd.setrndplayer.format");
 
-                    var test = handler.GetUsername();
+                    // get all playernames without the bot itself
                     string[] playernames = handler.GetOnlinePlayers().Where(name => name != handler.GetUsername()).ToArray();
+
+                    // if there are players apart from the bot online
                     if (playernames.Length > 0)
                     {
+                        // get a random one and assign it
                         Settings.SetVar("player", playernames[rand.Next(0, playernames.Length)]);
                         return string.Format("Set %{0}% to {1}.", "player", Settings.GetVar("player"));
                     }
+                    // you are the only one on the server
                     else { return Translations.Get("cmd.setrndplayer.lonely"); }
                 }
                 else if (args.Length > 1)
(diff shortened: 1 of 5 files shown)


ORelio:
Hi @Daenges and thanks for looking into this!

Regarding ScriptScheduler delay, I'd think about specifying a range like for AutoRelog:
https://github.com/MCCTeam/Minecraft-Console-Client/blob/4ba09754de04a2d87fe921be6e7dd4b445edf2ad/MinecraftClient/Settings.cs#L537-L548
You could then specify the task like this:
```ini
[Task]
triggerOnInterval=true
timeInterval=30-40
action=move up
```
This syntax would be backward compatible with existing tasks.

Regarding the setrnd command, this looks like a good addition for scripting, so why not! Things like random commands or so on would be facilitated with this command. The syntax looks a bit complicated though, I'm not 100% sure what would be the best way to implement it: both easy to use and the least restrictive (how about specifying a comma in one of the random strings? can't because comma is used as a separator). Let's think a bit about if, and if no better idea comes up, let's keep as is 🤔

Apart from that, the command looks well implemented 👍

Daenges:
Regarding the strings we could use quotes and include escaping. So something like this:

`setrnd varname "string1" "string2,string3" "\""` 

This would be a bit more complicated regarding the syntax but would provide the maximal freedom of character choice. We could then just run a function over it:

<details>
<summary>Expand code</summary>

```
static List<string> getArgsInQuotes(string quotestring)
        {
            bool copy = false;
            string tempStr = "";
            List<string> listOfArgs = new List<string>();

            for(int c = 0; c < quotestring.Length; c++)
            {
                if (quotestring[c] == '\\' && quotestring[c + 1] == '"')
                {
                    tempStr += "\"";
                    c += 2;

                    if (c == quotestring.Length - 1)
                        listOfArgs.Add(tempStr);

                    continue;
                }
                else if(quotestring[c] == '"' && (c == 0 || quotestring[c-1] != '\\'))
                {
                    copy = !copy;

                    if (!copy)
                    {
                        listOfArgs.Add(tempStr);
                        tempStr = "";
                    }
                }
                else if (copy)
                {
                    tempStr += quotestring[c];
                }
            }

            return listOfArgs;
        }
```

</details>

This would extract all text within quotes and even allows escaping, so using quotes as character.
I have the feeling that this is the **most practical**, but **not really user friendliest** way, since it could result in a mess if you try to explain this system to someone who never touched a programming language and only wants 3 different options to show up.
You mentioned nothing about the integer declaration.. So I am assuming you are fine with it?

Daenges:
Regarding the ScriptScheduler I would suggest to open a seperate PR later, so we can focus on one topic at a time and do not mess stuff up. ;)

ReinforceZwei:
Regarding the multiple random strings assignment, how about separating them with space? And for strings containing spaces, user must wrap the string with double quote. I think this would be easier to understand as Windows command prompt and other linux terminals uses this behavior.

Daenges:
@ReinforceZwei implemented your idea in the new commit. :)

New usage:
`setrnd variable string1 string2 "\"string3\" string4"`

Daenges:
@ReinforceZwei @ORelio Any further requests?

ReinforceZwei (dismissed):

Daenges:
Oh, I have an additional idea.
I want to add a `player` constant which is always set to a random playername on the server, excluding the bot itself. I'll add this later today.

Daenges:
Okay found some time and implemented the feature. :)

ReinforceZwei:
I'm not sure picking random player is a good idea. I've seen some kiddies spamming random people on the server. Perhaps you could explain more about the usage of this feature?

Daenges:
@ReinforceZwei Well... you are right there. This feature might be abused.
My idea was that it would be a feature some people might want and it seemed pretty easy to implement.
But even if I remove the feature it is as easy as doing a `/list`, remove the commas and do a `setrnd player name1 name2...`.
I can see your point, that this implementation is more automated and therefore more accessible for kiddies who try to fool around with it.
It was an idea of mine for what people might need. If you think that the damage might outweigh the benefits, I will remove it again.

Daenges:
If I should remove it, what do you think about just commenting this part of the code? Then the people who want to fool around with it will most likely not even notice the implementation, since they just download the compiled version and the people who might need it can go into the code, uncomment and compile it. This assumes that only people who are able to understand C# can implement it again and if they understand C# they could literally code it in themselves.

ReinforceZwei:
Remove the code would be better.

Daenges:
@ReinforceZwei Okay. Withdrew changes. Ready for review again.

ReinforceZwei (approved):

ORelio:
Thanks for this great contribution! Maybe `parseCommandLine()` could be moved to Command.cs?

ReinforceZwei:
> Thanks for this great contribution! Maybe `parseCommandLine()` could be moved to Command.cs?

Moved in https://github.com/MCCTeam/Minecraft-Console-Client/commit/e68a51dcff9a16939bdb81633290f99ff2b51e76.

PR #1855, opened 2021-12-17
Added docker-compose configuration

Archer57:
This pull request adds a docker-compose.yml for managing docker containers via docker-compose. Also the user manual has been extended to describe how to use docker-compose on this project.

### DIFF
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -6,4 +6,4 @@ services:
     container_name: 'MinecraftConsoleClient'
     # You could also comment out the volumes section if you don't want to mount the volume. That is up to you. If you do so it's harder to edit the .ini file if thats something you want to do.
     volumes:
-    - '<PATH_ON_YOUR_MACHINE_TO_MOUNT>:/opt/data'
+    - './data:/opt/data'
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -1,0 +1,9 @@
+version: '2'
+
+services:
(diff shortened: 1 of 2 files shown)


Docker/docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '2'
+
+services:
+  MCC:
+    image: 'minecraft-console-client:latest'
+    container_name: 'MinecraftConsoleClient'
+    # You could also comment out the volumes section if you don't want to mount the volume. That is up to you. If you do so it's harder to edit the .ini file if thats something you want to do.
+    volumes:
+    - '<PATH_ON_YOUR_MACHINE_TO_MOUNT>:/opt/data'
```
ReinforceZwei, line 9: How about filling relative path here?
```suggestion
    - './data:/opt/data'
```

Archer57: This would work as well. I used it in my example in the user manual as well. The only point with this is that a folder is created in the user's cloned repository. But it is fine for me to do it this way. Shall I change it and maybe clean up the user manual as well? Then the user has to change nothing by default.

ReinforceZwei: Yes, please.

ReinforceZwei (approved):

ReinforceZwei:
Thanks for your contribution again. Merging your work.

Archer57:
> Thanks for your contribution again. Merging your work.

Happy to contribute to this cool project. :)

Issue #1872, opened 2021-12-27
i have a error


shadowstar1109:
### Prerequisites

- [X] I made sure I am running the latest [development build](https://ci.appveyor.com/project/ORelio/minecraft-console-client/build/artifacts)
- [X] I tried to [look for similar issues](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one
- [X] I have set `debugmessages=true` in config to diagnose my issue
- [X] I have redacted session tokens and passwords before attaching screenshots

### Minecraft Version

1.16.5

### Console Client Version

ultimate

### Expected Behavior

.

### Actual Behavior

I have a error in a command /move i change the versions : 1.16.5, 1.8 and more but the error is the same 
Failed to compute path to X:201 Y:10 Z:70 (for example) can i help me i try all 

### Steps to Reproduce the bug

1. 
2. 
3. 

### Device

Laptop

### Operating System

Windows

Daenges:
Well there could be several reasons for this error to appear. 

1. The most likely one would be that you entered a position that is blocked by blocks or just generally unreachable. Remember the bot is currently unable to mine/set blocks. It only calculates a path, that is not blocked by anything and walks it off.

2. Your goal is too far away. The path finding algorithm has a timeout of 5 seconds, if nothing is found you get an error message. This includes coordinates that are too far away, for instance if your goal is out of render distance.

shadowstar1109:
It is certainly unattainable, I think because I want him to throw himself into a void and I don't know why he doesn't want to

Daenges:
Well.. The bot has a protection which prevents the execution of potentially lethal walking commands. You maybe need to find another way to kill it.

shadowstar1109:
> Bueno ... El bot tiene una protección que evita la ejecución de comandos para caminar potencialmente letales. Quizás necesites encontrar otra forma de matarlo.

I do not want it to die is the entry a server. you can help?

Daenges:
Your question is very difficult to understand. Do you want to know how you can join a server?

shadowstar1109:
> Tu pregunta es muy difícil de entender. ¿Quieres saber cómo puedes unirte a un servidor?

srry i use translate when you enter the server you need to enter and put the password and then throw yourself into a void the server is mc.srolemine.com

Daenges:
[...]

PR #1874, opened 2021-12-27
Add "-f" to the /move command

Daenges:
As a solution for #1872 I have implemented a `-f` argument, which forces the movement regardless of its potential consequences.

### DIFF
--- a/MinecraftClient/Commands/Move.cs
+++ b/MinecraftClient/Commands/Move.cs
@@ -50,8 +50,9 @@ namespace MinecraftClient.Commands
                     }
                     if (Movement.CanMove(handler.GetWorld(), handler.GetCurrentLocation(), direction))
                     {
-                        return handler.MoveTo(Movement.Move(handler.GetCurrentLocation(), direction), allowUnsafe: takeRisk) ? 
-                            Translations.Get("cmd.move.moving", args[0]) : Translations.Get("cmd.move.suggestforce");
+                        if (handler.MoveTo(Movement.Move(handler.GetCurrentLocation(), direction), allowUnsafe: takeRisk))
+                            return Translations.Get("cmd.move.moving", args[0]);
+                        else return !takeRisk ? Translations.Get("cmd.move.suggestforce") : Translations.Get("cmd.move.dir_fail");
                     }
                     else return Translations.Get("cmd.move.dir_fail");
                 }
@@ -64,8 +65,9 @@ namespace MinecraftClient.Commands
                         int z = int.Parse(args[2]);
                         Location goal = new Location(x, y, z);
 
-                        return handler.MoveTo(goal, allowUnsafe: takeRisk) ?
-                           Translations.Get("cmd.move.walk", goal) : Translations.Get("cmd.move.suggestforce");
+                        if (handler.MoveTo(goal, allowUnsafe: takeRisk))
+                            return Translations.Get("cmd.move.walk", goal);
+                        else return !takeRisk ? Translations.Get("cmd.move.suggestforce") : Translations.Get("cmd.move.fail", goal);
                     }
                     catch (FormatException) { return GetCmdDescTranslated(); }
                 }
--- a/MinecraftClient/Commands/Move.cs
+++ b/MinecraftClient/Commands/Move.cs
@@ -1,7 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using MinecraftClient.Mapping;
 
 namespace MinecraftClient.Commands
@@ -14,25 +13,24 @@ namespace MinecraftClient.Commands
 
         public override string Run(McClient handler, string command, Dictionary<string, object> localVars)
         {
-            string[] args = getArgs(command);
-            string argStr = getArg(command).Trim().ToLower();
+            string[] args = getArgs(command.ToLower());
 
-            if (argStr == "on")
+            if (args[0] == "on")
             {
                 handler.SetTerrainEnabled(true);
                 return Translations.Get("cmd.move.enable");
             }
-            else if (argStr == "off")
+            else if (args[0] == "off")
             {
                 handler.SetTerrainEnabled(false);
                 return Translations.Get("cmd.move.disable");
             }
             else if (handler.GetTerrainEnabled())
             {
-                if (args.Length == 1)
+                if (args.Length == 1 || (args.Length == 2 && args.Contains("-f")))
                 {
                     Direction direction;
-                    switch (argStr)
+                    switch (args[0])
                     {
                         case "up": direction = Direction.Up; break;
                         case "down": direction = Direction.Down; break;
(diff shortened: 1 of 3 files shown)


ORelio (changes requested):
Looks good! However, please document that the switch exists in command help or error message when path is unsafe, otherwise nobody will know about it.

Daenges:
@ORelio Changes applied.

MinecraftClient/Commands/Move.cs
```diff
@@ -1,38 +1,36 @@
...
 using MinecraftClient.Mapping;
 
 namespace MinecraftClient.Commands
 {
     public class Move : Command
     {
         public override string CmdName { get { return "move"; } }
-        public override string CmdUsage { get { return "move <on|off|get|up|down|east|west|north|south|x y z>"; } }
-        public override string CmdDesc { get { return "walk or start walking."; } }
+        public override string CmdUsage { get { return "move <on|off|get|up (-f)|down (-f)|east (-f)|west (-f)|north (-f)|south (-f)|x y z>"; } }
```
ReinforceZwei, line 11: I think the help message here became too complex and hard to read. 

How about:
```suggestion
        public override string CmdUsage { get { return "move <on|off|get|up|down|east|west|north|south|x y z> [-f]"; } }
```

MinecraftClient/Commands/Move.cs
```diff
@@ -1,38 +1,36 @@
...
+        public override string CmdDesc { get { return "walk or start walking. Included protection can be bypassed with \"-f\"."; } }
```
ReinforceZwei, line 12:
```suggestion
        public override string CmdDesc { get { return "walk or start walking. Unsafe path protection can be bypassed with \"-f\"."; } }
```

MinecraftClient/Commands/Move.cs
```diff
@@ -1,38 +1,36 @@
...
 
         public override string Run(McClient handler, string command, Dictionary<string, object> localVars)
         {
-            string[] args = getArgs(command);
-            string argStr = getArg(command).Trim().ToLower();
+            string[] args = getArgs(command.ToLower());
```
ReinforceZwei, line 16: I think the parsing of the arguments could be done like below:
https://github.com/MCCTeam/Minecraft-Console-Client/blob/259ef80cf97b18a55da0f4b810460e1e49e1711a/MinecraftClient/Program.cs#L63-L71

We look for `-f` in the whole argument string, if found, set a boolean flag to true and remove the `-f` from the argument string. Then the position of `-f` will no longer be important, and could be easily apply the flag to moving with absolute coordinate.

ReinforceZwei (commented):
How about also applying the `force` option for moving to absolute coordinate?

Daenges:
> How about also applying the `-f` option for moving with absolute coordinate?

@ReinforceZwei 
Also thought about that. The only problem here is, that the absolute movement consists of several individual movement desicions. I do not think that players can actually estimate what happens, if the client goes on full risk on every movement. This can result in major damage, since the behaviour of the client on the way can result in unpredictable dangerous situations. (Especially problematic since you can not stop the movement of the client without disconnecting.)

My idea is that the player lets the bot autowalk **near** a dangerous situation and then switches to direct control. Thereby we force the user to be attentive and to be responsible of their own decisions.

Daenges:
Despite my doubts I implemented the changes. Lets pray for the peoples common sense.

ReinforceZwei:
That could be a problem for those who don't pay attention to the situation. Maybe forcing people to type the full `-force` argument could raise some attention? 

Never mind if you don't like unleashing unsafe path for absolute coordinate. Feel free to revert it back.

Daenges:
I thought about git and how it also allows the -f command without any further barricades.

As long as people use the regular command, they get warned by a message and everything further they do is on their own responsibility.

Why restrict their freedom if they know what they want to do? Some need to touch the fire to know that it is hot.

MinecraftClient/Commands/Move.cs
```diff
@@ -1,38 +1,43 @@
...
 
 namespace MinecraftClient.Commands
 {
     public class Move : Command
     {
         public override string CmdName { get { return "move"; } }
-        public override string CmdUsage { get { return "move <on|off|get|up|down|east|west|north|south|x y z>"; } }
-        public override string CmdDesc { get { return "walk or start walking."; } }
+        public override string CmdUsage { get { return "move <on|off|get|up|down|east|west|north|south|x y z> [-f]"; } }
+        public override string CmdDesc { get { return "walk or start walking. Included protection can be bypassed with \"-f\"."; } }
```
ORelio, line 12:
```suggestion
        public override string CmdDesc { get { return "walk or start walking. \"-f\": force unsafe movements like falling or touching fire"; } }
```

MinecraftClient/Commands/Move.cs
```diff
@@ -41,26 +46,28 @@ public override string Run(McClient handler, string command, Dictionary<string,
...
                         int x = int.Parse(args[0]);
                         int y = int.Parse(args[1]);
                         int z = int.Parse(args[2]);
                         Location goal = new Location(x, y, z);
-                        if (handler.MoveTo(goal))
+
+                        if (handler.MoveTo(goal, allowUnsafe: takeRisk))
                             return Translations.Get("cmd.move.walk", goal);
-                        return Translations.Get("cmd.move.fail", goal);
+                        else return !takeRisk ? Translations.Get("cmd.move.suggestforce") : Translations.Get("cmd.move.fail", goal);
```
ORelio, line 70: How about removing the negation in ternary condition and also include coordinates:
```suggestion
                        else return takeRisk ? Translations.Get("cmd.move.fail", goal) : Translations.Get("cmd.move.suggestforce", goal);
```
(see below for coordinates in message)

MinecraftClient/Resources/lang/en.ini
```diff
@@ -310,6 +310,7 @@ cmd.move.moving=Moving {0}
 cmd.move.dir_fail=Cannot move in that direction.
 cmd.move.walk=Walking to {0}
 cmd.move.fail=Failed to compute path to {0}
+cmd.move.suggestforce=The implemented protection prevents a movement in this direction. Use the -f parameter to force the movement.
```
ORelio, line 313: The path either does not exists _or_ is unsafe. So the message could look like this instead:
```suggestion
cmd.move.suggestforce=Failed to compute a safe path to {0}. Try -f parameter to allow unsafe movements.
```

MinecraftClient/Resources/lang/de.ini
```diff
@@ -310,6 +310,7 @@ cmd.move.moving=Laufe {0}
 cmd.move.dir_fail=Kann nicht in diese Richtung laufen.
 cmd.move.walk=Gehe nach {0}
 cmd.move.fail=Konnte Pfad nach {0} nicht berechnen.
+cmd.move.suggestforce=Die implementierte Absicherung verhindert eine Bewegung in diese Richtung. Du kannst sie mit dem -f Parameter erzwingen.
```
ORelio, line 313: I cannot make suggestions in German but see the English version below.

ORelio (commented):
Looks good to me! Here are a few additional suggestions mainly to improve wording/readability:

ORelio (approved):

ORelio:
Good job! I'll leave @ReinforceZwei review again just in case and it should be good to merge 👍

MinecraftClient/Commands/Move.cs
```diff
@@ -41,26 +46,28 @@ public override string Run(McClient handler, string command, Dictionary<string,
...
-                        default: return Translations.Get("cmd.look.unknown", argStr);
+                        default: return Translations.Get("cmd.look.unknown", args[0]);
                     }
                     if (Movement.CanMove(handler.GetWorld(), handler.GetCurrentLocation(), direction))
                     {
-                        handler.MoveTo(Movement.Move(handler.GetCurrentLocation(), direction));
-                        return Translations.Get("cmd.move.moving", argStr);
+                        if (handler.MoveTo(Movement.Move(handler.GetCurrentLocation(), direction), allowUnsafe: takeRisk))
+                            return Translations.Get("cmd.move.moving", args[0]);
+                        else return !takeRisk ? Translations.Get("cmd.move.suggestforce") : Translations.Get("cmd.move.dir_fail");
```
ReinforceZwei, line 55:
```suggestion
                        else return takeRisk ? Translations.Get("cmd.move.dir_fail") : Translations.Get("cmd.move.suggestforce");
```

PR #1914, opened 2022-01-23
Fix /move crash and /setrnd

Daenges:
If you enter only `/move` the client crashes, since there is no first element [here](https://github.com/MCCTeam/Minecraft-Console-Client/blob/master/MinecraftClient/Commands/Move.cs#L25).


ORelio:
Thanks for the fix! 👍

PR #1915, opened 2022-01-23
Fix quote detection in /setrnd

Daenges:
Before the function crashed when entering quoted strings.


ORelio:
Why copy/paste the [existing method](https://github.com/MCCTeam/Minecraft-Console-Client/blob/e68a51dcff9a16939bdb81633290f99ff2b51e76/MinecraftClient/Command.cs#L100)?

Daenges:
@ORelio Whooops! I have not noticed, that the function was moved. I changed the first [if statement](https://github.com/MCCTeam/Minecraft-Console-Client/pull/1915/files#diff-87515f269ec8661c184297264d9ffc652fee75d8318d3bdf02985790cd4c89b4R91), which fails with an `Out of array` exception, if the first character is a quote, since it tries to find out, whether the character before it (that does not exist) is an escape.

ORelio:
I see, that's fine. Thanks for fixing that 😉

Issue #1508, opened 2021-03-20
Implement flight disabling


Vort:
Some server owners allow players to fly.
And also enable fly on login.

Since ore mining works bad while flying, it is good to have an option to disable fly mode.
Here is my attempt to implement this feature: https://github.com/ORelio/Minecraft-Console-Client/commit/b50f96e3968a0b727b2f8dafcd3a6826dd8315a1.

It is not tested well enough and may have a problems.
Also it is based on outdated fork.
Because of this I create not pull request, but issue.

ORelio:
Hi @Vort, this looks nice 👍

To make this future proof I think you should also abstract the flags in Player abilities because they may change between versions.

PR #1955, opened 2022-03-01
Option and Command to Disable Gravity

mvoppen:
Currently gravity is always active. But in real play there are many situations where no gravity is active. 

This PR creates an option to activate/deactivate gravity and extends the move command by two subcommands to activate/deactivate gravity.


ReinforceZwei (approved):

ReinforceZwei:
Looks good to me. Thanks for your work. 

@ORelio are you okay with this PR?

ORelio:
@ReinforceZwei Yes! Missed that PR, sorry.
@mvoppen Thanks for your contribution which can help on servers where flying is enabled.
I've adjusted the /move command compared to how you originally implemented it and added french translation as well.

I see two things that could be improved, if you have enough time:

1. Ideally, instead of a setting, gravity handling should be sourced from player capabilities (see #1508 and https://wiki.vg/Protocol#Player_Abilities_.28clientbound.29). The server sends a packet to the client to tell whether flying is enabled. MCC could read that packet and automatically (dis)able gravity handling accordingly.

2. Pathfinding would need to be adjusted as well to take flying into account. This can be done in CanMove() to allow moving up even when not on ground, IsSafe() to allow falling, and GetAvailableMoves() to allow normal movement in the air (enter the first if{ statement when gravity is not enabled). That should do the job but might require some more changes during testing.

Of course if you're busy I'm okay with the changes as is, so the pull request can be merged 😉

ORelio (approved):

mvoppen:
There are two components in that flying topic. The server decides if the client is allowed to fly and the client decides if the player actually is flying. So anyway we need an internal flag or a setting if gravity is enabled. The server cannot decide that allone.

There are additional topics not implemented yet like swimming and floating water.

So my approach was to start it simple and have the option to disable gravity.

ORelio:
Okay, let's keep it as is.

PR #1943, opened 2022-02-13
Implemented Entity, Inventory, Block palettes and Terrain Handling for 1.17/.1 and 1.18/.1/.2 (Tested and working)

milutinke:
Greetings,

I've done the following:

- Implemented Block, Item and Entity palettes for 1.17/.1 and 1.18/.1
- Added all Material/Item ID-s to utility classes
- Fixed a minor warning in `AttackBot`
- Fixed 2 packets related to Inventory handling (more on that bellow)
- Added an option to generate the Item, Block and Entity palettes using the command line: `--generate`, `--generator`, `--data-path` (Example: `MinecraftConsoleClient.exe --generate --generator=entity --data-path="C:\users\me\resources.json"`)

**NOTE: I've not done any detailed test, just a general test, we need further testing and some more changes related to inventory handling (1.18)**

Entity Handling Test:
[https://streamable.com/kz7lzx](https://streamable.com/kz7lzx)

Auto Tree Test:
[https://streamable.com/w3qkmq](https://streamable.com/w3qkmq)

In 1.18, there have been changes to player inventory packets. You now get a `State ID` int, which is supposed to be sent back to the server when manipulating the inventory items. I'll leave you to implement the best way for that state storing and sending.
You will see place holders, they just get the value so packets work.

One more thing, the `AutoAttack` bot does not register when entity dies, so it keeps attacking a dead entity forever.
My idea is to use the `EntitySound` packet to get the sound of death, and then use it to remove the entity from the dictionary.

PS: It would be best if this was moved to a separate branch, so it gets tested until it has been confirmed as stable.

PR #1961, opened 2022-03-06
Update version to 1.18.2

heuerleon:


Pokechu22:
This code also needs to be updated:

https://github.com/MCCTeam/Minecraft-Console-Client/blob/288994aeec7b18093d204aafad699bbc0cdf7b88/MinecraftClient/Protocol/Handlers/Protocol18.cs#L936-L949

https://pokechu22.github.io/Burger/diff_1.18.1_1.18.2.html - the effect ID is now a VarInt.  Not sure if 1.18 has entity handling currently, but it'd be best to update that now so it's not missed.  (The same applies to the remove packet, but that doesn't seem to be implemented currently?)

ORelio:
Thanks @Pokechu22 for the tip!
Maybe we should merge #1943 or #1951 first, then this one, then make the change for effect ID.
Otherwise there will be merge conflicts 😄

ORelio:
Merged while waiting for other changes.

PR #1981, opened 2022-03-23
Read EntityEffect from int instead of byte

heuerleon:
In the 1.18.2 protocol, the effect id is no longer sent as byte, but as integer. For more information, see https://pokechu22.github.io/Burger/diff_1.18.1_1.18.2.html#packets:play_clientbound_65 (Thanks to @Pokechu22)

Entity handling is currently not handled for the new version, so these changes are advance only.


ReinforceZwei:
I think implementing protocol changes version by version (1.16.5 -> 1.17 -> ... -> 1.18) would be better than skipping versions.

PR #1999, opened 2022-04-20
Add check to determine moving and improve CalculatePath() to allow approaching.

Daenges:
In previous projects ([SugarcaneFarmer](https://github.com/MCCTeam/Minecraft-Console-Client/blob/master/MinecraftClient/config/ChatBots/SugarCaneFarmer.cs#L15), [MineCube](https://github.com/MCCTeam/Minecraft-Console-Client/blob/master/MinecraftClient/config/ChatBots/MineCube.cs#L39)) I needed to check if the client has arrived at its walking destination before I could continue. I used to do that by checking the current distance to the goal and continue if it was below a certain threshold, which is a bit vague. This is why I read into the movement process and made this check to determine exactly, when the movement is over.


Daenges:
I additionally wrote a function to approach a certain block. This might be useful, if you want to get as close to some coordinates as possible, but do not know anything about the landscape.

MinecraftClient/McClient.cs
```diff
@@ -1084,6 +1085,102 @@ public bool MoveTo(Location location, bool allowUnsafe = false, bool allowDirect
...
+            // and sort them after their distance to goalToApproach. Take 5 by default to avoid too much calculation power.
+            List<Location> listOfCloseLocations = listOfBlocks.Where(block => GetWorld().GetBlock(new Location(block.X, block.Y + 1, block.Z)).Type == Material.Air &&      // AIR
+                                                                            GetWorld().GetBlock(block).Type == Material.Air &&                                              // AIR
+                                                                            GetWorld().GetBlock(new Location(block.X, block.Y - 1, block.Z)).Type.IsSolid())                // SOLID
+                                                                .OrderBy(closeLocation => goalToApproach.Distance(closeLocation)).ToList();
+
+            // If a minimal distance to the goal was given, remove all locations that are too close.
+            if (minDistance>0) 
+            {
+                listOfCloseLocations = listOfCloseLocations.Where(block => block.Distance(goalToApproach) >= minDistance).ToList();
```
ORelio, line 1120: Note that Distance does `sqrt()`, which is slow. `radius` being already a cube, maybe you can use `DistanceSquared`. Multiply `minDistance` by itself once: `minDistanceSquared = minDistance * minDistance`, then compare with `DistanceSquared`, distance will be a cube instead of a sphere. (Also I'm not sure why you'd want a minimum distance. Maybe for cactus? But `Movement.IsSafe()` should detect it)

MinecraftClient/McClient.cs
```diff
@@ -1084,6 +1085,102 @@ public bool MoveTo(Location location, bool allowUnsafe = false, bool allowDirect
...
+                            listOfPathfindInstances.Add(Task<Queue<Location>>.Run(() => Movement.CalculatePath(world, this.location, closeLoc, allowUnsafe)));
+
+                            // If a valid path was found, stop spawning new tasks and exit
+                            if (ct.IsCancellationRequested) { break; }
+                        }
+                    }
+                }, cts.Token);
+
+                // Search through the list of tasks for valid paths
+                // List is sorted from the closest possbile goal to the farthest
```
ORelio, line 1159:
```suggestion
                // List is sorted from the closest possible goal to the farthest

```

MinecraftClient/McClient.cs
```diff
@@ -1084,6 +1085,102 @@ public bool MoveTo(Location location, bool allowUnsafe = false, bool allowDirect
...
+                        foreach (Location closeLoc in listOfCloseLocations)
+                        {
+                            // Wait if there are too many parallel tasks
+                            concurrencySemaphore.Wait();
+
+                            // Create a Task for one path
+                            listOfPathfindInstances.Add(Task<Queue<Location>>.Run(() => Movement.CalculatePath(world, this.location, closeLoc, allowUnsafe)));
+
+                            // If a valid path was found, stop spawning new tasks and exit
+                            if (ct.IsCancellationRequested) { break; }
```
ORelio, line 1153: If exiting early, maybe the method will end up selecting the first valid path, which happens to be the farthest one in the searched radius? To avoid that while still retaining performance (i.e. ability to stop early), how about sorting candidates first by `distanceSquared()` with the destination (which should be relatively fast to compute), then start calculating paths? So stopping early should still yield a good (if not best) result.

MinecraftClient/McClient.cs
```diff
@@ -1084,6 +1085,102 @@ public bool MoveTo(Location location, bool allowUnsafe = false, bool allowDirect
...
+                        }
+                    }
+                }, cts.Token);
+
+                // Search through the list of tasks for valid paths
+                // List is sorted from the closest possbile goal to the farthest
+                int i = 0;
+                while(!pathfindSpawner.IsCompleted || i < listOfPathfindInstances.Count)
+                {
+                    if (listOfPathfindInstances.Count > 0)
```
ORelio, line 1163: I think the while() loop is made to check task while they are being added to the list, so the first condition `i < count` could be moved in place of the first test.
```suggestion
                while (!pathfindSpawner.IsCompleted)
                {
                    if (i < listOfPathfindInstances.Count)

```

MinecraftClient/McClient.cs
```diff
@@ -1084,6 +1085,102 @@ public bool MoveTo(Location location, bool allowUnsafe = false, bool allowDirect
...
+                        // If there is a valid path
+                        if (listOfPathfindInstances[i].Result != null)
+                        {
+                            // Stop creating new tasks
+                            cts.Cancel();
+                            path = listOfPathfindInstances[i].Result;
+                            return true;
+                        }
+                        else { i++; }
+                    }
```
ORelio, line 1175: When `pathfindSpawner` is not complete but `i >= listOfPathfindInstances.Count`, this mean the loop must wait for the next task to be added to the list. A tiny Sleep() should save some CPU cycles, right?
```suggestion
                    } else { Thread.Sleep(10); }
```

MinecraftClient/ChatBot.cs
```diff
@@ -990,6 +990,20 @@ protected bool MoveToLocation(Mapping.Location location, bool allowUnsafe = fals
             return Handler.MoveTo(location, allowUnsafe, allowDirectTeleport);
         }
 
+        protected bool ApproachTo(Mapping.Location location, int radius, int maxNumberOfLocationsToEvaluate=5, int minDistance=0, bool allowUnsafe = false) 
+        {
+            return Handler.ApproachTo(location, radius, maxNumberOfLocationsToEvaluate, minDistance, allowUnsafe);
```
ORelio, line 995: `maxNumberOfLocationsToEvaluate`: This parameter seems absent in `Handler.ApproachTo()` below? Instead, there is a `pathfindingThreads` parameter which is also an `int`.

ORelio (changes requested):
Thanks for contributing! The `ApproachTo()` method looks promising!
Please find my feedback below:

Daenges:
@ORelio Thank you for your feedback. I will implement this later. Currently I am looking for an optimal way to handle the asynchronous pathfinding instances. I tried to work around changing calculatePathFunction() but this gets increasingly difficult. Therefore I am looking to add cancellationToken handling to the function, so the function will be used as a task by default.

ORelio:
Actually I'm not sure ApproachTo() is thread-safe because the mapping stuff itself is not. When the client receives terrain data, it will modify the terrain data, but one of your threads might be trying to read it at the same time. Since neither ApproachTo() or CalculatePath() are _writing_ to the terrain data, the worst case would be an exception thrown in these methods. Some semaphore or locks would be needed to ensure thread safety, but of course it becomes tricky because terrain data is a complex structure.

Daenges:
This might cause most of the trouble in MineCube.... Solving this sounds very difficult.

ORelio:
On second thought, ensuring thread safety to terrain data should not be that difficult. Adding some [ReadWriterLockSlim](https://docs.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim) should be enough. This way unlimited threads can read terrain data at the same time, but main thread can gain exclusive access to the exact part it needs to write to (e.g. lock a single chunk, lock a chunk column, etc). The other thread might continue working on outdated data (e.g. keep working on an old chunk when a whole chunk has been replaced) but at least nothing will crash. Just committed that.

Daenges:
@ORelio so in the current state the pathcalculation for instance with the /move command fails, although I set the timeout higher than 5 seconds.

Daenges:
The current version of calculate path seems to work. So it must be an issue on my side.

ORelio:
If I'm correct, the goal of `ApproachTo()` is to find a safe location near a block you want to get close to. But isn't this a variant of the A* algorithm already implemented in `CalculatePath()`? Taking a look at the example at Wikipedia:
Astar_progress_animation

(Source: [A* search algorithm](https://en.wikipedia.org/wiki/A*_search_algorithm#Example))

A* tries to calculate a patch towards the destination. But if it gets stuck _near_ the destination, actually, we could stop there and retain that location instead of considering it's not possible to reach the destination. That would be a variant of `CalculatePath()`. For instance, by introducing an additional parameter to tell the algorithm that locations within a n-block range are also acceptable. Considering this, tweaking `CalculatePath()` would give something quicker to compute _and_ also easier to implement.

Do you think that makes sense?

Daenges:
@ORelio this is a good idea. I do not think that an `n-range` would solve our issue, since it would return the first result in range, which would be the farthest block, so the worst result.
[This Post](https://gamedev.stackexchange.com/a/35254) suggests, that we just keep track of the closest block to the goal, so the block with the lowest `h-cost`. Getting it should be as easy as:
```
Location closestGoal = ClosedSet.OrderBy(pair => pair.DistanceSquared(goal)).First();
```
Then we need to get the reverse path by entering `closestGoal` into the `Came_From` Dict, which should then return a valid path to our approximate location.

Only problem is: When should we stop the Algorithm? The only thing that comes to my mind is again a cancellation token with a timeout. Then we add a parameter to the function like `allowApproachIfGoalNotReached`, which determines if `null` or an approach path is returned on fail.

Daenges:
Short question: Am I missing something or why aren't we breaking out of the while loop [here](https://github.com/MCCTeam/Minecraft-Console-Client/blob/master/MinecraftClient/Mapping/Movement.cs#L165)? We found a way to the goal, why should we continue until the open set is empty? Even the Wikipedia source returns at this point. :o

ORelio:
> I do not think that an `n-range` would solve our issue, since it would return the first result in range, which would be the farthest block from the goal, so the worst result.
[This Post](https://gamedev.stackexchange.com/a/35254) suggests, that we just keep track of the closest block to the goal, so the block with the lowest `h-cost`. Getting it should be as easy as:
```
Location closestGoal = ClosedSet.OrderBy(checkedLocation => checkedLocation.DistanceSquared(goal)).First();
```

To get a minimum range (i.e. do not get closer than n-block distance of goal), just consider goal is reached once close enough.
To get a maximum range (i.e. consider that closer than n-block distance is ok), yes, that approach seems good 👍

> Only problem is: When should we stop the Algorithm? The only thing that comes to my mind is again a cancellation token with a timeout. Then we add a parameter to the function like `allowApproachIfGoalNotReached`, which determines if `null` or an approach path is returned on fail.

There's already a timeout so I'd say to stop when it's elapsed, maybe?

> Short question: Am I missing something or why aren't we breaking out of the while loop [here](https://github.com/MCCTeam/Minecraft-Console-Client/blob/master/MinecraftClient/Mapping/Movement.cs#L165)? We found a way to the goal, why should we continue until the open set is empty? Even the Wikipedia source returns at this point. :o

According to the [pseudocode](https://en.wikipedia.org/wiki/A*_search_algorithm#Pseudocode) the algorithm should return here, so try adding a `break` statement, maybe I missed something back when I implemented the algorithm...

Daenges:
> There's already a timeout so I'd say to stop when it's elapsed, maybe?

Yes, but not using `AutoTimeout.Perfom` but cancellation tokens. If we now have thread safety, why not using it? :P

Daenges:
> According to the [pseudocode](https://en.wikipedia.org/wiki/A*_search_algorithm#Pseudocode) the algorithm should return here, so try adding a break statement, maybe I missed something back when I implemented the algorithm...

Holy cow.. this thing is fast now! :o

ORelio:
> Yes, but not using `AutoTimeout.Perfom` but cancellation tokens. If we now have thread safety, why not using it? :P

AutoTimeout already starts a thread, but block the calling thread while waiting for the child thread to complete. I think cancellation tokens would result in the same thing. The benefit from having thread safety on terrain would be the ability for bots to do pathfinding in the background, and get a callback once result is available.

> > According to the [pseudocode](https://en.wikipedia.org/wiki/A*_search_algorithm#Pseudocode) the algorithm should return here, so try adding a break statement, maybe I missed something back when I implemented the algorithm...
> 
> Holy cow.. this thing is fast now! :o

I guess that the code was running until timeout regardless of result found or not. You nailed a nice bug 😅👏

Daenges:
> AutoTimeout already starts a thread, but block the calling thread while waiting for the child thread to complete. I think cancellation tokens would result in the same thing.

Now that the function is pretty fast, it is less of a problem. (We can even lower the timeout a bit.) But the advantage with Cancellation Tokens is, that the ChatBot developer can decide HOW LONG, WHEN and even IF to wait for the function to finish and meanwhile do something else. If you think that AutoTimeout is a better solution, I can implement it that way. But why restrict the developers freedom? (Beside the fact that the code gets a little more complicated, but you can always look in the `/move command`, which provides a basic example.)

Daenges:
So my idea is to pass the `path` variable, which holds the currently walked path, as `ref` parameter. Then we can make the thing completely async. Thereby we are theoretically even able to get rid of a timeout, since the client does not stop while calculating and just starts walking when the calculation has finished.

ORelio:
> But the advantage with Cancellation Tokens is, that the ChatBot developer can decide HOW LONG, WHEN and even IF to wait for the function to finish [...] why restrict the developers freedom?

Ah, yes, if the token is passed as argument, not defined locally, it's more flexible indeed. I've nothing against that as long as you make sure the existing `MoveToLocation` ChatBot API stays available for compatibility purposes (so you can add an optional parameter, a variant or a different method, but the existing version must stay available). For instance, a version that runs async with a callback would likely come handy.

> we are theoretically even able to get rid of a timeout

If you don't put a timeout, the algorithm will not stop evaluating possible paths until it processed every reachable block in all loaded chunks, which could take a long time. I think there should be one, at least by default. If a custom ChatBot wants to launch a long background processing and disable timeout, why not.

Daenges:
Okay. I implemented all changes.

> If you don't put a timeout, the algorithm will not stop evaluating possible paths until it processed every reachable block in all loaded chunks, which could take a long time. I think there should be one, at least by default. If a custom ChatBot wants to launch a long background processing and disable timeout, why not.

Completely agree. I also had problems with the command framework, which always expects a return, so in `/move` we force it to wait, but through the structure, we theoretically allow developers who want to use it async, to do so.

ORelio:
>I played around with the new CalculatePath() and made a small script that lets the client follow a specific player.

Looks nice! Of course it can be improved like registering an internal command, adding ability to stop following, but it's more than enough for testing.

> How big is MCCs render distance?

It's the server render distance as the server sends chunks based on its own render distance regardless of what is set client-side. The client reports a render distance to the server that you can set in config file but unless some plugin uses that value, it has no effect as far as I know.

> It might be a good Idea to split long walks into smaller pieces

I'd suggest to calculate path with large (infinite?) acceptable distance when the location is outside loaded chunks, then keep the retained location and walk there, then recalculate at a later time once more chunks are loaded. Should not be _that_ difficult to implement, a ChatBot could do this as long as CalculatePath() offers that as an option.

> But currently we should focus on the small issues. It does not help much if we can theoretically walk up to infinity, but the client gets stuck on the edge of a grass block after 10 seconds. :P

Indeed 🤣

MinecraftClient/ChatBot.cs
```diff
@@ -984,12 +984,23 @@ protected Mapping.Location GetCurrentLocation()
         /// <param name="location">Location to reach</param>
         /// <param name="allowUnsafe">Allow possible but unsafe locations thay may hurt the player: lava, cactus...</param>
         /// <param name="allowDirectTeleport">Allow non-vanilla teleport instead of computing path, but may cause invalid moves and/or trigger anti-cheat plugins</param>
+        /// <param name="allowApproachIfGoalNotReached">If no valid path can be found return a path that approaches your goal</param>
```
ORelio, line 987: How about replacing the boolean with a parameter that states maximum offset instead?
```suggestion
        /// <param name="maxOffset">If no valid path can be found, also allow locations within specified distance of destination</param>
```
That should be helpful in the case discussed before where we want to walk to a destination that is far away and chunks are not loaded yet. This way it's possible to say e.g. `3` for going near some block to break, `Int.MaxValue` for calculating path until edge of chunk.

MinecraftClient/ChatBot.cs
```diff
@@ -984,12 +984,23 @@ protected Mapping.Location GetCurrentLocation()
...
+        /// <param name="mindistance">A minimal distance you want to keep to your goal</param>
```
ORelio, line 988: If implementing `maxOffset`, I'd suggest changing this parameter to `minOffset`.
```suggestion
        /// <param name="minOffset">Do not get closer of destination than specified distance.</param>
```
Note that `maxOffset` must be `>= minOffset` so an ArgumentException may be used here.

MinecraftClient/McClient.cs
```diff
@@ -215,7 +216,7 @@ private void StartClient(string user, string uuid, string sessionID, string serv
                     if (Settings.ReplayMod_Enabled) { BotLoad(new ReplayCapture(Settings.ReplayMod_BackupInterval)); }
 
                     //Add your ChatBot here by uncommenting and adapting
-                    //BotLoad(new ChatBots.YourBot());
+                    // BotLoad(new ChatBots.YourBot());
```
ORelio, line 219: Just to avoid unnecessary change in PR:
```suggestion
                    //BotLoad(new ChatBots.YourBot());

```

ORelio (commented):
Before making a full review again, based on the current discussion:

Daenges:
@ORelio So if you have no further improvements, this is ready for merge.

MinecraftClient/ChatBot.cs
```diff
@@ -983,13 +983,25 @@ protected Mapping.Location GetCurrentLocation()
         /// </summary>
         /// <param name="location">Location to reach</param>
         /// <param name="allowUnsafe">Allow possible but unsafe locations thay may hurt the player: lava, cactus...</param>
-        /// <param name="allowDirectTeleport">Allow non-vanilla teleport instead of computing path, but may cause invalid moves and/or trigger anti-cheat plugins</param>
+        /// <param name="allowDirectTeleport">Allow non-vanilla direct teleport instead of computing path, but may cause invalid moves and/or trigger anti-cheat plugins</param>
+        /// <param name="maxOffset">If no valid path can be found, also allow locations within specified distance of destination</param>
+        /// <param name="minOffset">Do not get closer of destination than specified distance</param>
+        /// <param name="timeoutInSec">How long to wait until the path is evaluated</param>
```
ORelio, line 989: How about using a `TimeSpan` for this argument? I think this might be more flexible.

MinecraftClient/Mapping/Movement.cs
```diff
@@ -129,62 +129,89 @@ public static Queue<Location> Move2Steps(Location start, Location goal, ref doub
         /// <param name="start">Start location</param>
         /// <param name="goal">Destination location</param>
         /// <param name="allowUnsafe">Allow possible but unsafe locations</param>
+        /// <param name="maxOffset">If no valid path can be found, also allow locations within specified distance of destination</param>
+        /// <param name="minOffset">Do not get closer of destination than specified distance</param>
         /// <returns>A list of locations, or null if calculation failed</returns>
-        public static Queue<Location> CalculatePath(World world, Location start, Location goal, bool allowUnsafe = false)
+        public static Queue<Location> CalculatePath(World world, Location start, Location goal, CancellationToken ct, bool allowUnsafe=false, int maxOffset=0, int minOffset=0)
```
ORelio, line 135: The `ct` parameter is not documented.

MinecraftClient/Mapping/Movement.cs
```diff
@@ -129,62 +129,89 @@ public static Queue<Location> Move2Steps(Location start, Location goal, ref doub
...
         {
-            Queue<Location> result = null;
 
-            AutoTimeout.Perform(() =>
+            if (minOffset > maxOffset)
+                throw new ArgumentException("maxOffset must be >= minOffset");
```
ORelio, line 139: I'd suggest to reverse the argument exception since `minOffset` is specified after `maxOffset`.
```suggestion
            if (minOffset > maxOffset)
                throw new ArgumentException("minOffset must be lower than or equal to maxOffset", "minOffset");
```
Also changed to the constructor that explicitly states the name of the affected parameter.

MinecraftClient/Mapping/Movement.cs
```diff
@@ -129,62 +129,89 @@ public static Queue<Location> Move2Steps(Location start, Location goal, ref doub
...
+                if (maxOffset > 0 && ClosedSet.Count > 0)
+                    closestGoal = ClosedSet.OrderBy(checkedLocation => checkedLocation.DistanceSquared(goal)).First();
 
-                Dictionary<Location, int> g_score = new Dictionary<Location, int>(); //:= map with default value of Infinity
-                g_score[start] = 0; // Cost from start along best known path.
-                // Estimated total cost from start to goal through y.
-                Dictionary<Location, int> f_score = new Dictionary<Location, int>(); //:= map with default value of Infinity
-                f_score[start] = (int)start.DistanceSquared(goal); //heuristic_cost_estimate(start, goal)
+                // Stop when goal is reached or we are close enough
+                if (current == goal || (current.DistanceSquared(goal) <= minOffset && minOffset > 0))
```
ORelio, line 166: Skip calculating `DistanceSquared()` if not needed:
```suggestion
                if (current == goal || ( minOffset > 0 && current.DistanceSquared(goal) <= minOffset))
```

MinecraftClient/Mapping/Movement.cs
```diff
@@ -129,62 +129,89 @@ public static Queue<Location> Move2Steps(Location start, Location goal, ref doub
...
+                    Came_From[neighbor] = current;
+                    g_score[neighbor] = tentative_g_score;
+                    f_score[neighbor] = g_score[neighbor] + (int)neighbor.DistanceSquared(goal); //heuristic_cost_estimate(neighbor, goal)
                 }
-            }, TimeSpan.FromSeconds(5));
+            }
 
-            return result;
+            // Could not be reached. Set the path to the closest possible location
+            if (goal.DistanceSquared(closestGoal) <= maxOffset || maxOffset == int.MaxValue)
```
ORelio, line 193: Skip calculating distance if offset is already `int.MaxValue`
```suggestion
            // Goal could not be reached. Set the path to the closest location if close enough
            if (maxOffset == int.MaxValue || goal.DistanceSquared(closestGoal) <= maxOffset)            
```

MinecraftClient/Mapping/Movement.cs
```diff
@@ -129,62 +129,89 @@ public static Queue<Location> Move2Steps(Location start, Location goal, ref doub
...
-            return result;
+            // Could not be reached. Set the path to the closest possible location
+            if (goal.DistanceSquared(closestGoal) <= maxOffset || maxOffset == int.MaxValue)            
+                return reconstructPath(Came_From, closestGoal);
+            else
+                return null;
+        }
+
+        /// <summary>
+        /// Helper function for CalculatePath()
```
ORelio, line 200:
```suggestion
        /// Helper function for CalculatePath(). Backtrack from goal to start to reconstruct a step-by-step path.
```

MinecraftClient/Mapping/Movement.cs
```diff
@@ -129,62 +129,89 @@ public static Queue<Location> Move2Steps(Location start, Location goal, ref doub
...
+        /// </summary>
+        /// <param name="Came_From">The collection of Locations that leads back to the start</param>
+        /// <param name="current">Endpoint of our later walk</param>
+        /// <returns>the path that leads to current from the start position</returns>
+        private static Queue<Location> reconstructPath(Dictionary<Location, Location> Came_From, Location current)
```
ORelio, line 205: Should be renamed to `ReconstructPath` for consistency.

ORelio (commented):
This is getting really neat! I've only got a few minor suggestions left:

Daenges:
So some follow up ideas:
- `/move`
    - [ ] Implement all new parameters (approaching/timeout)
    - [ ] With `ClientIsMoving()` we could improve the user experience: So something like `/move info` to check from terminal if and where the client is moving
    - [ ] On top of that a `/move cancel` so `path=null` would come handy
- [ ] `SugercaneFarmer` and `MineCube` (especially last) could need a rework now, since they are mainly based on the pathfinding function
- [ ] new `/follow` command to follow entities through the world, which might be easier than manually entering walking commands
- further improvement to the pathfinding
  - [ ] larger distances (for goals far away: calculate path to renderdistance; wait while walking there; repeat until goal is in RD)
  - general improvement:
    - [ ] Client often gets stuck on block edges
    - [ ] Client is unable to walk diagonally

ORelio:
Just made a few additional changes to move code to `Movement.cs` and fix missing methods in .NET 4.0 (`Task.Run` and `CancellationTokenSource.CancelAfter` were not defined). I'll let you check if the changes are OK with you, if so, everything seems ready for merging.

Of course, further improvements can be made as you pointed out, but I think this can be addressed in a further pull request (unless you prefer to put everything now in the same PR).

ORelio (approved):

Daenges:
>  .NET 4.0 (Task.Run and CancellationTokenSource.CancelAfter were not defined)
Yeah.. VS basically forced me to upgrade the project locally to 4.8 :P

I am fine with the changes. This PR is already a bit bloated with the additional `ClientIsMoving()` check. New project, new PR.
And I don't want the people to wait for a properly working `/move` command. ;D

ORelio:
Thanks again for your work which will greatly allow the pathfinding capabilities of MCC 🥇

PR #2013, opened 2022-05-03
Square minOffset and maxOffset to match DistanceSquared

Daenges:
We need to square the parameters to be able to actually compare the distance, if we are not willing to generate the root.

### DIFF
--- a/MinecraftClient/Mapping/Movement.cs
+++ b/MinecraftClient/Mapping/Movement.cs
@@ -194,7 +194,7 @@ namespace MinecraftClient.Mapping
                     closestGoal = ClosedSet.OrderBy(checkedLocation => checkedLocation.DistanceSquared(goal)).First();
 
                 // Stop when goal is reached or we are close enough
-                if (current == goal || (minOffset > 0 && current.DistanceSquared(goal) <= minOffset))
+                if (current == goal || (minOffset > 0 && current.DistanceSquared(goal) <= Math.Pow(minOffset, 2)))
                     return ReconstructPath(Came_From, current);
                 else if (ct.IsCancellationRequested)
                     break;              // Return if we are cancelled
@@ -222,7 +222,7 @@ namespace MinecraftClient.Mapping
             }
 
             // Goal could not be reached. Set the path to the closest location if close enough
-            if (maxOffset == int.MaxValue || goal.DistanceSquared(closestGoal) <= maxOffset)            
+            if (maxOffset == int.MaxValue || goal.DistanceSquared(closestGoal) <= Math.Pow(maxOffset, 2))            
                 return ReconstructPath(Came_From, closestGoal);
             else
                 return null;
--- a/MinecraftClient/Mapping/Movement.cs
+++ b/MinecraftClient/Mapping/Movement.cs
@@ -198,7 +198,7 @@ namespace MinecraftClient.Mapping
                     closestGoal = ClosedSet.OrderBy(checkedLocation => checkedLocation.DistanceSquared(goal)).First();
 
                 // Stop when goal is reached or we are close enough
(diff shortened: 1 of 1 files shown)


Daenges:
There is one extra thing, I could fix. If several notes habe the same f-score, you usually pick the one with the lowest h-score. Currently we only pick the one that comes first in the list.
The problem here is that we currently do not save the h-score and instead direvtly evaluate the f-score with it.
So we would need to use a dictionary or a seperate class for each node to fix this. And tbh I do not know if this even has a significant impact.

ORelio:
Good catch. Instead of calculating the square every time, I'd just multiply the parameter by itself at the beginning of the function (after checking `minOffset > maxOffset`), might be quicker than calling math.pow for a power of 2.

> There is one extra thing, I could fix. If several nodes have the same f-score, you usually pick the one with the lowest h-score. Currently we only pick the one that comes first in the list.

Regarding h score, as you said I'm not 100% sure it's worth to make the change, but if i'm correct, h-score is distance from node to goal, right? So the best guess is currently selected by h-score, not f-score right?
```
closestGoal = ClosedSet.OrderBy(checkedLocation => checkedLocation.DistanceSquared(goal)).First();
```

Daenges:
>  I'd just multiply the parameter by itself ... might be quicker than calling math.pow for a power of 2.

From what I read online it is. I am going to change this.

So the problem with the selection of the next node is in here:
```cs
current = //the node in OpenSet having the lowest f_score[] value
                    OpenSet.Select(location => f_score.ContainsKey(location)
                    ? new KeyValuePair<Location, int>(location, f_score[location])
                    : new KeyValuePair<Location, int>(location, int.MaxValue))
                    .OrderBy(pair => pair.Value).First().Key;
```
We order all nodes after their `f-score` (so the sum of their distance to start `g-score` and the distance to the goal `h-score`) and just take the first one. This works fine up until the point, where we have 2 or more Nodes, that have the same `f-score`. Usually we would now pick the one out of them, that has the lowest `h-score` (distance to goal). This is currently not possible, since we do not save it, as we directly calculate the `f-score` with it here:

```cs
f_score[neighbor] = g_score[neighbor] + (int)neighbor.DistanceSquared(goal);
```
[Here](https://youtu.be/-L-WgKMFuhE?t=187) is a video that shows it a bit more graphical.

Daenges:
@ORelio something like this should do the job:
```cs
current = //the node in OpenSet having the lowest f_score[] value
                    OpenSet.Select(location => f_score.ContainsKey(location)
                    ? new KeyValuePair<Location, int>(location, f_score[location])
                    : new KeyValuePair<Location, int>(location, int.MaxValue))
                    .OrderBy(pair => pair.Value).ThenBy(pair => pair.Key.DistanceSquared(goal)).First().Key;
```
But we calculate the distance for each block twice, since we additionally set their f-score below.

Daenges:
Oh wait.... We know the f-score and the g-score... We can just subtract and obtain the h-score.
```cs
                current = //the node in OpenSet having the lowest f_score[] value
                    OpenSet.Select(location => f_score.ContainsKey(location)
                    ? new KeyValuePair<Location, int>(location, f_score[location])
                    : new KeyValuePair<Location, int>(location, int.MaxValue))
                    .OrderBy(pair => pair.Value).ThenBy(pair => f_score[pair.Key]-g_score[pair.Key]).First().Key;
```

ORelio:
Good catch! The PR looks good to me but it seems like you closed it?

Daenges:
> Good catch! The PR looks good to me but it seems like you closed it?

Whoops. I messed around with git / Github and accidentally deleted my private repo... Currently I am working on an enhancement of Minecube.cs, where I made further improvements to Movement.cs. You can either try to merge this PR if it is possible or it will come with the next one. :P

ORelio:
OK, I was able to reopen and merge the pull request so it's now included in the repository.

PR #2014, opened 2022-05-05
Rework of MineCube.cs and further improvements to CalculatePath()

Daenges:
The rework of CalculatePath() inspired me to take another look into this script. And - Oh boy, this script really needed a rework.

A major problem I wanted to fix was, that the script used the Cube class to precalculate all positions it wants to mine. This took a lot of time and ressources. I had some complex ideas on how to fix it, but after messing around with it, I realized that the whole class is not needed. The script now consists of several loops, which just iterate through all blocks. On top of that we pre-evaluate blocks in a seperate task to keep performance, while being able to react on block changes.

The script works really fine. The only problem I encountered is, that CalculatePath() function often leads to a position, where the client stands on top of the block it currently wants to mine. Unfortunately I reworked CalculatePath() in a way, that allows the client to come closer than minOffset and not **guarantee** a certain distance to the goal. It took a longer time for me to come up with a solution for it. 

If a minOffset is given, we disable the direct goal as returning condition. This leads the function to keep on discovering close nodes to the goal. At the same time we use the distance between minOffset and maxOffset as return condition. If the function now discovers a block, with a distance between minOffset and maxOffset to the goal, we return. Thereby we can guarantee a distance to the goal of minOffset.

This solution is not final, since the client currently also sees moving **up** as a reliable move to escape a fall down. xD

Included are all improvements from my old PR #2013. Before merging I want to test how my changes in CalculatePath() influence the performance.


ORelio:
Thanks for your work! Since I merged #2013 I think you'll need to make some adjustments to this pull request.

Daenges:
I think I found the optimal solution the `h-score` evaluation if the `f-score` is the same. We avoid an unnecessary `thenby()` sorting and even make the list smaller, the second time we sort. If there are no equal values, we do not even sort and return the first entry. :D

<details>
<summary>Click for code</summary>

```cs
                List<KeyValuePair<Location, int>> nodesOrderedByF_Score = // Nodes sorted by their f_score
                    OpenSet.Select(location => f_score.ContainsKey(location)
                    ? new KeyValuePair<Location, int>(location, f_score[location])
                    : new KeyValuePair<Location, int>(location, int.MaxValue)).OrderBy(pair => pair.Value).ToList();
                
                current = nodesOrderedByF_Score[0].Value == nodesOrderedByF_Score[1].Value ?                            // If the f-score of the first two nodes is the same
                    nodesOrderedByF_Score.TakeWhile(location => location.Value == nodesOrderedByF_Score[0].Value).      // Take all that have the same f-score from the beginning (list sorted ascending)
                    OrderBy(location => f_score[location.Key] - g_score[location.Key]).First().Key                      // Sort them after their distance to goal (h-score) and take the first
                    :                                                                                                   // OR
                    nodesOrderedByF_Score.First().Key;                                                                  // Just take the first location
```

</details>

ORelio:
Good idea. I get that the second sort is only done on the nodes with lowest f-score instead of the full collection 👍

If we push that idea further: OrderBy() may be overkill because we don't want to order, we just want to retain the lowest value(s), so something more like Min() but that could return several values instead of just one. Maybe it's not available through LINQ and would need to be implemented.

If implementing, the algorithm would look like this (totally wrote without actually checking if it compiles):
```
FindSmallestNodes(IEnumerable<Location> list, Func<Location, int> scoringFunction)
{
    int smallestScore = int.MaxValue;
    List<Location> smallestNodes;
    for (node in list)
    {
        int score = scoringFunction(node);
        if (score < smallestScore)
        {
            smallestValue = score;
            smallestNodes.Clear();
        }
        if (score == smallestScore)
            smallestNodes.Add(node);
    }
    return smallestNodes;
}
```
its complexity is o(n) because it iterates once on the list. I'm not sure what is the complexity of OrderBy() but maybe something like o(n log(n)) which is not really more expensive (depends on amount of nodes processed of course).

Daenges:
Thank you for your suggestion, it looks very promissing!

> OrderBy() but maybe something like o(n log(n))

Indeed I found a [comment](https://stackoverflow.com/a/41213573) which states that time complexity is (typically for something like Quicksort) `O(n*log(n))` and in the worst case `O(n²)`.

On top of that it also says, that the `thenby()` function only adds an additional sorting parameter to the corresponding `OrderBy()` and does not actually trigger a second iteration through the list. (What I initially thought)
So `thenby()` does not have a really high impact on complexity.

Nevertheless your suggestion is clearly better. Online there are only [suggestions](https://stackoverflow.com/a/3820058) with again two iterations. One to find the lowest element and one to get all with the same value. Technically more efficient than my approach, but less efficient than yours.

I will look into implementing your suggestion. :D

Daenges:
Okay, so I came up with this clunky function to solve the sorting issue. Unfortunately I did not quite get some points of your suggested *pseudocode*. Any improvement ideas?

I came across many easier implementations, for instance [MinBy()](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.minby?view=net-6.0) seems pretty helpful, but is not supported in our current .Net version.

A workaround, which is supported, can be seen in [this thread](https://coderanch.com/t/714845/languages/Find-minimum-values). Here we iterate through the list 2/3 times (Ignoring the initial creation of the KeyValuePair List of location to f-score). The current implementation would at max take one full and one light (through reduced elements) iteration through the list.
But here again I can not estimate on how far LINQ is optimized that it can compensate this and how much maintainability (3 lines are easier than a whole function) affects the decision on unoptimized code in favor of readability. (I think this issue applies to all current optimizations.. :/ )

In [this article](https://www.codeguru.com/csharp/a-star-a-implementation-in-c-path-finding-pathfinder/) a [PriorityQueue](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.priorityqueue-2?view=net-6.0) is suggested instead of a Hashset for the `OpenSet`, which makes obtaining the next node with the lowest `f-score` way easier. But it is not in our current .Net version and if we do not want to implement a [Binary Heap Class](https://www.geeksforgeeks.org/binary-heap/) on our own, we are tied to use Hashsets.

We might be able to replace the foreach loops with [.Aggregate()](https://stackoverflow.com/a/2736385) (**at least this is supported** :P), but here we need to use cascading `condition ? condition ? something : somethingElse : somethingElseElse` which will probably end up in a hell to maintain.

```cs
private static Location getNodeWithLowestF_Score(HashSet<Location> openNodes, Dictionary<Location, int> g_scoreDict, Dictionary<Location, int> f_scoreDict)
        {
            List<KeyValuePair<Location, int>> smallestF_ScoreNodes = new List<KeyValuePair<Location, int>>();

            foreach (Location currentNode in openNodes)
            {
                int currentF_score = f_scoreDict.ContainsKey(currentNode) ? f_scoreDict[currentNode] : int.MaxValue;
                if (smallestF_ScoreNodes.Count() > 0)
                {
                    if (currentF_score < smallestF_ScoreNodes[0].Value)
                        smallestF_ScoreNodes.Clear();

                    if (currentF_score <= smallestF_ScoreNodes[0].Value)
                        smallestF_ScoreNodes.Add(new KeyValuePair<Location, int>(currentNode, currentF_score));
                }
                else
                {
                    smallestF_ScoreNodes.Add(new KeyValuePair<Location, int>(currentNode, currentF_score));
                }
            }

            if (smallestF_ScoreNodes.Count == 1)
            {
                return smallestF_ScoreNodes[0].Key;
            }
            else
            {
                KeyValuePair<Location, int> lowH_ScoreNode = new KeyValuePair<Location, int>(Location.Zero, int.MaxValue);

                for(int i=0; i < smallestF_ScoreNodes.Count(); i++)
                {
                    int currentH_Score = f_scoreDict[smallestF_ScoreNodes[i].Key] - g_scoreDict[smallestF_ScoreNodes[i].Key];
                    if (currentH_Score < lowH_ScoreNode.Value)
                    {
                        lowH_ScoreNode = new KeyValuePair<Location, int>(smallestF_ScoreNodes[i].Key, currentH_Score);
                    }
                }

                return lowH_ScoreNode.Key;
            }

            throw new Exception("Unable to obtain the lowest f-Score");
        }
```

Daenges:
@ORelio Some time has passed and I tested around a bit. I built a small maze and tested the different approaches to get the node with the lowest `f-score`:

|Algorithm|Trial 1 </br>calculation Time </br> (in ms)|Trial 2|Tiral 3| Tiral 4|Trial 5|Trial 6|Tiral 7|AVG.|
|-|-|-|-|-|-|-|-|-|
|Current Implementation|90|56|57|53|56|59|57|61.14|
|Avoid 2nd sort|85|52|46|47|46|49|53|53|
|[Your suggestion](https://github.com/MCCTeam/Minecraft-Console-Client/pull/2014#issuecomment-1120495960)|60|35|33|36|35|34|34|38.14|

A binary heap sound pretty promising at this point with the time complexity O(log(N)). If we would be using .NET 6 we would have a PriorityQueue as native implementation, but as .NET 5 is currently idling as a pull request, it seems like if we want to try this, I need to implement this on my own.

Daenges:
@ORelio great news! After a few days and a lack of motivation I managed to write a BinaryHeap Implementation for our `Open Set`.

<details>
<summary>Click for code</summary>

```cs

public class Location
{
    public int X, Y, Z;

    public Location(int x, int y, int z)
    {
        X = x;
        Y = y;
        Z = z;
    }

}

public class BinaryHeap
{
    public class Node
    {
        public int G_score;
        public int F_score;
        public int H_score { get { return F_score - G_score; } }

        public Location Location;

        public Node(int g_score, int f_score, Location loc)
        {
            this.G_score = g_score;
            this.F_score = f_score;
            Location = loc;
        }
    }

    private List<Node> heapList;

    public BinaryHeap()
    {
        heapList = new List<Node>();
    }

    public void Insert(int newG_Score, int newF_Score, Location loc)
    {
        // Begin at the end of the list
        int i = heapList.Count;

        // Temporarily save the node created with the parameters to allow comparisons
        Node newNode = new Node(newG_Score, newF_Score, loc);

        // Add new note to the end of the list
        heapList.Add(newNode);

        // There is no need of sorting for one node.
        if (i > 0)
        {
            /// Go up the heap from child to parent and move parent down...
            // while we are not looking at the root node AND the new node has better attributes than the parent node ((i - 1) / 2)
            while (i > 0 && FirstNodeBetter(newNode /* Current Child */, heapList[(i - 1) / 2] /* Coresponding Parent */))
            {
                // Move parent down and replace current child -> New free space is created
                heapList[i] = heapList[(i - 1) / 2];
                // Select the next parent to check
                i = (i - 1) / 2;
            }

            /// Nodes were moved down at position I there is now a free space at the correct position for our new node:
            // Insert new node in position
            heapList[i] = newNode;
        }
    }

    public Location GetRootLocation()
    {
        // The heap is empty. There is nothing to return.
        if (heapList.Count == 0)
        {
            throw new InvalidOperationException("The heap is empty.");
        }

        // Save the root node
        Node rootNode = heapList[0];

        // Temporarirly store the last item's value.
        Node lastNode = heapList[heapList.Count - 1];

        // Remove the last value.
        heapList.RemoveAt(heapList.Count - 1);

        if (heapList.Count > 0)
        {
            // Start at the first index.
            int currentParentPos = 0;

            /// Go through the heap from root to bottom...
            // Continue until the halfway point of the heap.
            while (currentParentPos < heapList.Count / 2)
            {
                // Select the left child of the current parent
                int currentChildPos = (2 * currentParentPos) + 1;

                // If the currently selected child is not the last entry of the list AND right child has better attributes
                if ((currentChildPos < heapList.Count - 1) && FirstNodeBetter(heapList[currentChildPos + 1], heapList[currentChildPos]))
                {
                    // Select the right child
                    currentChildPos++;
                }

                // If the last item is smaller than both siblings at the
                // current height, break.
                if (FirstNodeBetter(lastNode, heapList[currentChildPos]))
                {
                    break;
                }

                // Move the item at index j up one level.
                heapList[currentParentPos] = heapList[currentChildPos];
                // Move index i to the appropriate branch.
                currentParentPos = currentChildPos;
            }
            // Insert the last node into the currently free position
            heapList[currentParentPos] = lastNode;
        }

        return rootNode.Location;
    }

    /// <summary>
    /// Compares two nodes and evaluates their position to the goal.
    /// </summary>
    /// <param name="firstNode">First node to compare</param>
    /// <param name="secondNode">Second node to compare</param>
    /// <returns>True if the first node has a more promissing position to the goal than the second</returns>
    private static bool FirstNodeBetter(Node firstNode, Node secondNode)
    {
        // Is the F_score smaller?
        return (firstNode.F_score < secondNode.F_score) ||
            // If F_score is equal, evaluate the h-score
            (firstNode.F_score == secondNode.F_score && firstNode.H_score < secondNode.H_score);
    }

    /// <summary>
    /// Obtain the heap for query operations.
    /// </summary>
    /// <returns>current heap as readonly</returns>
    public IReadOnlyCollection<Node> GetFullHeap()
    {
        return heapList.AsReadOnly();
    }
}
```

</details>

I took a lot of inspiration from [this post](https://gist.github.com/dsmith86/665c8b22c57e873c556e). I attached some comments and changed variable names so I could understand it a bit better. I wrote this in a new project, since it is literally impossible to debug something like this, while keeping the connection to a server. I will now implement this into the client and check if all the effort was worth it.

ORelio:
_(Sorry for responding late)_ Honestly I didn't think my suggestion would improve performance that much, it's very nice that you took the time to benchmark each approach! If I get it correctly, the BinaryHeap would allow pre-sorting elements so FindSmallestNodes() would no longer needed at all?

Daenges:
> (Sorry for responding late)

No worries, I get that people have a life and that there are more important things than a minecraft client hobby project. ;D

> If I get it correctly, the BinaryHeap would allow pre-sorting elements so FindSmallestNodes() would no longer needed at all?

Yeah, basically. We create a binary tree, where each node is a Location with its corresponding `Scores` if we add a node at the end, we compare it up the tree until we hit the root.



If we now want to get the current root node, so the node with the lowest `f-score` (or relative `h-score` if two nodes have the same `f-score`) we go from top to bottom and push values up.

ORelio:
Looks very nice. I'm not sure whether pre-sorting with binary heap (a little more computation every time the list is edited) or just selecting the smallest node at the end (no real computation when the list is edited, but a larger one at the end) is the fastest solution, but I'm sure you'll benchmark that 😄

Daenges:
After a **way** too many tries of implementing, huge frustration, a period of demotivation and a profile picture change, I **FINALLY** got a working implementation of a binary heap in A*. 🎉 

<details>
<summary>See awesome code (Not final, but working (too excited to show you))</summary>

```cs
/// <summary>
        /// Calculate a path from the start location to the destination location
        /// </summary>
        /// <remarks>
        /// Based on the A* pathfinding algorithm described on Wikipedia
        /// </remarks>
        /// <see href="https://en.wikipedia.org/wiki/A*_search_algorithm#Pseudocode"/>
        /// <param name="start">Start location</param>
        /// <param name="goal">Destination location</param>
        /// <param name="allowUnsafe">Allow possible but unsafe locations</param>
        /// <param name="maxOffset">If no valid path can be found, also allow locations within specified distance of destination</param>
        /// <param name="minOffset">Do not get closer of destination than specified distance</param>
        /// <param name="ct">Token for stopping computation after a certain time</param>
        /// <returns>A list of locations, or null if calculation failed</returns>
        public static Queue<Location> CalculatePath(World world, Location start, Location goal, bool allowUnsafe, int maxOffset, int minOffset, CancellationToken ct)
        {

            start = new Location(Math.Round(start.X), Math.Round(start.Y), Math.Round(start.Z));

            if (minOffset > maxOffset)
                throw new ArgumentException("minOffset must be lower or equal to maxOffset", "minOffset");

            // We always use distance squared so our limits must also be squared.
            minOffset *= minOffset;
            maxOffset *= maxOffset;

            Dictionary<Location, Location> CameFrom = new Dictionary<Location, Location>();
            BinaryHeap openSet = new BinaryHeap();
            Dictionary<Location, int> g_score = new Dictionary<Location, int>();
            openSet.Insert(0, (int)start.DistanceSquared(goal), start);
            g_score[start] = 0;

            while (openSet.Count() > 0)
            {
                BinaryHeap.Node current = openSet.GetRootLocation();

                if (current.Location == goal)
                {
                    return ReconstructPath(CameFrom, current.Location);
                }

                foreach (Location neighbor in GetAvailableMoves(world, current.Location, allowUnsafe))
                {
                    int tentativeG_Score = current.G_score + (int)current.Location.DistanceSquared(neighbor);

                    if (!g_score.ContainsKey(neighbor) || (g_score.ContainsKey(neighbor) && tentativeG_Score < g_score[neighbor]))
                    {
                        CameFrom[neighbor] = current.Location;
                        g_score[neighbor] = tentativeG_Score;

                        if (!openSet.ContainsLocation(neighbor))
                            openSet.Insert(new BinaryHeap.Node(tentativeG_Score, (int)neighbor.DistanceSquared(goal), neighbor));
                    }
                }
            }

            throw new Exception("Mööp");

            //// Goal could not be reached. Set the path to the closest location if close enough
            //if (maxOffset == int.MaxValue || goal.DistanceSquared(closestGoal) <= maxOffset)
            //    return ReconstructPath(Came_From, closestGoal);
            //else
            //    return null;
        }

        private static Func<Dictionary<Location, int>, Location, int> getValueOrInfinity = (Dictionary<Location, int> d, Location l) => d.ContainsKey(l) ? d[l] : int.MaxValue;

        public class BinaryHeap
        {
            public class Node
            {
                public int G_score;
                public int H_score;
                public int F_score { get { return H_score + G_score; } }

                public Location Location;

                public Node(int g_score, int h_score, Location loc)
                {
                    this.G_score = g_score;
                    this.H_score = h_score;
                    Location = loc;
                }
            }

            private List<Node> heapList;
            private HashSet<Location> locationList;

            public BinaryHeap()
            {
                heapList = new List<Node>();
                locationList = new HashSet<Location>();
            }

            public void Insert(Node n)
            {
                Insert(n.G_score, n.H_score, n.Location);
            }

            public void Insert(int newG_Score, int newH_Score, Location loc)
            {
                // Begin at the end of the list
                int i = heapList.Count;

                // Temporarily save the node created with the parameters to allow comparisons
                Node newNode = new Node(newG_Score, newH_Score, loc);

                // Add new note to the end of the list
                heapList.Add(newNode);
                locationList.Add(loc);

                // There is no need of sorting for one node.
                if (i > 0)
                {
                    /// Go up the heap from child to parent and move parent down...
                    // while we are not looking at the root node AND the new node has better attributes than the parent node ((i - 1) / 2)
                    while (i > 0 && FirstNodeBetter(newNode /* Current Child */, heapList[(i - 1) / 2] /* Coresponding Parent */))
                    {
                        // Move parent down and replace current child -> New free space is created
                        heapList[i] = heapList[(i - 1) / 2];
                        // Select the next parent to check
                        i = (i - 1) / 2;
                    }

                    /// Nodes were moved down at position I there is now a free space at the correct position for our new node:
                    // Insert new node in position
                    heapList[i] = newNode;
                }
            }

            public Node GetRootLocation()
            {
                // The heap is empty. There is nothing to return.
                if (heapList.Count == 0)
                {
                    throw new InvalidOperationException("The heap is empty.");
                }

                // Save the root node
                Node rootNode = heapList[0];
                locationList.Remove(rootNode.Location);

                // Temporarirly store the last item's value.
                Node lastNode = heapList[heapList.Count - 1];

                // Remove the last value.
                heapList.RemoveAt(heapList.Count - 1);

                if (heapList.Count > 0)
                {
                    // Start at the first index.
                    int currentParentPos = 0;

                    /// Go through the heap from root to bottom...
                    // Continue until the halfway point of the heap.
                    while (currentParentPos < heapList.Count / 2)
                    {
                        // Select the left child of the current parent
                        int currentChildPos = (2 * currentParentPos) + 1;

                        // If the currently selected child is not the last entry of the list AND right child has better attributes
                        if ((currentChildPos < heapList.Count - 1) && FirstNodeBetter(heapList[currentChildPos + 1], heapList[currentChildPos]))
                        {
                            // Select the right child
                            currentChildPos++;
                        }

                        // If the last item is smaller than both siblings at the
                        // current height, break.
                        if (FirstNodeBetter(lastNode, heapList[currentChildPos]))
                        {
                            break;
                        }

                        // Move the item at index j up one level.
                        heapList[currentParentPos] = heapList[currentChildPos];
                        // Move index i to the appropriate branch.
                        currentParentPos = currentChildPos;
                    }
                    // Insert the last node into the currently free position
                    heapList[currentParentPos] = lastNode;
                }

                return rootNode; //.Location
            }

            /// <summary>
            /// Compares two nodes and evaluates their position to the goal.
            /// </summary>
            /// <param name="firstNode">First node to compare</param>
            /// <param name="secondNode">Second node to compare</param>
            /// <returns>True if the first node has a more promissing position to the goal than the second</returns>
            private static bool FirstNodeBetter(Node firstNode, Node secondNode)
            {
                // Is the F_score smaller?
                return (firstNode.F_score < secondNode.F_score) ||
                    // If F_score is equal, evaluate the h-score
                    (firstNode.F_score == secondNode.F_score && firstNode.H_score < secondNode.H_score);
            }

            /// <summary>
            /// Obtain the heap for query operations.
            /// </summary>
            /// <returns>current heap as readonly</returns>
            public List<Node> GetHeapCopy()
            {
                return new List<Node>(heapList);
            }

            public int Count()
            {
                return heapList.Count;
            }

            public bool ContainsLocation(Location loc)
            {
                return locationList.Contains(loc);
            }
        }

        /// <summary>
        /// Helper function for CalculatePath(). Backtrack from goal to start to reconstruct a step-by-step path.
        /// </summary>
        /// <param name="Came_From">The collection of Locations that leads back to the start</param>
        /// <param name="current">Endpoint of our later walk</param>
        /// <returns>the path that leads to current from the start position</returns>
        private static Queue<Location> ReconstructPath(Dictionary<Location, Location> Came_From, Location current)
        {
            List<Location> total_path = new List<Location>(new[] { current });
            while (Came_From.ContainsKey(current))
            {
                current = Came_From[current];
                total_path.Add(current);
            }
            total_path.Reverse();
            return new Queue<Location>(total_path);
        }
```

</details>

And **surprise surprise** @ORelio: It was totally worth it!

|Algorithm|Trial 1 | Trial 2|Trial 3|Trial 4|Trial 5| Trial 6| Trial 7|AVG|
|-|-|-|-|-|-|-|-|-|
|Binary Heap|28|15|17|17|15|15|15|17.43|
 
We halved the time **AGAIN**! \o/

ORelio:
Wow, congrats!! Your previous PR already upgraded the basic pathfinding code to one that had decent performance, but optimizing again from 50-90ms down to 15-28ms is ... amazing. Did not expect such significant improvement 🤯 👏 🥇

Daenges:
Apposite to `Pride Month` we encourage non straight paths. 🏳️‍🌈

ORelio:
Good idea 😆

Daenges:
Although I have some ideas left, this PR is slowly approximating its completion. I would appreciate some feedback and improvements.
I will include them with the following commits.

Daenges:
I have decided to deal with further ideas in separate PRs.
@ORelio if you find some time between reviewing the `.Net 5` rework, this is ready for merge.

ORelio:
Thanks for your work! I've been away for a while but better late than never. Merged 🙂

PR #2044, opened 2022-06-10
Adds support for multiple interact modes

Booquefius:


MinecraftClient/ChatBots/AutoAttack.cs
```diff
@@ -37,6 +38,14 @@ public AutoAttack(string mode, string priority, bool overrideAttackSpeed = false
                 priorityDistance = false;
             else LogToConsoleTranslated("bot.autoAttack.priority", priority);
 
+            if (attack == "ATTACK")
+                interactMode = 0;
+            else if (attack == "INTERACT")
+                interactMode = 1;
+            else if (attack == "INTERACT_AT")
+                interactMode = 2;
```
ORelio, line 46: Looks nice! Would you please use an `enum` for this field?

ORelio (changes requested):
Thanks for contributing!
I'd suggest to use an `enum` to facilitate usage and avoid typos.

ORelio (approved):

PR #2059, opened 2022-07-03
Fix automated build for .NET releases

breadbyte:


ReinforceZwei:
Could you also update the readme `Building from source` section for .NET 6?

breadbyte:
Updated the README file. Needs review.

Verified that the build action works on my own fork, and that the manual build instructions are correct.

Since it is currently configured as a single file application, should we skip bundling the `.pdb` files and just put the executable files themselves as a release?
Keeping them zipped is beneficial though, it reduces the total download size from ~27mb (windows executable) to only ~8mb (zip)

Currently osx/mac releases are not built, should we add it as well?

image

ReinforceZwei:
> should we skip bundling the .pdb files and just put the executable files themselves as a release?

I think those files should be hide from user, either not include them in the zip file or embed them in the exe file.

> Currently osx/mac releases are not built, should we add it as well?

Sure.

ORelio:
Thanks for your work! 👍 (sorry for taking this long to review your PR)

ORelio:
Just noticed that the automated build for Windows has no Exe icon. Apart from that, it works 🙂

breadbyte:
I'm honestly a bit torn about whether to commit this PR due to #2082 being a devastating bug that I'm still trying to get to the root of.


> Just noticed that the automated build for Windows has no Exe icon. Apart from that, it works 🙂

Noted. I'll try and figure something out.

PR #2076, opened 2022-07-22
ConsoleIO: fix timestamp & BasicIO-NoColor

xdavidwu:


MinecraftClient/ConsoleIO.cs
```diff
@@ -123,29 +123,29 @@ public static void WriteLineFormatted(string str, bool acceptnewlines = false, b
...
+                    output.Append(str);
                 }
                 if (BasicIO)
                 {
-                    if (BasicIO_NoColor)
-                    {
-                        output.Append(ChatBot.GetVerbatim(str));
-                    }
                     Console.WriteLine(output.ToString());
                     return;
```
breadbyte, line 150: Since `BasicIO` is `true` when `BasicIO_NoColor` is `true`, it's possible to put the if check for `BasicIO_NoColor` back with a little bit of extra code.

```cs
    if (BasicIO)
    {
        if (BasicIO_NoColor)
            output.Append(ChatBot.GetVerbatim(str));
        else 
            output.Append(str);

        Console.WriteLine(output.ToString());
        return;
    }
    else
    {
        output.Append(str);
    }
```

I don't recall why I added a `StringBuilder` here, and it doesn't look to me like it's doing anything useful, so it might be best to just revert the `StringBuilder` changes.

breadbyte:
Thank you for your contribution! Just a little bit of comment that I had, otherwise this looks good to me!

ORelio:
Thanks for this fix! 👍

PR #2114, opened 2022-08-20
Allow newer SSL/TLS protocol versions

Daenges:
Fixes [this issue](https://github.com/MCCTeam/Minecraft-Console-Client/discussions/1885#discussioncomment-3434649).

### DIFF
--- a/MinecraftClient/config/ChatBots/DiscordWebhook.cs
+++ b/MinecraftClient/config/ChatBots/DiscordWebhook.cs
@@ -279,6 +279,12 @@ class HTTP
 {
     public static byte[] Post(string url, NameValueCollection pairs)
     {
+        ServicePointManager.Expect100Continue = true;
+		ServicePointManager.SecurityProtocol = SecurityProtocolType.Tls
+                   | SecurityProtocolType.Tls11
+                   | SecurityProtocolType.Tls12
+                   | SecurityProtocolType.Ssl3;
+        
         using (WebClient webClient = new WebClient())
         {
             return webClient.UploadValues(url, pairs);


ORelio:
Would you mind adding a check so that the fix is only applied when launching the script using Mono?

haventhadadonutyet:
This works! - Question though, is it possible to have it relay messages only if there is a word match? Like, messages without a sender, such as custom messages from the server?

Daenges:
@ORelio the error appeared on a Windows 10 machine. There was no mono installed.

milutinke:
> @ORelio the error appeared on a Windows 10 machine. There was no mono installed.

Since .NET 5.0 mono is no longer required, neither it works.

Daenges:
@milutinke I tried to execute this script on the most recent build and it failed to compile. To verify this issue I downloaded an [older build](https://github.com/MCCTeam/Minecraft-Console-Client/releases/tag/20220506-20). So they might be using the 4.8 Framework version.

Daenges:
> Like, customchatdetection but only if a certain word or phrase is included

@haventhadadonutyet You can try to work with the ping system, so you actually get notified if the message contains a certain phrase. Or you try around with [Chatformats](https://github.com/MCCTeam/Minecraft-Console-Client/blob/master/MinecraftClient/Resources/config/MinecraftClient.ini#L94) and regular expressions to find messages with certain phrases.

ORelio:
> > @ORelio the error appeared on a Windows 10 machine. There was no mono installed.

> Since .NET 5.0/6.0 port, mono is no longer required, neither it works.

Oh, OK. This is weird because on .NET 5+ you should not need to manually set the TLS protocols like in .NET 4
Maybe the issue is related to this hotfix here that is likely no longer required:
https://github.com/MCCTeam/Minecraft-Console-Client/blob/bb4c13168beaefb278d21140f3ae56ef788b8b00/MinecraftClient/Protocol/ProxiedWebRequest.cs#L127

Daenges:
@ORelio as I mentioned [here](https://github.com/MCCTeam/Minecraft-Console-Client/pull/2114#issuecomment-1221370815) it is very likely that they didn't use the new .NET build, as the script is not working in it.

ORelio:
> @ORelio as I mentioned [here](https://github.com/MCCTeam/Minecraft-Console-Client/pull/2114#issuecomment-1221370815) it is very likely that they didn't use the new .NET build, as the script is not working in it.

OK so this is a fix for use with legacy builds. Okay, merging the PR.

PR #2116, opened 2022-08-20
Fix the browser not opening on Linux.

milutinke:
Tested on Windows, Arch and Raspbian.
Needs testing on macOS x64 and Arm64.


milutinke:
Update: Tested on macOS 12 Arm x64 Monterey, working.

milutinke:
Update, fixed on Windows (sometimes it would not work).
Tested on Pop_OS, working.

PR #2120, opened 2022-08-22
[1.19] Added Entity, Block, Material and Item palettes, Enabled Entity and Inventory handling + Fixed a bug

milutinke:
Added Entity, Block, Material and Item palettes.
Fixed a bug in `SpawnEntity` packet which was causing crashes when Entity handling was enabled.
Temporarily disabled `BlockChanged` packet for 1.19 due to crashing (I am not familiar with Terrain and Movement Handling code).

Tested Auto Attack bot and inventory handling, it works, tho ASCII art is broken due to ConsoleInteractive.

PR #2123, opened 2022-08-24
Basic support for 1.19.1 / 1.19.2

### LINKED ISSUE #2102: [Idea] 1.19.2 Support?
### Prerequisites

- [X] I have read and understood the [user manual](https://github.com/MCCTeam/Minecraft-Console-Client/tree/master/MinecraftClient/config)
- [X] I made sure I am running the latest [development build](https://ci.appveyor.com/project/ORelio/Minecraft-Console-Client/build/artifacts)
- [X] I tried to [look for similar feature requests](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

### Console Client Version

1.19

### Describe your problem

Can't connect to 1.19.2 servers

### Suggest a possible solution

_No response_

_No response_

### Minecraft Version

1.19.2

### Device

Desktop

### Operating System

Windows

BruceChenQAQ:
Current Progress:
- [x] Login Support with online-mode=false
- [x] Login Support with online-mode=true, enforce-secure-profile=true
- [x] Processing the text of received chats
- [x] Present the received chat in the correct format
- [x] Handle MessageHeader packet
- [x] Signature verification for chat messages
- [x] Send MessageAcknowledgment packet when receiving chat
- [x] Send chat with signatures
- [x] Send commands with signatures
- [x] Terrain support for 1.19
  (Entity and Inventory handling at https://github.com/MCCTeam/Minecraft-Console-Client/pull/2120 )
- [x] Terrain support for 1.19.1 / 1.19.2

Resolved #2102


milutinke:
There is this bug, I've wanted to test out the terrain movement.
This crashes the client immediately.
image

milutinke:
Seems like that terrain handling doesn't work for me on 1.18.2 with this build.
It just hangs.

milutinke:
I've figured out what is happening, so basically chunks are not loaded, and then the client just hangs until they're loaded, but by that time it times out.
We could stop the movement until the chunks are loaded (like enqueue it), or try to figure out if the movements are in the already loaded chunk.
And i think that movement should be moved to a separate thread if it is not already in a separate thread.
And adding a custom setting for timeout would be nice too.

Also, one more thing I've noticed is that when your destination coordinate is an edge, it often fails to compute the path, I think that this is due to bot moving 2 blocks at a time (not familiar with that part of the code), when I add an additional block it then works.

BruceChenQAQ:
> 



> I've figured out what is happening, so basically chunks are not loaded, and then the client just hangs until they're loaded, but by that time it times out. We could stop the movement until the chunks are loaded (like enqueue it), or try to figure out if the movements are in the already loaded chunk. And i think that movement should be moved to a separate thread if it is not already in a separate thread. And adding a custom setting for timeout would be nice too, and logging the status of chunk loading from time to time would be nice, so people do not need to manually check with `/move`.
> 
> Also, one more thing I've noticed is that when your destination coordinate is an edge, it often fails to compute the path, I think that this is due to bot moving 2 blocks at a time (not familiar with that part of the code), when I add an additional block it then works.

The problem of slow chunk loading should be fixed by https://github.com/MCCTeam/Minecraft-Console-Client/pull/2124, basically all chunks can be loaded within a second.

Currently all commands are blocking for the result, maybe I'll rewrite it later.

The problem of moving to the wrong position is probably due to a conflict in https://github.com/MCCTeam/Minecraft-Console-Client/pull/2014 and https://github.com/MCCTeam/Minecraft-Console-Client/pull/1943, which should be fixed in https://github.com/MCCTeam/Minecraft-Console-Client/pull/2124 https://github.com/MCCTeam/Minecraft-Console-Client/pull/2124/commits/ed8e97fd2dc94759621a55d7f7b43e4ead28c218.

BruceChenQAQ:
> There is this bug, I've wanted to test out the terrain movement. This crashes the client immediately.

Sorry I forgot to check if it works properly in offline mode, should be fixed in https://github.com/MCCTeam/Minecraft-Console-Client/pull/2123/commits/da02f8004f929853eecdddf2ad91c9d76fa7c189

milutinke:
Awesome.
Also, I have an idea, I do not know if it is doable since I'm not familiar with terrain and movement part of the code.
But I think we could port Baritone to MCC, at least the movement part, it has a very good path finding code with heuristics, I believe it uses `A*`, but a modified version.

BruceChenQAQ:
Coincidentally, I had this idea too.

[Baritone AStarPathFinder.java](https://github.com/cabaletta/baritone/blob/master/src/main/java/baritone/pathing/calc/AStarPathFinder.java)

What could be improved:
* Support Sprint.
* Gets the actual height of the blocks, such as slab and trapdoor in various states.
* Distinguish the time consumption in different states. [Baritone ActionCosts.java](https://github.com/cabaletta/baritone/blob/master/src/api/java/baritone/api/pathing/movement/ActionCosts.java)
* More efficient data structures. [Baritone BinaryHeapOpenSet.java]
  * It has been optimized a lot in https://github.com/MCCTeam/Minecraft-Console-Client/pull/2014

BruceChenQAQ:
Data for all dimensions (In Registry Codec):
* https://wiki.vg/Protocol#Login_.28play.29
* [Default value in JSON as of 1.19](https://gist.github.com/rj00a/f2970a8ce4d09477ec8f16003b9dce86)
* Handle in ``/Mapping/World.cs`` ``void StoreDimensionList(Dictionary<string, object> registryCodec);``

Dimensional changes in:
* https://wiki.vg/Protocol#Login_.28play.29
* https://wiki.vg/Protocol#Respawn
* Handle in ``/Mapping/World.cs`` ``void SetDimension(string name);``

About Chunk Data Packet: 
* https://wiki.vg/Protocol#Chunk_Data_and_Update_Light
* https://wiki.vg/Chunk_Format#Data_structure
* Handle in ``/Protocol/Handlers/Protocol18Terrain.cs``
* ``ProcessChunkColumnData -> ReadBlockStatesField -> World.StoreChunk``

milutinke:
> Coincidentally, I had this idea too.
> 
> [Baritone AStarPathFinder.java](https://github.com/cabaletta/baritone/blob/master/src/main/java/baritone/pathing/calc/AStarPathFinder.java)
> 
> What could be improved:
> 
>     * Support Sprint.
> 
>     * Gets the actual height of the blocks, such as slab and trapdoor in various states.
> 
>     * Distinguish the time consumption in different states. [Baritone ActionCosts.java](https://github.com/cabaletta/baritone/blob/master/src/api/java/baritone/api/pathing/movement/ActionCosts.java)
> 
>     * More efficient data structures. [Baritone BinaryHeapOpenSet.java](https://github.com/cabaletta/baritone/blob/master/src/main/java/baritone/pathing/calc/openset/BinaryHeapOpenSet.java)
>       
>       * It has been optimized a lot in [Rework of MineCube.cs and further improvements to CalculatePath() #2014](https://github.com/MCCTeam/Minecraft-Console-Client/pull/2014)
> 
> 
> Note that Baritone uses the LGPL 3.0 license, https://stackoverflow.com/questions/5634239/does-translating-a-library-into-another-language-make-it-a-derivative-work-for

I understood that correctly, only the project code that uses "derivative work" needs to licensed under LGPL and not the entire project, so I think that it won't be a problem, we also could ask for a permission too, from the original author.

milutinke:
Is this complete or there is more things to add?
And should this be merged first or #2124?
I can do the conflict resolution.

BruceChenQAQ:
Already available for merging.
Maybe it would be better to merge #2124  in the end.

milutinke:
On a server with offline mode you get `[MCC] [multiplayer.disconnect.chat_validation_failed]` when someone sends a chat message error and then you get disconnected.

BruceChenQAQ:
@milutinke 
Message signature-related issues shoule be fixed in https://github.com/MCCTeam/Minecraft-Console-Client/pull/2123/commits/f4ab4997c889aaa2f9359288f973b122f8346b04.
Signed messages in offline mode on the server will be kicked out.

I don't know much about scripting related issues, this commit doesn't change the code.

milutinke:
Another issue, sometimes the chunk loading gets stuck on `63,57%` and does not go further.

BruceChenQAQ:
> Another issue, sometimes the chunk loading gets stuck on `63,57%` and does not go further. `14:42:58 [MCC] Chunk loading status: 63,57% - 342 out of 538 load completed.`

This version had issues with chunk loading count errors and chunk loading not being cancelled when switching worlds, fixed in #2124.

BruceChenQAQ:
There seems to be some problem with the chunk loading in #2124, I'll try to fix it.

milutinke:
Tested out changes from this branch, and fixed bugs, everything except that bug with chunks and slow script compilation which happens sometimes is working fine.

Use the following:
[MaterialExtensions.cs`](https://pastebin.com/26bgjHDK)
[DataTypes.cs](https://pastebin.com/AQzY4Msa)
[Protocol18.cs](https://pastebin.com/xeeuT0Jk)

BruceChenQAQ:
> Oh yeah, one more thing, the client can't find any paths when in `1x2` tunnels

It should also be fixed in #2124 . I'll go test it now.

milutinke:
@ORelio Tested on a local and on 2 live servers, as mentioned everything now works except the stuff that is fixed in #2124 
Can I merge or you want to review the code too?

MinecraftClient/Protocol/Handlers/Protocol18.cs
```diff
@@ -91,85 +98,105 @@ public Protocol18Handler(TcpClient Client, int protocolVersion, IMinecraftComHan
...
+            if (this.protocolVersion >= MC_1_19_2_Version)
+                ChatParser.ChatId2Type = new()
+                {
+                    { 0,  ChatParser.MessageType.CHAT },
+                    { 1,  ChatParser.MessageType.SAY_COMMAND },
+                    { 2,  ChatParser.MessageType.MSG_COMMAND_INCOMING },
+                    { 3,  ChatParser.MessageType.MSG_COMMAND_OUTGOING },
+                    { 4,  ChatParser.MessageType.TEAM_MSG_COMMAND_INCOMING },
+                    { 5,  ChatParser.MessageType.TEAM_MSG_COMMAND_OUTGOING },
+                    { 6,  ChatParser.MessageType.EMOTE_COMMAND },
```
nixerus, line 186: Tested and crashed because chat type 7 (RAW_MSG) isn't included for > 760.
Screenshot

BruceChenQAQ: What form will this message be displayed in the vanilla client, and can I get the server address?
Only these six message types are preconfigured in minecraft's source code. RAW_MSG was added manually for compatibility with 1.19, and it should not be used in 1.19.1+.

BruceChenQAQ: Tested that chat is presented as a system message when using the tellraw command.
But in [RegistryCodec‘s default value](https://gist.github.com/rj00a/f2970a8ce4d09477ec8f16003b9dce86) you can find the id 7 in "minecraft:chat_type" and its type is "minecraft:tellraw_command".

Added in https://github.com/MCCTeam/Minecraft-Console-Client/pull/2123/commits/2c5b444ffdf9f8c33dc899a5b39aa96a52a24a9b

BruceChenQAQ:
Todo: Parsing the message format from the RegistryCodec nbt tag.

MinecraftClient/Protocol/Handlers/PacketTypesIn.cs
```diff
@@ -132,6 +132,11 @@ public enum PacketTypesIn
         ChatPreview,
         ServerData,
         SetDisplayChatPreview,
-        SystemChat
+        SystemChat,
+
+        // Added in 1.19.1 (1.19.2)
+        ChatSuggestions,
+        HideMessage,
+        MessageHeader
```
ORelio, line 140: I'd suggest to order the items alphabetically, and, if you want, put a comment on the right stating in which version the packet was introduced.

MinecraftClient/Protocol/Handlers/PacketTypesOut.cs
```diff
@@ -67,5 +67,8 @@ public enum PacketTypesOut
         // Added in 1.19
         ChatCommand,
         ChatPreview,
+
+        // Added in 1.19.1 (1.19.2)
+        MessageAcknowledgment
```
ORelio, line 72: Same here (I'd suggest to order the items alphabetically, and, if you want, put a comment on the right stating in which version the packet was introduced.)

ORelio (approved):
> @ORelio Tested on a local and on 2 live servers, as mentioned everything now works except the stuff that is fixed in #2124 Can I merge or you want to review the code too?

@milutinke @BruceChenQAQ The changes look good to me 👌
Just a small comment when browsing through the code:

MinecraftClient/Protocol/Handlers/PacketTypesIn.cs
BruceChenQAQ: Okay, sorted them.

BruceChenQAQ:
@milutinke I've completed the changes, is there anything else I need to do before it merges?

milutinke:
Tested everything once more, everything works.
So it can be merged.

PR #2126, opened 2022-08-26
Added a WebSocket Chat Bot

milutinke:
Added a Web Socket bot that enables remote control and sends all events from `ChatBot` API over the network.

The motivation for this Chat Bot:
 - Enables people who know other languages and do not want to learn C# to make Chat Bots in a language of their choice
 - Enables people to make nice UIs for controlling their MCC instances

Current state of the Chat Bot:
- All events from the ChatBot API are being sent
- Has authentication
- Can support multiple connections at the same time
- Can receive and execute MCC commands

Plans:

1. To add some Pre-Made utility commands that could be called using the WebSocket
2. In the following days I am going to write a JavaScript library which will enable people to make Chat Bots in JS.
(I've tried embedding JavaScript as a second scripting way with ClearScript, but it turned out to be too complex of a task to make a clean looking API, maybe someone else can give it a try).
3. To make an application that will use the library from above and Vue JS and release a: Web, Electron (Desktop) and a mobile versions.

Additional notes:

I've tested this only by myself on the local network, I need more people to test it out.
I'm using a fork of WebSocketSharp by sta: https://www.nuget.org/packages/SIPSorcery.WebSocketSharp/0.0.1?_src=template

Here is a screenshot:
image

### DIFF
--- a/MinecraftClient/Resources/config/MinecraftClient.ini
+++ b/MinecraftClient/Resources/config/MinecraftClient.ini
@@ -258,5 +258,6 @@ backupinterval=300                # How long should replay file be auto-saved, i
 # (!) it is recommended to protect your Web Scoket instance with a password.
 # /!\ For the time being this should be considered as an experimental feature susceptible to change
 enabled=false
+bind_ip=127.0.0.1                     # An IP address to bind the server on (default is localhost: 127.0.0.1) NOTE: Doesn't support domain names or localhost string, just IP addresses
 port=8043                        # Web Socket port (0-65535)
 password=                        # Password, set if you want to protect against unauthorized access
--- a/MinecraftClient/Resources/config/MinecraftClient.ini
+++ b/MinecraftClient/Resources/config/MinecraftClient.ini
@@ -247,3 +247,16 @@ items=                            # separate each item name with a comma ',': It
 # /!\ You SHOULD use /replay stop or exit the program gracefully with /quit OR THE REPLAY FILE MAY GET CORRUPT!
 enabled=false
 backupinterval=300                # How long should replay file be auto-saved, in seconds. Use -1 to disable
+
+[WebSocket]
+# Enables you to remotely control the MCC over Web Socket
+# Get real-time data from the client that you could use for external processing and decision making
+# Example usage could be writting chat bots in other languages (although for the time being this is limited in functionality)
+# Technical details:
+#	1. The route for the Web Socket server is: /mcc (Example connection string: "ws://mydomain.com:8043/mcc")
+#	2. If you set a password, you're required to log in with the following command: Authenticate yourpasswordhere
+# (!) it is recommended to protect your Web Scoket instance with a password.
+# /!\ For the time being this should be considered as an experimental feature susceptible to change
+enabled=false
+port=8043                        # Web Socket port (0-65535)
+password=                        # Password, set if you want to protect against unauthorized access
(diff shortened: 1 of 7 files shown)


milutinke:
The JS library is almost done
image

milutinke:
PS: Not ready for merge yet, I still need to add an ability to execute ChatBot functions.
Only opened a PR so people can test it out and give a feedback.

milutinke:
Added a functionality for RPC, so basically people can execute methods from the ChatBot over websocket asynchronously, the tracking will be done by the JS library.
I still need to enable all useful methods and test out everything.

MinecraftClient/Resources/config/MinecraftClient.ini
```diff
@@ -247,3 +247,16 @@ items=                            # separate each item name with a comma ',': It
...
+# Enables you to remotely control the MCC over Web Socket
+# Get real-time data from the client that you could use for external processing and decision making
+# Example usage could be writting chat bots in other languages (although for the time being this is limited in functionality)
+# Technical details:
+#	1. The route for the Web Socket server is: /mcc (Example connection string: "ws://mydomain.com:8043/mcc")
+#	2. If you set a password, you're required to log in with the following command: Authenticate yourpasswordhere
+# (!) it is recommended to protect your Web Scoket instance with a password.
+# /!\ For the time being this should be considered as an experimental feature susceptible to change
+enabled=false
+port=8043                        # Web Socket port (0-65535)
```
ORelio, line 261: I'd suggest adding a bind address setting as well and defaulting to `127.0.0.1` to make sure no remote access is possible by default.

ORelio (commented):
Looks very promising! Indeed, that requires duplicating the API (implementing each method from ChatBot.cs and adding new methods here as well) but this is a good idea to add a standard way to communicate with MCC from another program written in any language 👍

I've made a quick security-related comment:

MinecraftClient/Resources/config/MinecraftClient.ini
milutinke: Yeah, that is a good Idea.

milutinke:
> Looks very promising! Indeed, that requires duplicating the API (implementing each method from ChatBot.cs and adding new methods here as well) but this is a good idea to add a standard way to communicate with MCC from another program written in any language 👍
> 
> I've made a quick security-related comment:

I am not that familiar with all things that C# offers with reflection like in Java.
I had an idea to make schemas for checking the incoming procedures/method calls, then executing them and auto convert parameters to types that methods need too, but I have no C# knowledge on this level yet. Seems kinda tricky with auto converting types to proper ones, so I opted for a manual way.
Do you have any ideas on this?

ORelio:
I think it is feasible like in Java, but I'm not sure how you would convert complex classes like Entity, maybe with JSON serialization 🤔

milutinke:
The Bot is almost done, I need to test all methods and fix if something isn't working, then I need to finish the JS library and document everything.

MinecraftClient/Resources/config/MinecraftClient.ini
milutinke: Done

milutinke:
I'll most likely drop the front end support for the time being since I could not find a way to solve the issue with code loading on the front end, I'll most likely use some bundler later for the front end version of the JS library.
Also, might close this PR in favor of a new one to avoid dealing with a bunch of merge conflicts, there are quite a lot because the branch is really dated.

milutinke:
Waiting for toml branch to merge.

BruceChenQAQ:
> Waiting for toml branch to merge.

Ready to merge it, some people have been using the pre-release version for a while.
(Manually releasing a release is a bit of a pain)

PR #2167, opened 2022-09-07
Added a /reload command for reloading the settings.

milutinke:
Added a /reload command for reloading the settings.
**NOTE: I've only tested some changes on a local server, haven't tested all Chat Bots, I need more people to test out chat bots with the `/reload` command.**

To test, you can clone my branch with the following command:
```bash
git clone https://github.com/milutinke/Minecraft-Console-Client --recursive -b reload-config
```

Here you can find out how to install Git and how to build the project from the source (Make sure to use my clone command above, not the documentation one): https://mccteam.github.io/docs/guide/installation.html#building-from-the-source-code

### DIFF
--- a/MinecraftClient/ChatBots/Alerts.cs
+++ b/MinecraftClient/ChatBots/Alerts.cs
@@ -19,14 +19,13 @@ namespace MinecraftClient.ChatBots
         /// </summary>
         public override void Initialize()
         {
-            dictionary = LoadDistinctEntriesFromFile(Settings.Alerts_MatchesFile);
-            excludelist = LoadDistinctEntriesFromFile(Settings.Alerts_ExcludesFile);
+            Setup();
         }
 
         /// <summary>
         /// Update settings when reloaded
         /// </summary>
-        public override void OnSettingsReloaded()
+        public override void OnSettingsReload()
         {
             if (!Settings.Alerts_Enabled)
             {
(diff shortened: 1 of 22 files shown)


milutinke:
Is it maybe better to unload all bots and load them again?

ORelio:
Nice work! It would likely not be as reliable as a true restart but could come handy.
How about calling `OnSettingsReloaded()` when initializing the bot to avoid duplicating code?
(I'd suggest `OnSettingsReload` for the event name)

milutinke:
> Nice work! It would likely not be as reliable as a true restart but could come handy. How about calling `OnSettingsReloaded()` when initializing the bot to avoid duplicating code? (I'd suggest `OnSettingsReload` for the event name)

In most of the bots I have a separate method Setup which I use in Initialize and OnSettingsReloaded, so there is no need in calling it, also in some bots I need to do a check if they've been disabled, which would be redundant on `Initialize`.
I'll rename the method.

milutinke:
I've added a possibility to hard reload bots using `/reload hard`, this will first unload all bots, then load enabled ones, and then it will call `AfterGameJoined` on all of them if the client is still connected.

milutinke:
Added a command to list loaded bots and to unload a specific bot or all bots.

milutinke:
Can someone test this on a live server and report if everything is working fine, I currently don't have much time.

Manual on how to build yourself: https://mccteam.github.io/docs/guide/installation.html#building-from-the-source-code

Just use this command to clone my repo and branch instead the one from the documentation:
```bash
git clone https://github.com/milutinke/Minecraft-Console-Client --recursive -b eload-config
```

milutinke:
Testing (Checked tested and works):
- [x] Alerts 
- [x] Auto Attack
- [x] Anti AFK
- [x] Auto Craft
- [x] Auto Eat
- [x] Auto Drop
- [x] Auto Respond
- [x] Replay Mod (soft reload)
- [x] Remote Control

- Disabled hard reloading due to crashing:
- [x] Replay Mod

- Couldn't test:
- Mailer
- Fishing Bot

milutinke:
Opening a new PR.

Issue #2125, opened 2022-08-24
[Idea] Add follow player function


kvr25565:
### Prerequisites

- [X] I have read and understood the [user manual](https://github.com/MCCTeam/Minecraft-Console-Client/tree/master/MinecraftClient/config)
- [X] I made sure I am running the latest [development build](https://ci.appveyor.com/project/ORelio/Minecraft-Console-Client/build/artifacts)
- [X] I tried to [look for similar feature requests](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

### Console Client Version

1.19 latest

### Describe your problem

Following player/entity 

### Suggest a possible solution

Add /follow or /move follow command

### Minecraft Version

1.16.5

### Device

Desktop

### Operating System

Windows

deosync:
ye I need this

milutinke:
It's a cool idea.
I'll be looking into this after #2120, #2123, #2324 are merged.

milutinke:
Implemented in #2174

PR #2174, opened 2022-09-09
Implemented the Follow Player chat bot.

milutinke:
Added an ability to follow the player using the `/follow` command as requested in #2125
Command usage: `/follow <player name|stop> [-f]`

NOTE: The bot sometimes can be a bit slow, we need to implement a method that can cancel the current Walking task, that would speed it up.
I've added a default update timer to make the bot a bit more faster, it does not go where it doesn't need to, we need to figure out what setting is the most optimal one.
Did some basic testing, seems that it is working as intended, haven't had any issues, as far as I am asked, it's ready to go, tho I want to get some feedback/ideas to improve it.

The video (Sorry for the bad quality an update to the OBS wiped my settings): 
https://streamable.com/g2gdl4

### DIFF
--- a/MinecraftClient/ChatBots/FollowPlayer.cs
+++ b/MinecraftClient/ChatBots/FollowPlayer.cs
@@ -1,0 +1,175 @@
+﻿using System;
+using System.Linq;
+using MinecraftClient.Mapping;
+
+namespace MinecraftClient.ChatBots
+{
+    public class FollowPlayer : ChatBot
+    {
+        private string? _playerToFollow = null;
+        private int _updateCounter = 0;
+        private int _updateLimit;
+        private int _stopAtDistance;
+        private bool _unsafeEnabled = false;
+
+        public FollowPlayer(int updateLimit = 15, int stopAtDistance = 3)
+        {
+            _updateLimit = updateLimit;
+            _stopAtDistance = stopAtDistance;
+        }
+
+        public override void Initialize()
+        {
+            if (!GetEntityHandlingEnabled())
+            {
+                LogToConsoleTranslated("extra.entity_required");
+                LogToConsoleTranslated("general.bot_unload");
+                UnloadBot();
+                return;
+            }
+
+            if (!GetTerrainEnabled())
+            {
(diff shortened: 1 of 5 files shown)


Splarkszter:
How can i test this script?

Splarkszter:
I think the bot should have this parameters:
- Threshold: How much the bot has to be from the current followed entity to stop walking, with both vertical and horizontal thresholds and with it's own update rate setting.
- Update rate: How often the path to the current followed entity is updated.

Update rate i think would be good to be between 100ms to 200ms default if it's not too heavy for the CPU, that will need to be tested. Maybe a more refined way is to choose from the last destination to reach to calculate the path and queue them (hopefully without memory leaks) so the bot does not stop walking unless it is inside the threshold.

Also, i don't know if those strange head movements can be fixed (But i think that is more for the TerrainAndMovements maintainer).

milutinke:
> I think the bot should have this parameters:
> 
>     * Threshold: How much the bot has to be from the current followed entity to stop walking, with both vertical and horizontal thresholds and with it's own update rate setting.
> 
>     * Update rate: How often the path to the current followed entity is updated.
> 
> 
> Update rate i think would be good to be between 100ms to 200ms default if it's not too heavy for the CPU, that will need to be tested. Maybe a more refined way is to choose from the last destination to reach to calculate the path and queue them (hopefully without memory leaks) so the bot does not stop walking unless it is inside the threshold.
> 
> Also, i don't know if those strange head movements can be fixed (But i think that is more for the TerrainAndMovements maintainer).

It already does have those 2.

You can clone it and build it yourself, here is the manual:

Use this command for cloning instead the one from the documentation: https://mccteam.github.io/docs/guide/installation.html#building-from-the-source-code
```bash
git clone https://github.com/milutinke/Minecraft-Console-Client/ --recursive -b follow-player
```

milutinke:
If there are no complaint or suggestions on improving, I'll be merging in few days, it's in a decent state.
We could improve it by improving the Terrain Handling code, like an ability to cancel a running task when a new one is needed.

PR #2180, opened 2022-09-11
Added a way to execute multiple commands with a single command

milutinke:
Added a way to execute multiple commands with a single command with:
```
/execmulti <command 1> -> <command 2> -> <command 3> -> ...
```

Another idea, we could implement a field in `PerformInternalCommand` which would indicate a failure of the execution, so we could make a `/chain` command which would stop if one of the commands has failed to execute, this could be useful in some scenarios, tho this would be a bit tricky to implement and to keep backwards compatibility.

### DIFF
--- a/MinecraftClient/Commands/ExecIf.cs
+++ b/MinecraftClient/Commands/ExecIf.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using DynamicExpresso;
+
+namespace MinecraftClient.Commands
+{
+    class ExecIf : Command
+    {
+        public override string CmdName { get { return "execif"; } }
+        public override string CmdUsage { get { return "execif <condition/expression> ---> <command>"; } }
+        public override string CmdDesc { get { return "cmd.execif.desc"; } }
+
+        public override string Run(McClient handler, string command, Dictionary<string, object> localVars)
+        {
+            if (hasArg(command))
+            {
+                string commandsString = getArg(command);
+
+                if (!commandsString.Contains("--->"))
+                    return GetCmdDescTranslated();
+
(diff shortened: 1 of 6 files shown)


ORelio:
I'm not sure what is this for? Bots like AutoRespond which runs a single command on a match? To avoid creating a script?
The main difficulty with this will be handling quoted arguments, this would become messy really quick.
(I'm not against adding this, just questioning).

milutinke:
> I'm not sure what is this for? Bots like AutoRespond which runs a single command on a match? To avoid creating a script? The main difficulty with this will be handling quoted arguments, this would become messy really quick. (I'm not against adding this, just questioning).

Just created it if someone needs to execute multiple commands in directly in the console, useful when testing, you can write a long command somewhere and then just copy and paste it a whole instead of manually typing a lot of commands every time.

Also added a conditional command, `/execif <condition/c# expression> ---> <command>`
It can use the variables from the ini file and it has an access to the McClient class.
It might come handy for some very basic checking in text scripts.

milutinke:
Should we add this?

BruceChenQAQ:
I think it's okay to add these.
Would it be better to require the condition in command A to return a boolean, since there may be ambiguities in the treatment of integers, etc.

milutinke:
> I think it's okay to add these. Would it be better to require the condition in command `/execif` to return a boolean, since there may be ambiguities in the treatment of integers, etc. `integer > 0` or `integer != 0` `string != null` or `!string.IsNullOrEmpty()` or `!string.IsNullOrWhiteSpace()`

Yeah, found a way to make it return only boolean.
Fixed.

milutinke:
I did some test once more, I'll merge, it's useful for debugging sometimes and the `/execif` can come in handy in txt script.

PR #2189, opened 2022-09-17
Fully implemented Map Data packet + Fixed 1.14 not working.

milutinke:
This will allow us to process map data and render maps into images with a chat bot, so people can solve those annoying captchas.
However, we need to figure out how the maps are being rendered,

Packet info: https://wiki.vg/Protocol#Map_Data
Map Data format: https://minecraft.fandom.com/wiki/Map_item_format (*NOTE: Colors on the wiki are outdated, we will have to look at the code*)
Minecraft Classes:
 - Packet class: `net.minecraft.network.protocol.game.ClientboundMapItemDataPacket`
 - Map Data class: `net.minecraft.world.level.saveddata.maps.MapItemSavedData`

Found some libraries for reference (a bit dated tho):
 - https://github.com/LB--/MCModify/blob/java/src/main/java/com/lb_stuff/mcmodify/minecraft/Map.java
 - https://github.com/spookymushroom/minecraftmap/blob/master/minecraftmap/__init__.py
 - https://github.com/joodicator/mcmapimg/blob/master/mcmapimg/mcmapimg.py

My proposal is making a chat bot that will hold a cache of all maps being sent, and then when you need to see them, you can render all, a single map or newest N maps.
As for the term `rendering` I am referring to rendering and saving images in a folder, rather than trying to display them in the console somehow.

### DIFF
--- a/MinecraftClient/Mapping/MapIcon.cs
+++ b/MinecraftClient/Mapping/MapIcon.cs
@@ -1,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using MinecraftClient.Protocol.Handlers;
+
+namespace MinecraftClient.Mapping
+{
+    public class MapIcon
+    {
+        public MapIconType Type { set; get; }
+        public byte X { set; get; }
+        public byte Z { set; get; }
+        public byte Direction { set; get; }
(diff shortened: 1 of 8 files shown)


milutinke:
I'll merge and open a proper issue for the chat bot for discussion.

milutinke:
Fixed 1.14 not working at all.
Tested the packet on all versions from 1.8 +
Ready for merge.

PR #2200, opened 2022-09-25
Reload Settings and Chat Bots, Fixed PlayerListLogger Chat Bot, added %datetime%

milutinke:
This is a new PR for #2167, the reason for a new PR is that I do not want to deal with all the merge conflicts, since the version of the code I've worked in is quite outdated.

`/reload` is now only hard reload, no more soft reload (It unloads all Chat Bots and loads them back in by default).
Other command are unchanged.
All Chat Bots are now back to their original state.
Tested everything except Mailer and it works, I assume that Mailer Will work too, I do not see a reason for it not to work.

The PR is ready for merge as far as I am considered, going to leave it a bit for potential feedback.


milutinke:
Fixed the PlayerListLogger Chat Bot and added configuration section for it.

PR #2218, opened 2022-09-28
Updated Anti AFK Bot to have Terrain Handling and random intervals option

milutinke:
Updated the Anti AFK Chat Bot to have an option of using the terrain handling and to use sneaking when using the old command method.
Note: The terrain handling option is disabled by default.

Now the chat bot can walk randomly in the given range, if it fails to get to a randomly generated location it will try 20 more times before using the old command method, in addition to sending a command now the bot also does toggle sneak.
Also, now the bot supports random intervals, example: `delay=30-200`

**NOTE:**
If you're using Terrain Handling feature, make sure to enclose the bot within an enclosure of some size to prevent it from wandering off randomly, the recommended size is 5x5x5.

### DIFF
--- a/MinecraftClient/ChatBots/AntiAFK.cs
+++ b/MinecraftClient/ChatBots/AntiAFK.cs
@@ -42,6 +42,14 @@ namespace MinecraftClient.ChatBots
                     useTerrainHandling = false;
                     LogToConsole(Translations.TryGet("bot.antiafk.not_using_terrain_handling"));
                 }
+                else
+                {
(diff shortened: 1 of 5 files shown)


milutinke:
Leaving until tomorrow at this time for potential feedback.

ReinforceZwei:
How about saving the original location the bot is at and limit the range within that location?

milutinke:
Actually that would interfere if someone would like to move a bot somewhere with the move command, then it would not work with the movement.

PR #2219, opened 2022-09-28
Added a command for easier usage of the bed (/bed), Added relative coordinates support for the /dig command.

milutinke:
Added a new command for usage of the bed:
- Sleep in a bed: `/bed sleep <x> <y> <z>` (Supports relative coordinates, eg. `/dig 124 ~ ~-1`) (Shorter: `/bed s <x> <y> <z>`)
- Leave a bed: `/bed leave` (Shorter: `/bed l`)

Added relative coordinates support for the  `/dig` command.

### DIFF
--- a/MinecraftClient/ChatBots/Map.cs
+++ b/MinecraftClient/ChatBots/Map.cs
@@ -33,7 +33,7 @@ namespace MinecraftClient.ChatBots
             deleteAllOnExit = Settings.Map_Delete_All_On_Unload;
             notifyOnFirstUpdate = Settings.Map_Notify_On_First_Update;
 
-            RegisterChatBotCommand("maps", "bot.map.cmd.desc", "maps <list/render <id>> | maps <l/r <id>>", OnMapCommand);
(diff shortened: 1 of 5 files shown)


breadbyte:
It would be a nice addition if the `/bed sleep` command, without arguments, would simply go to the nearest bed within reach of the player, removing the need to specify exact coordinates for the bed.

milutinke:
> It would be a nice addition if the `/bed sleep` command, without arguments, would simply go to the nearest bed within reach of the player, removing the need to specify exact coordinates for the bed.

Going to implement it tomorrow.
I was thinking to add it as a separate chat bot after this PR is merged, but it's more intuitive to be in the command itself.

milutinke:
> It would be a nice addition if the `/bed sleep` command, without arguments, would simply go to the nearest bed within reach of the player, removing the need to specify exact coordinates for the bed.

Tried implementing but I've failed, I do not know how to make the bot wait until it gets to the bed location, if I make a while loop it will freeze the client and it will not move until the timeout has been reached, I'll merge it like this. If you know how to pull it off you can do it.  I am not that familiar with threads, especially in C# as I almost never use them as a web dev.

milutinke:
Everything tested and working. Going to merge.

PR #2221, opened 2022-09-29
Update builds to be self-contained

breadbyte:
Currently, we are using the `framework-dependent` method of building our binaries. This means that the user has to install .NET 6 before they are able to use the app but has the added benefit of reducing the final binary's file size.

Making a self-contained app would help reduce friction with previous users that are used to just running the executable without anything to install, as opposed to having to install .NET 6.0 on their system. 

(The previous version was running on .NET Framework, and .NET Framework is pre-installed on all Windows installs starting with Windows 7, so no additional installation is required (except when updating to a newer version of the framework.))

This comes with the unfortunate downside of massively increasing the file size (as we are also bundling .NET alongside our app.)

We would need to move our compression from `.zip` to `.7z` (which is also widely supported by various compression tools (being an open format) and gives us better compression) to reduce the download size.

For reference, the size of the binaries that we currently provide are (using build 20220928-75 as the baseline):

| Binary type | Build type | File size |
| ------------- | ------------- | ------------- | 
| MinecraftClient.exe | framework-dependent binary | 27.9mb |
| MinecraftClient-windows.zip | framework-dependent zipped | 8.5mb |
| **MinecraftClient-windows.7z** | **framework-dependent 7zipped** | **5.8mb** |
| MinecraftClient.exe | self-contained binary | 90.2mb |
| MinecraftClient-windows.zip | self-contained zipped | 35.7mb |
| **MinecraftClient-windows.7z** | **self-contained 7zipped** | **26.4mb** |

Should we continue providing the binary in `zip` format or should we move to `7z`?
The additional space savings using 7z may not be worth it if the user has to install a decompression tool.

More than likely, on users' desktops, they already have a decompression tool that supports 7z (WinRAR, 7zip, PeaZip, etc), but on servers (VPSes, Virtual Machines, Docker, etc.) they more than likely have to install a separate tool.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -7,7 +7,7 @@ on:
 env:
   PROJECT: "MinecraftClient"
   target-version: "net6.0"
-  compile-flags: "--no-self-contained -c Release -p:UseAppHost=true -p:IncludeNativeLibrariesForSelfExtract=true -p:DebugType=None"
+  compile-flags: "--self-contained=true -c Release -p:UseAppHost=true -p:IncludeNativeLibrariesForSelfExtract=true -p:DebugType=None"
 
 jobs:
   build:


ORelio:
Good idea! Why not provide all options? Zip/7z Framework & Standalone?
Would need very meaningful archive names but apart from that I see no particular downside 🤔

BruceChenQAQ:
How about compress with 7z and creating it as a self-extracting file?

milutinke:
> Good idea! Why not provide all options? Zip/7z Framework & Standalone? Would need very meaningful archive names but apart from that I see no particular downside 🤔

We could add _Compressed suffix, so the people would not be confused with which one to download.

breadbyte:
> How about compress with 7z and creating it as a self-extracting file?

Good idea. I could do that for the windows binaries.

For Linux and Mac, I think it would be better to just provide a generic .zip file for greater compatibility.

BruceChenQAQ:
`(Full)MinecraftClient-windows.exe`
`(Full)MinecraftClient-osx.???`
`(Full)MinecraftClient-linux.sfx`
`(Full)MinecraftClient-linux-arm64.sfx`
`(Without-Framework)MinecraftClient-Windows.exe`
`(Without-Framework)MinecraftClient-osx.???`
`(Without-Framework)MinecraftClient-linux.7z`
`(Without-Framework)MinecraftClient-linux-arm64.7z`

breadbyte:
I would prefer having only the self-contained version and dropping the framework-dependent version to reduce confusion over multiple versions.

I haven't had time to work on this (and my other outstanding issues) at the moment as I just got back from a week-long trip.

breadbyte:
I've decided to just turn on self-contained builds without any change to the current distribution method that we are using.

Issue #2210, opened 2022-09-27
[Question] A ChatBot (cs script) for mine block which I was looking at.


Sergapr:
Hi guys! Do u have a script to mine block which bot was looking at?
I have a script which mining a block which above player, or in front of player, but it is uncomfortable to use.
Thanks a lot!

deosync:
use for  please

milutinke:
The [ChatBot API](https://github.com/MCCTeam/Minecraft-Console-Client/blob/master/MinecraftClient/Scripting/ChatBot.cs) does not contain a method to get a block where the player is looking at currently, but I think it could be done with [`GetYaw()`](https://github.com/MCCTeam/Minecraft-Console-Client/blob/aec38d83c767cf4fff64a6ad2b7dfcebb076e6ab/MinecraftClient/Scripting/ChatBot.cs#L1114) and [`GetPitch()`](https://github.com/MCCTeam/Minecraft-Console-Client/blob/aec38d83c767cf4fff64a6ad2b7dfcebb076e6ab/MinecraftClient/Scripting/ChatBot.cs#L1123) methods.
Otherwise you can use [`GetCurrentLocation()`](https://github.com/MCCTeam/Minecraft-Console-Client/blob/aec38d83c767cf4fff64a6ad2b7dfcebb076e6ab/MinecraftClient/Scripting/ChatBot.cs#L987) to dig a block where the bot is standing.

BruceChenQAQ:
![image](https://user-images.githubusercontent.com/26017411/193409276-4b0fdcee-0c08-492b-a3ea-458d24e75b89.png)
![image](https://user-images.githubusercontent.com/26017411/193409417-c9260dd3-5ca2-4bcd-8029-64d2c3e8bfcd.png)

A simple version is implemented.
Since MCC currently cannot handle hitboxes precisely, all blocks are considered complete cubes.

BruceChenQAQ:
Implemented in https://github.com/MCCTeam/Minecraft-Console-Client/pull/2227.

BruceChenQAQ:
This PR has been merged, just use `GetLookingBlock` in the script to get the block position.

PR #2227, opened 2022-10-01
Support finding the blocks on the line of sight.

BruceChenQAQ:
* Use `/look` to get the block the player is currently looking at.
* Using `/dig` will dig the block the player is looking at.
* New ChatBot API
  * `Tuple<bool, Location, Block> GetLookingBlock(double maxDistance = 4.5, bool includeFluids = false);`

resolves https://github.com/MCCTeam/Minecraft-Console-Client/issues/2210


ReinforceZwei:
Actually there is a method for doing the similar thing. Is there difference between that method and this PR?
https://github.com/MCCTeam/Minecraft-Console-Client/blob/5eb1ffde480add18ff9a3e54bd2cf91e771fafca/MinecraftClient/Mapping/World.cs#L229

BruceChenQAQ:
> Actually there is a method for doing the similar thing. Is there difference between that method and this PR?
> 
> https://github.com/MCCTeam/Minecraft-Console-Client/blob/5eb1ffde480add18ff9a3e54bd2cf91e771fafca/MinecraftClient/Mapping/World.cs#L229

The version implemented in this PR is able to control the maximum distance precisely.
And `GetLookingBlockLocation` will give wrong results in some cases.

PR #2267, opened 2022-10-12
Added a Farmer Bot

milutinke:
Made a bot that can **farm crops** like Baritone but better.

**Features:**
 - Planting crops
 - Collecting/Breaking crops
 - Bonemealing crops
 - Using an axe to collect/break pumpkins/melons

**Supported types:**
 - Beetroot
 - Carrot
 - Melon
 - Netherwart
 - Pumpkin
 - Potato
 - Wheat

**A video demo (Note: This is a way older version, but you get the point):**
https://streamable.com/yzqnuy

**Current limitations/problems:**
 - Sometimes the bot jumps on to the farmland when farming on different heights or when coming to the farmland from some other block height (this case happens rarely but still happens).
 - Sometimes the bot can get kicked by the server when farming on soul sand for "invalid movement" packets, tho this is dependent on the part of the world, I haven't been able to figure out what exactly causes this, needs further investigation.
 - Sometimes the bot won't bone meal carrots/potatoes or pumpkin/melon stems, I've tried providing the exact block heights but with no success. The interesting thing is that it's a pattern, no matter the configuration of farmland it happens.
 - Does not work bellow `1.13`.
 - The bot doesn't pickup items if they fly too far away.

**Ideas for features:**
- We could add a way for bot to drop items off to a chest or multiple chests and a way to refill bone meal from a chest.
  
  For this we need an utility function that checks if the chest is empty or that is has enough space for the items in that need to be transferred, also an utility function for picking up certain item type and a specified amount of it.

- We could add a stage for picking up items after the bone meal stage, this requires a utility function to check whether the inventory is full.

**Usage:**
 
 - Starting: 
 
   ```/farmer start <crop type> [radius:<radius = 30>|unsafe:<true/false>|teleport:<true/false>|debug:<true/false>```
   
   Example:
   ```
   /farmer start wheat radius:50 debug:true unsafe:true
   ```

 - Stopping: `/farmer stop`
 
 **NOTE: Tested only on 1.19+ (Needs testing on older versions)**


AbdobiZark:
is cocoa possible to add?

milutinke:
> is cocoa possible to add?

That would needed to be implemented as a custom bot, because all the code here is written for crops that are planted on farmland.

PR #2282, opened 2022-10-17
Updated Farmer bot and Enchant command translations

milutinke:
Updated the command description to make it more readable and translated 2 remaining strings.


BruceChenQAQ:
Also missing this translation: `cmd.enchant.desc`

milutinke:
> Also missing this translation: `cmd.enchant.desc`

I thought I've added that.
I'll add it.

milutinke:
Thanks for the note.
I'll make the command better now with more safe guards.

PR #2303, opened 2022-10-24
Turkish language support

ByDexterTR:
Because I was sick, I waited for recovery, but my recovery took longer than I expected, so I did it in a hurry, there may be inaccurate/incomplete translations, but I was as careful as possible. (I didn't translate some terms on purpose, because they are well-known and common terms)

### DIFF
--- a/MinecraftClient/Resources/lang/tr.ini
+++ b/MinecraftClient/Resources/lang/tr.ini
@@ -1,9 +1,9 @@
 [mcc]
 # Messages from MCC itself
-mcc.help_us_translate=MCC'yi çevirmemize yardımcı olun: {0}
+mcc.help_us_translate=MCCyi çevirmemize yardımcı olun: {0}
 mcc.run_with_default_settings=\nMCC varsayılan ayarlarla çalışıyor.
 mcc.settings_generated=§cAyarlar dosyası MinecraftClient.ini oluşturuldu.
-mcc.has_update=§eMCC'nin yeni sürümü mevcut: {0}
+mcc.has_update=§eMCCnin yeni sürümü mevcut: {0}
 mcc.login=Login :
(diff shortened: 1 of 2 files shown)


milutinke:
I've translated the text to English and did a diff with the original, everything seems legit, as far as I concerned you have my approval.
Do you want to add something else or it's ready for merge?
By the way, thank your for your hard work, this requires a lot of time.

milutinke:
Also, If you are willing, you can also updated the documentation website, although it has much more text than the ini file.
If you are, I can setup the necessary code for the translation, ping me if you're willing.

BruceChenQAQ:
I've changed the code and after deleting the settings file and re-running MCC should automatically switch to use Turkish.
In the master branch, some texts related to WebSocketBot has been added, would you like to translate it along the way, or just merge the existing version.

ByDexterTR:
> Also, If you are willing, you can also updated the documentation website, although it has much more text than the ini file. If you are, I can setup the necessary code for the translation, ping me if you're willing.

That would be great, looks better than calling one by one to see what's changed

milutinke:
> > Also, If you are willing, you can also updated the documentation website, although it has much more text than the ini file. If you are, I can setup the necessary code for the translation, ping me if you're willing.
> 
> That would be great, looks better than calling one by one to see what's changed

The merge was reverted due to a problem with the library, there is no need to translate websocket yet.
If you agree I'll merge this.

ByDexterTR:
> > > Also, If you are willing, you can also updated the documentation website, although it has much more text than the ini file. If you are, I can setup the necessary code for the translation, ping me if you're willing.
> > 
> > 
> > That would be great, looks better than calling one by one to see what's changed
> 
> The merge was reverted due to a problem with the library, there is no need to translate websocket yet. If you agree I'll merge this.

It's okay

PR #2349, opened 2022-11-23
Fixes the bug in entity command that doesn't allow passing entity types as second argument

### LINKED ISSUE #2345: [BUG] Running the 'entity' command using an entity type will always result in getting the command description
### Prerequisites

- [X] I made sure I am running the latest [development build](https://github.com/MCCTeam/Minecraft-Console-Client/releases/latest)
- [X] I tried to [look for similar issues](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one
- [X] I have set `debugmessages=true` in config to diagnose my issue
- [X] I have redacted session tokens and passwords before attaching screenshots

### Minecraft Version

1.19

### Console Client Version

Latest

### Expected Behavior

The expected behavior would be that the player will use/attack all the entities with the given type.

### Actual Behavior

The description of the command gets printed to the screen.

### Steps to Reproduce the bug

1. Connect to any server
2. Run the entity command using an entity type making sure the an entity with that type is around the bot

_No response_

### Anythings that could help diagnosing the bug

```txt
I think the issue comes mainly from an error in the Entitycmd class.

In the Run method we see that the first argument of the command gets parsed using the int.Parse method and if it throws a FormatException and then the command description gets printed. Instead what it should be doing is trying to parse that argument as a field of the EntityType enum as it it's done in the else statement when it checks if the entity ID is 0.

Also this check seems kind of useless and it probably could be removed.
```

### Device

Desktop

### Operating System

Windows

### Server Address (If applicable)

_No response_

ITHackerstein:
Fixes #2345.

Now the command works as expected and when the type of interactions isn't passed and the second argument is an entity type we show a list of entities of that type.

### DIFF
--- a/MinecraftClient/Commands/Entitycmd.cs
+++ b/MinecraftClient/Commands/Entitycmd.cs
@@ -126,7 +126,16 @@ namespace MinecraftClient.Commands
                     }
                     catch (FormatException)
                     {
-                        EntityType interacttype = Enum.Parse<EntityType>(args[0]);
+                        EntityType interacttype;
+                        try
+                        {
+                            interacttype = Enum.Parse<EntityType>(args[0]);
+                        }
+                        catch (Exception)
+                        {
+                            return Translations.cmd_entityCmd_not_found;
+                        }
+
                         string action = args.Length > 1
                         ? args[1].ToLower()
                         : "list";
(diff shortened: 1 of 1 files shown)


milutinke:
You should add try catch for `EntityType interacttype = Enum.Parse<EntityType>(args[0]);` because if someone uses a wrong name it will crash the client.

ITHackerstein:
I've added a try-catch statement around the `Enum.Parse` instruction to avoid throwing an exception when running the command with an invalid entity type.

BruceChenQAQ:
Thanks for your contribution

PR #2353, opened 2022-11-28
entitynear: New command to interact with the entity closest to the player.

### LINKED ISSUE #2346: [Idea] Add the possiblity for the entity command to use/attack the nearest entity if using the entity type
### Prerequisites

- [X] I have read and understood the [user manual](https://mccteam.github.io/guide/)
- [X] I made sure I am running the latest [development build](https://github.com/MCCTeam/Minecraft-Console-Client/releases/latest)
- [X] I tried to [look for similar feature requests](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

### Console Client Version

Latest

### Describe your problem

The entity command when called with an entity type it uses/attacks all entities with the type specified. 

### Suggest a possible solution

What I'd like is for the command to be able to choose the closest and use/attack that one.

_No response_

### Minecraft Version

1.19

### Device

Desktop

### Operating System

Windows

ITHackerstein:
Closes #2346.

As the title says this new command allows the user to interact with entity closest to him.
The syntax is the same as the 'entity' command: it expects an optional entity type follwed by the type of action you want to perform (use/attack).
If no entity type and no action is specified then the data of the closest entity will be printed to the screen.
If an entity type is specified but no action is specified then the data of the closest entity of that type will be printed to the screen.


ITHackerstein:
Yeah sure, that makes sense, I don't know why I didn't think about that.

I'll work on it as soon as I can.

MinecraftClient/Commands/Entitycmd.cs
```diff
@@ -160,12 +188,40 @@ public override string Run(McClient handler, string command, Dictionary<string,
...
+                        if (args.Length > 1)
+                        {
+                            if (Enum.TryParse(args[1], true, out EntityType entityType))
+                            {
+                                if (TryGetClosetEntity(handler.GetEntities(), handler.GetCurrentLocation(), entityType, out Entity? closest))
+                                    return InteractionWithEntity(handler, closest, args.Length > 2 ? args[2].ToLower() : "list");
+                                else
+                                    return Translations.cmd_entityCmd_not_found;
+                            }
+                            else if (int.TryParse(args[0], NumberStyles.Any, CultureInfo.CurrentCulture, out int entityID2))
```
ITHackerstein, line 208: I think there is a typo here it should be
```csharp
int.TryParse(args[1], NumberStyles.Any, CultureInfo.CurrentCulture, out int entityID2)
```
otherwise it will always try to parse "near"

BruceChenQAQ: > I think there is a typo here it should be
> 
> ```cs
> int.TryParse(args[1], NumberStyles.Any, CultureInfo.CurrentCulture, out int entityID2)
> ```
> 
> otherwise it will always try to parse "near"

Yes, it's fixed in the latest version.

PR #2358, opened 2022-12-06
Implement command completion suggestions.

BruceChenQAQ:
(Waiting for https://github.com/breadbyte/ConsoleInteractive/pull/20 to be merged first)

* Switch to use [Brigadier.NET](https://github.com/AtomicBlom/Brigadier.NET/) as the command parser.
* **Implement a simulated pop-up window to display command completion suggestions on the console.**  Ability to correctly handle different widths of text and various foreground background colors.

* Use "//" as an alias of "/send".
image

* Fuzzy matching.
image

* Truncate overly long text.
image


BruceChenQAQ:
Localized Tooltips.

image

image

Splarkszter:
Can't use it on offline mode?

I did use - as password but it doesn't make it go offline mode, tried to connect to Microsoft servers and it fails to.
I do not have a spare Minecraft account so i have to use it on offline mode.

image

Splarkszter:
... i did have the gameport wrong... but still can't connect.
My client is on offline mode and i did the 'Open to LAN' thingy. It still seems that it doesn't enable it's offline mode on MCC?
image

Splarkszter:
Forget it, probably some user error, can't make it work on an older MCC version... strange.

PR #2378, opened 2022-12-15
Fixed entity yaw and pitch

DevBobcorn:
Fixed number value type during calculation of entity yaw & pitch


BruceChenQAQ:
Thanks for your contribution.

PR #2388, opened 2022-12-23
Update start-latest.sh

merlinlcb:
found that randomly after updating to the latest version execution was not being permitted on multiple linux devices running docker, this solves for that

### DIFF
--- a/Docker/start-latest.sh
+++ b/Docker/start-latest.sh
@@ -18,5 +18,8 @@ unzip MinecraftClient-linux.zip
 # Remove the ZIP
 rm -- MinecraftClient-linux.zip
 
+# Set Executable
+chmod +x /opt/start-latest.sh
+
 # Start the Client
 ./MinecraftClient


milutinke:
That might be an issue with the building script.
Does it work under normal linux?
Also, thank you for your contribution.

merlinlcb:
i ran into the same issue in a vm environment running Linux, so i'm not quite sure, but if its just an edge case it still shouldn't hurt to reapply executable permissions in this script to ensure its set, it ran just fine without this change on my daily driver Linux machine so its definitely not an issue for all devices and there doesn't seem to be a rhyme or reason for it, i made sure to update all the devices to latest release of either ubuntu/rhel/linux mint depending on what they had installed before testing with the latest pull and it was sporadically occurring

milutinke:
I thought I have set the permissions to executable when I made the script.
It might have something to do with cloning or containerization, so yeah, it won't do any damage if execute perms are added.

Issue #2384, opened 2022-12-18
[Idea] 1.19.3


UsainSrht:
### Prerequisites

- [X] I have read and understood the [user manual](https://mccteam.github.io/guide/)
- [X] I made sure I am running the latest [development build](https://github.com/MCCTeam/Minecraft-Console-Client/releases/latest)
- [X] I tried to [look for similar feature requests](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one

### Console Client Version

GitHub build 176, built on 2022-12-02 from commit 597af24

### Describe your problem

1.19.3

### Minecraft Version

1.19.3

### Device

Desktop

### Operating System

Windows

milutinke:
Already being worked on.

agitationswapping:
Any update on the progress/timeline and/or difficulties being faced?

Is there a candidate version that has chat only anywhere?

Apologize if this information is already somewhere else. Thanks so much for all your hard work!

JIBSIL:
> Any update on the progress/timeline and/or difficulties being faced?
> 
> Is there a candidate version that has chat only anywhere?
> 
> Apologize if this information is already somewhere else. Thanks so much for all your hard work!

It was partially completed in this PR: #2393 

I don't think the chat part is fully complete yet though.

BruceChenQAQ:
I'm working on extracting the protocol processing library from MCC into a separate repository, and I'll probably work on adapting 1.19.3 after I'm done with that, otherwise I may have to deal with this part of the code twice.

There is still a problem that we need to wait for the merge of https://github.com/MCCTeam/Minecraft-Console-Client/pull/2358 and https://github.com/breadbyte/ConsoleInteractive/pull/20. Since #2393 is developed based on the master branch and the latest build is based on the brigadier-dev branch, we need to wait for the brigadier-dev branch to be merged into the master branch. (breadbyte may have been busy recently and he hasn't reviewed pull request https://github.com/breadbyte/ConsoleInteractive/pull/20 yet.)

milutinke:
1.19.3 is out in a pre-release

JIBSIL:
> 1.19.3 is out in a from of a pre-release

It appears that chat signing is not functional in the pre-release (or it could be my settings). Testing on a 1.19.3 PaperSpigot server, no plugins, online mode + enforce-secure-profile on.

BruceChenQAQ:
[...]

PR #2405, opened 2023-01-10
AutoAttack: Add setting for attack range

ReinforceZwei:
> AutoAttack chat bot could use a distance/range option in the settings

First time contribute after switching to toml format. Please review my PR before merging.


BruceChenQAQ:
Moves the limit checking of a variable to be done when the setting is changed, and the change is to itself, so that the change to the value can be shown in the written back settings file.

Pull requests to the master branch now need to be marked "SKIP_BUILD" to skip the CI build, details of why can be found [here](https://github.com/MCCTeam/Minecraft-Console-Client/issues/2384#issuecomment-1374422175) :)

PR #2424, opened 2023-02-05
Update github build workflow [skip ci]

breadbyte:
This PR makes the following changes to the build and publish workflow:

- Replace [nanzm/get-time-action](https://github.com/nanzm/get-time-action) with the built-in linux command `date`.
- Remove the use of [actions/setup-dotnet](https://github.com/actions/setup-dotnet) since all github runners [already include](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#net-tools) the latest .NET SDK (7.0.102 is available at the Ubuntu 22.04 runner at this time, both are the latest as of this writing.)
- **Simplify the addition of new build targets.** Instead of having to write extensive boilerplate code, the build target can now be added to a list of build targets, and a release build with that target will automatically be generated.
	- As a benefit of streamlining the build process, we are now able to **build each target in parallel**, significantly reducing our build times.
- The translation keys are fetched and cached per commit build to save on time. The translation cache can be manually deleted before its TTL in the Actions area, under Caches. The TTL for Caches is 7 days.
	- An alternative to caching _per commit_ would be to cache _per day_, in which we could make a scheduled github action to update the translation cache every day.
- The build will fail when the translation keys cannot be fetched from Crowdin.
	- Given that we have the translation cache, we can simply use the last available cache. This would supplement the previous idea of a per-day update of the translation cache.
- The translation keys are now _no longer fetched_ whenever the target repository is not `MCCTeam`'s repository, as fetching the translations require Crowdin secrets. (this helps me debug the workflow and run my own builds)
- The `DOTNET_NOLOGO` environment variable is set to `true`. This disables the .NET Welcome messages in the logs.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -10,23 +10,38 @@ env:
   PROJECT: "MinecraftClient"
   target-version: "net7.0"
   compile-flags: "--self-contained=true -c Release -p:UseAppHost=true -p:IncludeNativeLibrariesForSelfExtract=true -p:EnableCompressionInSingleFile=true -p:DebugType=None"
+  ci-skip-message: "skip ci"
 
 jobs:
-  Build:
+  determine-build:
     runs-on: ubuntu-latest
-    if: ${{ !contains(github.event.head_commit.message, 'skip ci') }}
-    timeout-minutes: 15
-
+    strategy:
+      fail-fast: true
+    if: ${{ !contains(github.event.head_commit.message, env.ci-skip-message) }}
     steps:
-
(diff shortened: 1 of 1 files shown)


breadbyte:
Added the transition version into the builds.

This is the entire build matrix:
image

breadbyte:
I forgot I had this PR still open. Yes, it is mergeable.

ReinforceZwei:
The workflow for fetching translation failed. (https://github.com/MCCTeam/Minecraft-Console-Client/actions/runs/4476842516/jobs/7867644294)

PR #2452, opened 2023-03-25
FollowPlayer: Improve player detection

ReinforceZwei:
Putting limit counter at the topmost will cause it to miss many target player movement event. In my case it receive zero and doesn't move at all.

We should first filter out target player movement event before applying the limit counter so that the bot always receive target player movement.


milutinke (approved):
That was an oversight on my end, haven't thought of that.
Seems good to me.

PR #2473, opened 2023-04-08
Fixed crashing when having an OP on 1.19.4 (DeclareCommands fix)

milutinke:
A crash only happens if the player is an OP.
The server sends about 1300++ nodes in DeclareCommands packet.
Now, i am not sure if this is a bug with number of nodes on their side or not, because at one point they send a node with a name of `end`, after that one it crashes.
So, I've implemented a quick dirty fix that breaks out of the node reading loop, it works, but I am not sure if it prevents any more declared commands from being read.
From my quick investigation, the parsers and the packet reading code matches the wiki.
I'll not merge yet, if someone has some free time, please investigate further, I do not have more free time, neither I am that familiar with this part of the code.

### DIFF
--- a/MinecraftClient/Protocol/Handlers/Packet/s2c/DeclareCommands.cs
+++ b/MinecraftClient/Protocol/Handlers/Packet/s2c/DeclareCommands.cs
@@ -76,6 +76,11 @@ namespace MinecraftClient.Protocol.Handlers.packet.s2c
                 string? suggestionsType = ((flags & 0x10) == 0x10) ? dataTypes.ReadNextString(packetData) : null;
 
                 Nodes[i] = new(flags, childs, redirectNode, name, paser, suggestionsType);
+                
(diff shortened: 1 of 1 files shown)


ReinforceZwei:
I think something have changed in 1.19.4 and broke the parsing, as you can see it start reading garbage at the middle.
image
I will take a look into it.

milutinke:
> I think something have changed in 1.19.4 and broke the parsing, as you can see it start reading garbage at the middle. image I will take a look into it.

I can't debug with Rider on Linux, it's painful without it, thanks.
I should make a Windows VM for Visual Studio tbh.

ReinforceZwei:
I found that all `parser` are misspelled as `paser` (missing `r`). No wonder it looks a bit weird 🤣

ReinforceZwei:
I have fixed the crash and opened a new PR #2474

milutinke:
Awesome

PR #2496, opened 2023-05-06
Made Alert.cs TriggerByWords method case insensitive

Dariusel:
Made Chatbot.Alert TriggerByWords case insensitive. until now if you placed any string that contains an **uppercase** in "Matches" or "Excludes" array, it wouldn't be detected in chat This was because chat text would go trough .ToLower() before comparing and Matches array didn't.


ReinforceZwei:
Looks good to me, merging your PR.

ORelio:
That bot was initially case insensitive but seems like this was lost when migrating from INI to TOML. Good catch 👍

Issue #300, opened 2017-07-29
slabs are handled incorrectly


beppe9000:
i get a `Cannot move in that direction` error if i try to get on a slab with /move

ORelio:
slabs are viewed as full blocs, a special treatment must be added client-side to handle them properly

milutinke:
Good news,
Added with  #2504

PR #2504, opened 2023-05-16
Slab handling

milutinke:
Added slab handling and walking under slabs, this closes: #300
Currently supported: 1.13 - 1.19.4 (Tested on 1.19.4, should work on older versions)
Note: Trapdoors are not yet supported, but now can easily be supported with this code, it's only a matter of making a palette of trapdoor states for all versions.
I will do trapdoor support as a separate pr, after this one gets tested by the player base.

Video demonstration:
https://streamable.com/vuo519

### DIFF
--- a/MinecraftClient/Commands/Sneak.cs
+++ b/MinecraftClient/Commands/Sneak.cs
@@ -6,10 +6,9 @@ namespace MinecraftClient.Commands
 {
     public class Sneak : Command
     {
-        private bool sneaking = false;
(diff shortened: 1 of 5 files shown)


milutinke:
So far tested on 1.17.1 - 1.19.4

milutinke:
Tested, works on all versions.
PS: Removed 1.13 because you can't walk under slabs in 1.13 and bellow.

PR #2591, opened 2023-09-24
Docker: allow skip MinecraftClient download if MCC_SKIP_REDOWNLOAD defined

CAMOBAP:
Really useful during development/testing docker.
Simply pass MCC_SKIP_REDOWNLOAD


ReinforceZwei:
Thanks.

PR #2610, opened 2023-11-12
Add Yggdrasil Login

GLDYM:
## How to use
To use Yggdrasil, u need to change the following in MinecraftClient.ini.
```
AccountType = "Yggdrasil"
AuthServer = { Host = "littleskin.cn", Port = 443 }
```
And use USERNAME to login in.
## Problems
- May meeting some errors when trying to join a [Velocity](https://papermc.io/software/velocity) Server.
- When facing a multi-player account, the `selectedProfile` is None and waits the client to select from `availableProfiles`. If someone try to implement the selection of `selectedProfile`, it could login in with email. But I don't have enough time :-(
- I'm poor in C#, some logics may be strange.

Some code comes from https://github.com/Hikari1nVoid/MCC-authlib  ,but there are many errors in the original fork.


lizheyo77:
Littleskin.cn is not the only skin website, it cannot link to other skin websites

PR #2615, opened 2023-11-18
Add an IP lookup to the WebSocketBot

Zorua162:
To allow for running the WebSocket Bot easily in containers I wanted a way to be able to have the bot do a DNS lookup of a given string, this PR adds the desired functionality.

I have tested it with a few cases, including that the default "127.0.0.1" works (NOTE: I'm using my Python bindings to test this which can be found at [mcc.py](https://github.com/Zorua162/mcc.py). 
However, with these changes "localhost" can now also be used as an option. The IP still gets resolved to "127.0.0.1", but its a nice quality of life thing.

I also tested the case of Non-sensical IPs such as "127.0.0.0.1", which resulted in the expected error, and IPs such as "aaa", which also resulted in the expected error.

This implementation **doesn't** allow dangerous wildcards as the hostname such as "*", however if the Dns.GetHostAddress method is disliked, then this method may need to be allowed instead, otherwise running the WebHook bot in a container simply be too complicated to be worth doing.

### DIFF
--- a/MinecraftClient/ChatBots/WebSocketBot.cs
+++ b/MinecraftClient/ChatBots/WebSocketBot.cs
@@ -291,7 +291,17 @@ public class WebSocketBot : ChatBot
 
     public WebSocketBot()
     {
-        var match = Regex.Match(Config.Ip!, @"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}");
+        // Lookup the given address
+        try
+        {
+            _ip = Dns.GetHostAddresses(Config.Ip!).First().ToString();
+        }
+        catch (Exception e)
+        {
+            // Set Ip to a non-acceptable value to fail the Ip check
+            _ip = "not found";
+        }
+        var match = Regex.Match(_ip, @"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}");
 
         if (!match.Success)
         {
@@ -305,7 +315,6 @@ public class WebSocketBot : ChatBot
             return;
         }
 
-        _ip = Config.Ip;
         _port = Config.Port;
         _password = Config.Password;
         _authenticatedSessions = new();
--- a/MinecraftClient/ChatBots/WebSocketBot.cs
+++ b/MinecraftClient/ChatBots/WebSocketBot.cs
@@ -287,13 +287,21 @@ public class WebSocketBot : ChatBot
 
         [TomlInlineComment("$ChatBot.WebSocketBot.DebugMode$")]
         public bool DebugMode = false;
+
+        [TomlInlineComment("$ChatBot.WebSocketBot.AllowIpAlias$")]
+        public bool AllowIpAlias = false;
     }
 
     public WebSocketBot()
     {
+        _password = Config.Password;
+        _authenticatedSessions = new();
+        _waitingEvents = new();
+
         var match = Regex.Match(Config.Ip!, @"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}");
 
-        if (!match.Success)
+        // If AllowIpAlias is set to true in the config, then always ignore this check
+        if (!match.Success & !Config.AllowIpAlias!)
         {
             LogToConsole(Translations.bot_WebSocketBot_failed_to_start_ip);
             return;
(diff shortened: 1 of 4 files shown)


Zorua162:
Alternatively, it might be worth just removing the requirement for "ip" to be in the form "x.x.x.x", as I've just realized that even the above implementation won't work for me, as I actually need the hostname in the WebSocket uri to be "mcc_MCC_1" so that my docker network can handle it. 

I could do with input on this though before making this change, as I am unsure if there is a security reason behind requiring this

milutinke:
> Alternatively, it might be worth just removing the requirement for "ip" to be in the form "x.x.x.x", as I've just realized that even the above implementation won't work for me, as I actually need the hostname in the WebSocket uri to be "mcc_MCC_1" so that my docker network can handle it.
> 
> I could do with input on this though before making this change, as I am unsure if there is a security reason behind requiring this

Maybe add a configuration option like `IsRunningInDocker` that turns off the IP regex.
As for the above mentioned check, I've added it just to prevent an exception.
You can remove the check, but then add an exception handling if the web socket creation fails.

Zorua162:
Yeah, the configuration option sounds like a  much better implementation than what I was going for, I'll change my PR in a bit to do that instead.

I don't think `IsRunningInDocker` is quite descriptive of it though, so I might go with something like `AllowIpAlias`, which would be set to false by default, to preserve the current behaviour, then I'll explain the use case in the comment for it

Zorua162:
Just wanted to note, although your testing may have found different results, however I have found that giving non-existent IP addresses seems to work fine (it just obviously wouldn't be possible to connect to them), for example:
`[MCC] [WebSocketBot] Server started on ip aaa port: 8043`

Instead though there does seem to have an error that only occurs when "Failed to start a server!" error gets called, I think it relates  to `        _waitingEvents = new();` not getting created, which means that it doesn't exist. I'll move this to above where the error gets called, so that this error doesn't happen in the future.
```
[MCC] [WebSocketBot] Failed to start a server! The provided IP address is not a valid one!
[MCC] Version is supported.
Logging in...
[MCC] Server is in offline mode.
[MCC] Server was successfully joined.
Type '/quit' to leave the server.
[ERROR] OnGameJoined: Got error from MinecraftClient.ChatBots.WebSocketBot: System.NullReferenceException: Object reference not set to an instance of an object.    at MinecraftClient.ChatBots.WebSocketBot.AfterGameJoined() in /home/zorua162/repos/MCC/Minecraft-Console-Client/MinecraftClient/ChatBots/WebSocketBot.cs:line 1065    at MinecraftClient.McClient.<>c.<OnGameJoined>b__172_0(ChatBot bot) in /home/zorua162/repos/MCC/Minecraft-Console-Client/MinecraftClient/McClient.cs:line 2511    at MinecraftClient.McClient.DispatchBotEvent(Action`1 action, IEnumerable`1 botList) in /home/zorua162/repos/MCC/Minecraft-Console-Client/MinecraftClient/McClient.cs:line 2440```
 
In testing doing the change as mentioned above, with moving the definition of _waitingEvents above the IP check means that error doesn't happen, but instead the server gets started with no IP set, idk if there is a way to make bots fail out more gracefully though, or if the error is preffered:
```
[MCC] [WebSocketBot] Failed to start a server! The provided IP address is not a valid one!
[MCC] [WebSocketBot] Starting the Websocket server...
[MCC] [WebSocketBot] Server started on ip  port: 0
```

Zorua162:
I'm also a little confused how the translation setup works, I attempted to add in the extra translation string wherever I could find, however when I was building some stuff seems to have been completely regenerated in the file `MinecraftClient/Resources/ConfigComments/ConfigComments.Designer.cs`. If this isn't desired to be committed (I'd guess due to things like merge conflicts) then I'm happy to remove that from my PR, as I don't think its supposed to be there.

milutinke:
> Just wanted to note, although your testing may have found different results, however I have found that giving non-existent IP addresses seems to work fine (it just obviously wouldn't be possible to connect to them), for example: `[MCC] [WebSocketBot] Server started on ip aaa port: 8043`

I haven't tested that, I just assumed it would throw an exception.
The changes look fine, as for the translation, the generated files are fine, we have a Github Action that submits the translations to the website where we do the translations.

The PR looks fine, so I'll merge without building if you are fully done, then I'll submit the translations and then run the build manually.

PR #2617, opened 2023-11-19
Websocket bot: Create new Responder object that will respond to the session's new name

Zorua162:
I noticed that I wasn't getting any reply when I changed my session name, upon looking at the code it turns out this response was getting sent to the old session name. I have added a line which creates a new responder object with the new id to fix this.


milutinke:
Nice catch.
Thanks for contributing.

PR #2648, opened 2024-01-28
Fixed chat issues

PeachesMLG:
Fixed issues with chat packets

```
Unhandled exception. System.IO.InvalidDataException: Failed to process incoming packet of type ChatMessage. (PacketID: 55, Protocol: 764, LoginPhase: False, InnerException: System.Collections.Generic.KeyNotFoundException).

 ---> System.Collections.Generic.KeyNotFoundException: The given key 'text' was not present in the dictionary.

   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)

   at MinecraftClient.Protocol.Handlers.Protocol18Handler.HandlePlayPackets(Int32 packetId, Queue`1 packetData)

   at MinecraftClient.Protocol.Handlers.Protocol18Handler.HandlePacket(Int32 packetId, Queue`1 packetData)

   --- End of inner exception stack trace ---

   at MinecraftClient.Protocol.Handlers.Protocol18Handler.HandlePacket(Int32 packetId, Queue`1 packetData)

   at MinecraftClient.Protocol.Handlers.Protocol18Handler.Updater(Object o)

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)

--- End of stack trace from previous location ---

   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)

Aborted (core dumped)

```

PR #2650, opened 2024-01-28
Chat Crash Fix

milutinke:
Fixed a crash when a chat message was received on a server that uses protocol lib and viaversion.


milutinke:
Fixed in #2648 , I haven't seen it on time.
I'll mark this one as abandoned.

PR #2670, opened 2024-02-10
Add Sentry Error Tracking

breadbyte:
- Added a notice that we're using Sentry when first generating the config file
- Added a config option to disable Sentry crash logging

Currently sends information about the current .NET Runtime and the Operating System.
Here is a screenshot of all the information that is sent to Sentry whenever a crash occurs:

<details>
<summary>Very long screenshot</summary>

Screenshot_11-2-2024_4480

</details>

Currently uses my personal DSN.
Will need to change that to our DSN before merging.


milutinke:
Haven't looked in to the code, so you might have already added this.
But it would be a good idea to allow this only for binaries that we release here, because someone could modify the code and cause an error.

ReinforceZwei:
We could put the DSN into repository secret/variable and populate the DSN during build-time, just like how we populate the build info. Correspondingly, the code would need to check the DSN string before initializing Sentry SDK.

PR #2693, opened 2024-03-03
DigBlock AntiCheat Error Fix

Nekiplay:


Nekiplay:
I most likely copied an old version, expect an update

Nekiplay:
If anything, the problem is that now breaking the block is sent from the bottom, and if the player is on top of the block being broken and he breaks it from below, the antichit does not allow to break the block.

Nekiplay:
Kinda ready for cojones since I haven't found any comites breaking it.

ReinforceZwei:
Your forked repository is outdated. Please update your fork before opening PR.

PR #2697, opened 2024-03-05
Fix prevent AntiCheat Block Breaking (In Scripts)

Nekiplay:
- [x]  Fixed block breaking for anticheats in custom scripts
- [x] Added auxiliary methods for Direction

### DIFF
--- a/MinecraftClient/ChatBots/AutoDig.cs
+++ b/MinecraftClient/ChatBots/AutoDig.cs
@@ -285,7 +285,7 @@ namespace MinecraftClient.ChatBots
                     if (Config.Mode == Configs.ModeType.lookat ||
                         (Config.Mode == Configs.ModeType.both && Config._Locations.Contains(blockLoc)))
                     {
-                        if (DigBlock(blockLoc, lookAtBlock: false))
+                        if (DigBlock(blockLoc, Direction.Down, lookAtBlock: false))
                         {
                             currentDig = blockLoc;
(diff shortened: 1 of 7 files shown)


Nekiplay:
Ready for review

milutinke:
Looks good.
Also what is preventing you from porting that `getDirection` method to the MCC?
If it's the `mc.player.GetEyeHeight`, the eye height is `1.62` blocks high, so it can be hardcoded.

Nekiplay:
> Looks good. Also what is preventing you from porting that `getDirection` method to the MCC? If it's the `mc.player.GetEyeHeight`, the eye height is `1.62` blocks high, so it can be hardcoded. Also, where is that method from by the way?

getDirection method from Meteor Client
GetHorizontalFacing from Minecraft Code

milutinke:
> > Looks good. Also what is preventing you from porting that `getDirection` method to the MCC? If it's the `mc.player.GetEyeHeight`, the eye height is `1.62` blocks high, so it can be hardcoded. Also, where is that method from by the way?
> 
> getDirection I probably won't reschedule. getDirection method from Meteor Client GetHorizontalFacing from Minecraft Code

Does this work as is without getDirection method?

Nekiplay:
> Does this work as is without getDirection method?

yes

milutinke:
> > Does this work as is without getDirection method?
> 
> yes

Alright, thanks, I'll merge it.

PR #2707, opened 2024-03-12
Item Mappings for 1.8 - 1.12 + Crash Fix

milutinke:
Added item mappings for:
- 1.8
- 1.9
- 1.10
- 1.11
- 1.12

Fixed a crash with `Player Block Placement` packet.

TODO: Investigate why the `Player Block Packet` doesn't do anything on 1.8


milutinke:
Merging this in order to fix crashes, I'll investigate further 1.8 issue and create a separate PR to fix it.

PR #2723, opened 2024-04-13
Add the ability to be able to use different platforms with the Dockerfile

Zorua162:
Currently the example Dockerfile only works with amd64 architecture machines. This PR retains that default behaviour, however provides the ability to specify the MCC_PLATFORM environment variable, which changes the URL that the MCC release zip is downloaded from and so can be used to change the architecture to the desired one.


ReinforceZwei:
Looks good, merging your PR. Thank you for your contribution.

PR #2781, opened 2024-08-10
add Yggdrasil authlib multi-user selection.

a08381:
Some Yggdradil Authlib Server can provide multi-user creation, this can force which user should login in MCC.


old42356-byte:
can it work when my pc is turned off ?

PR #2821, opened 2024-12-05
Fixed bug in `SetDimension` method of `World` class, where it would c…

vinibretas:
Fixed bug in `SetDimension` method of `World` class, where it would crash if joining a paper server. Added error handling as well. 
The previous behavior remains unchanged, but instead of crashing, I added the possibility to have the dimension starting with "minecraft:". And after all that if still fails it throws an error.


milutinke:
Thank you for your contribution

PR #2864, opened 2025-04-29
Fix reading window items packet in versions below 1.17.1

DevBobcorn:


milutinke:
Thanks for you contibution

PR #2890, opened 2025-08-01
Fixes for authlib-injector accounts

evan-goode:
This PR includes several fixes for authlib-injector (here called "Yggdrasil" or "Authlib") accounts. For https://github.com/unmojang/drasl/issues/66.

- Add a new config option, `Main.General.AuthServer.AuthlibInjectorAPIPath`, to allow specifying the path to the root of the authlib-injector API. Whereas Blessing Skin places their API at `/api/yggdrasil`, other authlib-injector servers (Drasl, Ely.By) do not, so this change enables compatibility with them.
- Add docstrings for some `AuthServer` config options
- Don't send an invalid ("dummy") profile key signature for authlib-injector accounts; this breaks vanilla Minecraft servers
- Replace the custom HTTP implementation in `DoHTTPSRequest` with `System.Net.Http.HttpClient`, but still respect the configured proxy.
  - Note: I am unsure of this change. According to a comment, the custom HTTP client was used "since we must directly use a TcpClient because of the proxy." But it seems straightforward to pass a custom `TcpClient` to `HttpClient`---that's what I did here. It sent the request through my TCP proxy when I tested it. Is this correct?
- Fetch certificates for authlib-injector accounts if the authlib-injector server has the `feature.enable_profile_key` flag set

See commit message for more details. I'm a C# amateur, so a thorough review would be appreciated :)

### DIFF
--- a/MinecraftClient/Protocol/ProfileKey/KeyUtils.cs
+++ b/MinecraftClient/Protocol/ProfileKey/KeyUtils.cs
@@ -12,17 +12,73 @@ namespace MinecraftClient.Protocol.ProfileKey
     {
         private static readonly SHA256 sha256Hash = SHA256.Create();
 
-        private static readonly string certificates = "https://api.minecraftservices.com/player/certificates";
-
-        public static PlayerKeyPair? GetNewProfileKeys(string accessToken, bool isYggdrasil)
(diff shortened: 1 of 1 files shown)


milutinke:
Thanks for the work.
Have you tested this, does it full work?
The code looks fine, other than translations being hard coded, but we can add those in later.

PR #2972, opened 2026-03-23
misc: Added a prompt engineering skill

milutinke:
Created an advanced prompt engineering skill that produces the best (most suited) prompts based on the initial user request/prompt.

The skill can ask for clarifications if they're needed, and it adds:
- Role
- Context
- Plan
- Advanced reasoning framework / protocol
- Goals
- Scope / Constraints
- Output format
- Sub-agents list and purpose (if necessary)
- Adversarial review
- Anti hallucination measures / checks.

### DIFF
--- a/.skills/mcc-prompt-engineer/SKILL.md
+++ b/.skills/mcc-prompt-engineer/SKILL.md
@@ -1,0 +1,406 @@
+---
+name: mcc-prompt-engineer
+description: >
+  Manually triggered skill for the Minecraft Console Client (MCC) project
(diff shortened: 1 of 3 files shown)


milutinke:
@BruceChenQAQ Check this out.
Now when you want to do a task, trigger this skill as a command, give it your request and information and let it create a good prompt, then you can create a Specification or a Plan based on it. And get better results.

i've also improved the end-to-end testing skill.
Now with all these skills, we can do blazing fast development.

Please collect future conversation using this skill and other ones, so we can improve them.

PR #3075, opened 2026-04-10
fix: correct /bed translation typo

acamq:
## Summary
- correct the `/bed` translation string so MCC reports `Found a bed` instead of `Found a bet`
- update the tracked generated translation designer comment to match the resource value

## Validation
- verified the `/bed` command uses `Translations.cmd_bed_found_a_bed_at` in `MinecraftClient/Commands/Bed.cs`
- verified there are no remaining `Found a bet` matches in the repo
- `dotnet build MinecraftClient.sln -c Release` could not run in this environment because `dotnet` is not installed


BruceChenQAQ:
Thanks for your contribution

PR #3133, opened 2026-06-06
Update TelegramBridge.cs

werfrag:
The names of players containing the "_" symbol were not read.

A method has been added, which sends messages without Markdown formatting and used that in the text viewing function.


milutinke:
Thank you for your contribution

PR #3163, opened 2026-06-27
feat(inventory): expose NBT/component data in inventory MCP tools

italoseara:
Right now the inventory tools only return item type and count, which makes it impossible to distinguish a renamed item from a plain one, read server-specific metadata (custom_data), or check enchantments and potion contents. The material name alone isn't enough for most real automation tasks.

This adds a nullable `Nbt` field to `MccInventorySnapshotSlot`, `MccInventorySearchMatch`, and `MccItemStackSnapshot`:

- On 1.20.6+ it's a map of component name -> serialized component, e.g. `"minecraft:custom_data"`, `"minecraft:enchantments"`, `"minecraft:custom_name"`
- On older versions it falls back to the raw NBT dictionary
- Omitted entirely when the item has no extra data (plain Stone, Dirt, etc. come back clean with no Nbt field)

To make this work, `StructuredComponent` now stores the registry name it was parsed under (`ComponentName`), set in `ParseComponent()` alongside the existing `TypeId`. Both are marked `[JsonIgnore]` so they don't appear in the serialized output. `BuildNbt` uses `c.GetType()` when calling `JsonSerializer.SerializeToElement` - necessary because the list elements are typed as the abstract base class and the default serializer would otherwise produce empty objects.

Tested against a 1.21.11 server with a mix of custom items (custom_data, enchantments, block_entity_data) and vanilla items.

### DIFF
--- a/MinecraftClient/Protocol/Handlers/StructuredComponents/Core/StructuredComponent.cs
+++ b/MinecraftClient/Protocol/Handlers/StructuredComponents/Core/StructuredComponent.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Text.Json.Serialization;
 using MinecraftClient.Inventory.ItemPalettes;
 
 namespace MinecraftClient.Protocol.Handlers.StructuredComponents.Core;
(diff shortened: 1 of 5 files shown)


milutinke:
Merging

Issue #3134, opened 2026-06-08
[BUG] AutoRelog


Somersault41:
### Prerequisites

- [x] I made sure I am running the latest [development build](https://github.com/MCCTeam/Minecraft-Console-Client/releases/latest)
- [x] I tried to [look for similar issues](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one
- [x] I have set `debugmessages=true` in config to diagnose my issue
- [x] I have redacted session tokens and passwords before attaching screenshots

### Minecraft Version

1.20.4

### Console Client Version

459

### Expected Behavior

Reconnecting AutoRelog

### Actual Behavior

<img width="875" height="53" alt="Image" src="https://github.com/user-attachments/assets/2a6af5c3-ac8f-4825-8d3a-69d1d0b138cd" />

After displaying "connection lost," it says it will reconnect after a while, but it doesn't. In fact, the MCC freezes and nothing can be typed.

<img width="24" height="24" alt="Image" src="https://github.com/user-attachments/assets/3c8ab9e8-e281-4508-a622-99cd4f9f1dd6" />

This also happens when you left-click on CMD a few times, scroll down (to see the messages above), wait for a minute (this time may vary), and then click any key on the keyboard.

I can log in to the server when I close and reopen cmd.

The connection was interrupted because my internet connection went down momentarily, or my connection to the server was lost for a short time. AutoRelog shouldn't stop in this situation; it should keep trying to reconnect.

This happened when switching from a 2.4GHz to a 5GHz Wi-Fi connection (and vice versa).

This error doesn't occur in version 456, but it does in versions 459, 458, and 457. (I tested it.)

### Steps to Reproduce the bug

1. internet outage
2. disconnect and reconnect the internet connection
3. 

### Anythings that could help diagnosing the bug

```txt

```

### Device

Desktop

### Operating System

Windows

milutinke:
@Somersault41  Try this version: https://github.com/milutinke/Minecraft-Console-Client/releases/tag/20260613-164

Somersault41:
It didn't work, and many things broke. I can't navigate between lines with the mouse wheel in CMD. There's some kind of freezing issue in this version.

milutinke:
> It didn't work, and many things broke. I can't navigate between lines with the mouse wheel in CMD. There's some kind of freezing issue in this version.

[...]

PR #3169, opened 2026-07-18
This solves issue #3134.

Somersault41:
Problems:
1) The AutoRelog ChatBot stopped trying to log in after the 3rd attempt.
2) The AutoRelog ChatBot stopped trying to log in after the internet connection was interrupted. MCC froze.

2 problems have been solved.
This solves issue #3134.

### DIFF
--- a/MinecraftClient/ChatBots/AutoRelog.cs
+++ b/MinecraftClient/ChatBots/AutoRelog.cs
@@ -131,7 +131,7 @@ namespace MinecraftClient.ChatBots
             else if (CanReconnect())
             {
                 message = GetVerbatim(message);
-                string comp = message.ToLower();
+                string comp = message.ToLowerInvariant();
 
                 LogDebugToConsole(string.Format(Translations.bot_autoRelog_disconnect_msg, message));
 
--- a/MinecraftClient/ClassicConsoleBackend.cs
+++ b/MinecraftClient/ClassicConsoleBackend.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Text.RegularExpressions;
+using System.Threading;
 
 namespace MinecraftClient
 {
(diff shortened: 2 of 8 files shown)


milutinke:
There are some good changes here. The zero-byte socket read should prevent packet-reading loops after a clean disconnect, and the protocol cleanup is more reliable when cancellation throws.

However, the reconnect changes introduce serious problems:
- `RequestImmediateInput()` returns an empty string after one second. In "classic" console mode, the offline prompt treats that as an exit request, so MCC  can terminate without any user input.
- Each timeout starts another blocked `Task.Run()`. These tasks are not cancelled or tracked, and TUI mode also accumulates handlers. `BasicIO` still blocks indefinitely.
- `disconnectInProgress` only avoids the script-bot deadlock when the threads happen to run in a particular order. It does not fix the underlying synchronization problem. When it skips the deadlock, it also skips bot unloading and other cleanup, which can leak resources across reconnects.
- The new background console-stop thread is untracked and can race with another reconnect or a newly started session. Swallowed exceptions make these failures difficult to diagnose.-
- Restart requests can be replayed after a successful reconnect, causing MCC to disconnect and reconnect again unexpectedly. The restart-generation field is currently unused.
- The kick-message comparison uses `ToLowerInvariant()` on one side and culture-sensitive `ToLower()` on the other, which can break matching for some locales.
- The configuration locking changes are unrelated to the reported issue, add substantial complexity, and can block forever while saving settings.

So I cannot merge this in this state.
I got some gist of what the issue is actually, and I'll try fixing it and debugging it myself, as well as sending you a build to test, as I use Linux and macOS.

Edit: And I'll fix the merge issue.

milutinke:
@Somersault41 Try this one: https://github.com/MCCTeam/Minecraft-Console-Client/releases/tag/20260719-481

Somersault41:
It didn't work. MCC is crashing.

Minecraft Konsol İstemcisi v26.2 - MC 1.4.6 için 26.2 - Github.com/MCCTeam
16:07:43 GitHub build 481, built on 2026-07-19 from commit 64c1dc5
MCCyi çevirmemize yardımcı olun: https://crowdin.com/project/minecraft-console-client
16:07:43 Çevrimdışı modda giriş yaptınız.
16:07:43 Minecraft 1.20.4 sürümü kullanılıyor (protokol v765)
16:07:44 [MCC] Sürüm destekleniyor.
Giriş yapılıyor...
16:07:44 [MCC] Giriş başarısız :
16:07:44 [MCC] You are logging in too fast, try again later.
16:07:44 [ERROR] Bu sunucuda oturum açılamadı.
16:07:44 Geçerli ayarlar RealSaruhanAFK.ini dosyasına kaydediliyor
16:07:44 Geçerli ayarlar RealSaruhanAFK.ini dosyasına kaydediliyor
>

Sometimes it says it's saving to the .ini file twice, sometimes only once.
Note: I am opening more than one MCC at the same time.

It crashes when the internet connection is lost. AutoRelog doesn't work at all. It said it would be saved to the .ini file once. It didn't say it twice. It's not freezing anymore, it's collapsing.

Somersault41:
@milutinke The application crashes when the ExitOnFailure value is set to true. It doesn't crash when I set it to false.
I hadn't noticed that the ExitOnFailure setting was true. I've now set it to false and it's working fine. Sorry for overlooking it. The problem is fixed.

milutinke:
Ok, Awesome. I've merged my fix. Thanks for the feedback.

Issue #3152, opened 2026-06-17
[BUG] timestamps colored


Somersault41:
### Prerequisites

- [x] I made sure I am running the latest [development build](https://github.com/MCCTeam/Minecraft-Console-Client/releases/latest)
- [x] I tried to [look for similar issues](https://github.com/MCCTeam/Minecraft-Console-Client/issues?q=is%3Aissue) before opening a new one
- [x] I have set `debugmessages=true` in config to diagnose my issue
- [x] I have redacted session tokens and passwords before attaching screenshots

### Minecraft Version

1.20.4

### Console Client Version

471

### Expected Behavior

timestamps being colorless

### Actual Behavior

timestamps colored

This problem also existed in issue number #3056.

### Steps to Reproduce the bug

1. 
2. 
3. 

### Attach screenshot here (If applicable)

<img width="945" height="893" alt="Image" src="https://github.com/user-attachments/assets/b50fac18-6510-46d4-8861-d463f34dbe1f" />

### Anythings that could help diagnosing the bug

```txt

```

### Device

Desktop

### Operating System

Windows

milutinke:
@Somersault41 Try this one: https://github.com/milutinke/Minecraft-Console-Client/releases/tag/20260618-167

Somersault41:
colors are bad again

<img width="78" height="37" alt="Image" src="https://github.com/user-attachments/assets/380624ee-fcd7-422b-ae46-8465f3930ffe" />

However, the timestamps have been corrected.
@milutinke

PR #3197, opened 2026-08-01
It solves the problem of timestamps being colored according to the last message's color.

Somersault41:
#3152 solves the problem.


milutinke:
Merged, thanks for the contribution

PR #3216, opened 2026-08-27
Backup bots before unloading on transfer initiation

xiaozeqwq:
Added logic to backup current bots before unloading during transfer.


milutinke (changes requested):
Please keep the comments English only as this is an international project. Thanks.

